PR #27, opened 2024-02-21
[ORA-564] Multi-Topic Registration per Wallet + Restaking

guilherme-brandao:
Wallets can now sign up to be reputers AND/OR workers across multiple topics
"Registration" to a topic i.e. signing up to do work on a topic is a separate notion than "staking" in a topic i.e. placing money on someone doing work in a topic
Wallets can have stake in the system without being registered to do work on any topic -> but then they shouldn't get any rewards for this
proto Errors -> Cosmos SDK error style


x/emissions/keeper/genesis.go
```diff
@@ -15,9 +16,19 @@ func (k *Keeper) InitGenesis(ctx context.Context, data *state.GenesisState) erro
 
 	// ensure the module account exists
 	stakingModuleAccount := k.authKeeper.GetModuleAccount(ctx, state.AlloraStakingModuleName)
-	k.authKeeper.SetModuleAccount(ctx, stakingModuleAccount)
+	if stakingModuleAccount == nil {
```
relyt29, line 19: no. This code is wrong and got pulled in from the ORA-631 commit

this code should not be here, it needs to be deleted it will break the chain

relyt29 (approved):
thank you for your hard work.

PR #28, opened 2024-02-21
Kenny/ora 682 create per topic pricing mechanism

kpeluso:
These 2 tix:

https://linear.app/upshot/issue/ORA-617/optimize-when-weight-adjustment-is-ran-by-modifying-endblock

https://linear.app/upshot/issue/ORA-682/create-per-topic-pricing-mechanism

TODO:
Testing + places where I said TylerTODO

PR #29, opened 2024-02-21
[ORA-682][ORA-617][ORA-684][ORA-685][ORA-686] Inference Request Pricing Mechanism

relyt29:
This is branched off of PR #28 

Includes code to do inference request pricing
Move funds for those inferences to the rewards module to be paid out with rewards
Kick off inferences for the topics with the highest demand from those requests
and unit tests for the components of these flows


kpeluso (approved):
great!

PR #30, opened 2024-02-22
[ORA-687] Add met demand to rewards payments

kpeluso:
Needs testing, but in the same way that the base branch needs testing => not a draft PR, just a regular PR.


relyt29 (changes requested):
Ok, i think this should work theoretically but there's one issue which is that this code as currently written doesn't clear out accumulated demand in the rewards flow after rewards are given. So accumulated rewards will just keep getting bigger and bigger and bigger 

Otherwise I think this PR is a good idea and basically solves ORA-687

relyt29 (approved):
Approved pending a merge of #32 into this PR

kpeluso:
@relyt29  fixed what had to be fixed => mergable

PR #31, opened 2024-02-22
Exposed get registered topics query and fixes

guilherme-brandao:


x/emissions/keeper/msg_server.go
```diff
@@ -217,6 +217,13 @@ func (ms msgServer) AddNewRegistration(ctx context.Context, msg *state.MsgAddNew
 	if err != nil {
 		return nil, err
 	}
+	// check if topic exists
+	exist, err := ms.k.TopicExists(ctx, msg.TopicId)
```
xmariachi, line 221: rename `exist` to `topicExists`  here and in RemoveRegistration (and anywhere else applicable)

xmariachi (changes requested):
Small rename change, otherwise LGTM!

xmariachi (approved):

PR #34, opened 2024-02-23
Kenny/worker liveness check in reward calc 2

kpeluso:
Mask weights used in reward calculation if the worker has insufficient inferences in the past reward epoch.

Increment the underlying mapping when new inferences are added.

Remove these counts across all topics entirely at end of global reward calc.


x/emissions/keeper/msg_server.go
```diff
@@ -41,14 +41,13 @@ func (ms msgServer) CreateNewTopic(ctx context.Context, msg *state.MsgCreateNewT
...
-	// 	return nil, fmt.Errorf("inference cadence must be at least 60 seconds (1 minute)")
-	// }
+	if msg.InferenceCadence < 60 {
+		return nil, state.ErrInferenceCadenceBelowMinimum
+	}
 
-	// if msg.WeightCadence < 10800 {
-	// 	return nil, fmt.Errorf("weight cadence must be at least 10800 seconds (3 hours)")
-	// }
+	if msg.WeightCadence < 10800 {
```
relyt29, line 48: these should be hardcoded ALL_CAPS_VALUES at the top of the file so that when we go to do [ORA-618] (global emissions module params) we can easily find them to put them in the global params

x/emissions/keeper/msg_server.go
```diff
@@ -41,14 +41,13 @@ func (ms msgServer) CreateNewTopic(ctx context.Context, msg *state.MsgCreateNewT
 		return nil, err
 	}
 
-	// TODO: Add after demo
-	// if msg.InferenceCadence < 60 {
-	// 	return nil, fmt.Errorf("inference cadence must be at least 60 seconds (1 minute)")
-	// }
+	if msg.InferenceCadence < 60 {
```
relyt29, line 44: ALL_CAPS_VALUE

x/emissions/module/rewards_calc.go
```diff
@@ -195,7 +248,7 @@ func emitRewards(ctx sdk.Context, am AppModule) error {
 		return err
 	}
 
-	return nil
+	return am.keeper.ResetNumInferencesInRewardEpoch(ctx)
```
relyt29, line 251: I don't love this line of code, i think it's weirdly confusing

I understand what you're doing here and it's logically equivalent, but for readability's sake i'd much prefer

```
err = am.keeper.ResetNumInferencesInRewardEpoch(ctx)
if err != nil {
    return err
}

return nil
```

Resetting the number of inferences in the reward epoch is a task the emitRewards function should do, not the return of that function, it's weird


it's a nitpick feel free to ignore me if you disagree

x/emissions/module/rewards_calc.go
```diff
@@ -125,6 +130,54 @@ func GetParticipantEmissionsForTopic(
 	return rewards, nil
 }
 
+// This function checks topic weights and then masks them if not enough inferences were collected in that timestep
+// It should:
+//
+//	mask the inputted weights above with GetNumInferencesInRewardEpoch(),
+//	checking inference cadence by reward epoch length / topic inference cadence
+//	simple forgiveness check (only if many inferences are missing, don't reward)
+func MaskWeightsIfInsufficientLiveness(
```
relyt29, line 139: I would sleep much better at night if we had a unit test for this function that actually ensures that the function does as we expect

relyt29: Also this function probably does not need to be exported out of this module and can be lowercased

relyt29 (approved):
Broadly acceptable for merge, but as pointed out, I think this code should really include a specific unit test for just `MaskWeightsIfInsufficientLiveness`

PR #36, opened 2024-02-23
Update Autocli to support all RPC Methods

relyt29:
#34 and various other PRs have added a number of new transaction types to the tx.proto and RPC exposed by a node. This PR updates the autocli.go file to ensure that the node CLI tool can send every type of transaction that nodes support.


relyt29:
And yes I did do make sure the chain still runs after this lol

guilherme-brandao (changes requested):
Did you try to initialize the chain after adding this?

relyt29:
yes

PR #40, opened 2024-02-24
Script to deploy local multinode testnet

guilherme-brandao:
I created this script to try to replicate the demo issue locally, I think it will be very useful for future tests.

To access logs of the node after starting with `tmux` you need to run:

1) Open a new terminal
2) Run `tmux attach -t validator1`

### DIFF
--- a/scripts/multinode-local-testnet.sh
+++ b/scripts/multinode-local-testnet.sh
@@ -1,0 +1,89 @@
+#!/bin/bash
+set -e
+
+# Clean up any existing data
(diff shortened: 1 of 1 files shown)


relyt29 (approved):
I ran and confirmed this works on my computer


question of whether docker would be a more appropriate tool in the long run though

guilherme-brandao:
> I ran and confirmed this works on my computer
> 
> question of whether docker would be a more appropriate tool in the long run though

PR #42, opened 2024-02-27
add a couple badges, add more over time

dmikey:
badges


relyt29 (approved):
opinionated take but can we reduce the massive size of the allora logo, too?

kpeluso (approved):

kpeluso (approved):

PR #46, opened 2024-02-28
Block proposer triggers inference/weight generation 

guilherme-brandao:


app/topics_handler.go
```diff
@@ -0,0 +1,80 @@
...
+func NewTopicsHandler(emissionsKeeper emissionskeeper.Keeper) *TopicsHandler {
+	return &TopicsHandler{
+		emissionsKeeper: emissionsKeeper,
+	}
+}
+
+func (th *TopicsHandler) PrepareProposalHandler() sdk.PrepareProposalHandler {
+	return func(ctx sdk.Context, req *abci.RequestPrepareProposal) (*abci.ResponsePrepareProposal, error) {
+		fmt.Printf("\n ---------------- TopicsHandler ------------------- \n")
+		currentTime := uint64(ctx.BlockTime().Unix())
```
relyt29, line 26: shouldn't we probably use block time here?

relyt29: nvm i'm blind and stupid and tired

app/topics_handler.go
```diff
@@ -0,0 +1,80 @@
...
+		activeTopics, err := th.emissionsKeeper.GetActiveTopics(ctx)
```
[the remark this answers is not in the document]
relyt29: the actual list that needs to be handled is returned by the Churn function, active topics is just topics that have requests, the list returned by the churn is a subset of all active topics

x/emissions/keeper/keeper.go
```diff
@@ -1614,6 +1617,46 @@ func (k *Keeper) IncrementNumInferencesInRewardEpoch(ctx context.Context, topicI
 	return k.numInferencesInRewardEpoch.Set(ctx, key, newNumInferences)
 }
 
+// Reset the mapping entirely. Should be called at the end of every block
+func (k *Keeper) ResetChurnReadyTopics(ctx context.Context) error {
```
relyt29, line 1621: Hmm this is fine and will work but I don't think the most efficient way to do it. I would do it like:

In the proto I'd put a new type to have the automatic collections CollValue made for you otherwise you have to go in and write an encoder + decoder for []state.Topic which would be a pain in the ass, but doable
```
$types.proto

message TopicList {
    repeated Topic topics
}
```
make proto-gen blah blah 

```
$keeper.go

churnReadyTopics collections.Item[ state.TopicList ]

...

churnReadyTopics: collections.NewItem(sb, state.ChurnReadyTopicsKey, "churn_ready_topics", codec.CollValue[state.TopicList](cdc)),
```

then ResetChurnReadyTopics is O(1) instead of O(n)

```
churnReadyTopics.Set(nil);
```

and the Get operation also becomes O(1) instead of O(n) since you no longer need to walk the collection to create a []state.Topic

relyt29 (approved):
I left a comment about how I would optimize this but this code looks good


I have done 0 testing whatsoever that this actually works in any way, so uhhh maybe make sure to do that somehow before merging

x/emissions/keeper/keeper.go
kpeluso: Could maybe also reinitialize the map instead.

Also think it's fine to merge as-is anyway.

If you do do something different though: Note we have the same in `ResetNumInferencesInRewardEpoch()` so should apply there too.

kpeluso (approved):
+1 to what tyler said ab testing but everything good as-is too. Should get this deployed asap so we can guarantee it works for multiple live validators.

x/emissions/keeper/keeper.go
guilherme-brandao: Added suggested changes

PR #50, opened 2024-03-03
Install dasel to edit configs

vladupshot:


okedeji (approved):
LGTM if this is already tested locally by running docker build

PR #54, opened 2024-03-06
reset + rename some default params

kpeluso:
Diego researched what parameter values should be and I reset them. I also renamed some parameters to more easily-usable and predictable names in the process.


dmikey (approved):

x/emissions/proto/emissions/state/v1/tx.proto
```diff
@@ -27,74 +28,100 @@ service Msg {
...
+    (cosmos_proto.scalar) = "cosmos.Uint",
     (gogoproto.customtype) = "cosmossdk.io/math.Uint",
-    (gogoproto.nullable)   = false,
+    (gogoproto.nullable) = false,
     (amino.dont_omitempty) = true
-  ]; // total unmet demand for a topic < this => don't run inference solicatation or weight-adjustment
-  repeated uint64 max_topics_per_block = 5; // max number of topics to run cadence for per block
+  ]; // total unmet demand for a topic < this => don't run inference
+     // solicatation or weight-adjustment
+  repeated uint64 max_topics_per_block =
```
relyt29, line 96: why is the formatting in this file all weird, is that necessary

relyt29: the excess newlines, i mean

relyt29:
> looks ok to me pending you fix whatever broken test that is

seems like it might be demand_test.go:478

x/emissions/proto/emissions/state/v1/tx.proto
kpeluso: linter just did that idk. all around that file

x/emissions/proto/emissions/state/v1/tx.proto
```diff
@@ -27,74 +28,100 @@ service Msg {
...
+      5; // max number of topics to run cadence for per block
   repeated string min_request_unmet_demand = 7 [
```
xmariachi, line 98: No field number 6 here. Is that jump done for a reason? It wont affect Proto compiler but it just looks better.

xmariachi (commented):
Comment on proto seq

x/emissions/proto/emissions/state/v1/tx.proto
kpeluso: true

kpeluso: fixed in #56

PR #56, opened 2024-03-06
add error logs; handle div-0 for reward calc

kpeluso:
* Added error logs
* Added catch for div-0 error that may occur if workers aren't sufficiently live in a topic given the epoch length. If we catch that specific error then we don't do more by way of reward emission and simply continue with `EndBlock`


guilherme-brandao (approved):

guilherme-brandao:
Suggestion: use logger instead of `fmt.Println()`

It can be good for now to make these logs more visible during edgenet

relyt29 (approved):

relyt29:
> Suggestion: use logger instead of `fmt.Println()`
> 
> It can be good for now to make these logs more visible during edgenet

I agree we should use the logger but I think that we do fmt.Println all over the codebase so at this point I think i would be appropriate to make a backlog ticket for someone to go in and replace everywhere in the codebase with appropriate logging

relyt29:
> > Suggestion: use logger instead of `fmt.Println()`
> > It can be good for now to make these logs more visible during edgenet
> 
> I agree we should use the logger but I think that we do fmt.Println all over the codebase so at this point I think i would be appropriate to make a backlog ticket for someone to go in and replace everywhere in the codebase with appropriate logging

Created ORA-882

PR #60, opened 2024-03-07
Checking existing stake during registration

guilherme-brandao:
Details about this PR in this ticket: 
https://linear.app/upshot/issue/ORA-534/if-stake-was-set-before-node-was-registered-its-existing-stake-isnt


relyt29 (approved):
as far as I can tell lgtm

PR #63, opened 2024-03-08
Added get latest inference by worker and topic

guilherme-brandao:


x/emissions/keeper/query_server.go
```diff
@@ -154,10 +154,30 @@ func (qs queryServer) GetWeight(ctx context.Context, req *state.QueryWeightReque
...
+	}
+
+	topicExists, err := qs.k.TopicExists(ctx, req.TopicId)
+	if !topicExists {
+		return nil, status.Errorf(codes.NotFound, "topic %v not found", req.TopicId)
+	} else if err != nil {
+		return nil, err
+	}
+
+	inference, err := qs.k.GetInference(ctx, req.TopicId, workerAddr)
```
xmariachi, line 175: how do we know this gets the latest and not just any inference from that topic/worker key?

guilherme-brandao: because we are overwriting the inferences for the worker every time a new batch of inferences arrives

x/emissions/codec.go
```diff
@@ -10,7 +10,6 @@ import (
 func RegisterInterfaces(registry types.InterfaceRegistry) {
 	registry.RegisterImplementations((*sdk.Msg)(nil),
 		&MsgSetWeights{},
-		&MsgSetInferences{},
```
relyt29, line 13: what does this line of code do exactly?

relyt29 (approved):
lgtm

PR #65, opened 2024-03-09
Major Upgrade

kpeluso:
- Add new topic + global params;
- Add foundation whitelist to control subsidies + treasury actions
- Add core team members to foundation whitelist dedicated to control subsidies + treasury;
- Add tx to set foundation params per topic;
- Propagate changes to Topic struct + rest of relevant state;


x/emissions/params.go
```diff
@@ -20,6 +20,12 @@ func DefaultParams() Params {
 		MinWeightCadence:            uint64(60 * 60),               // 1 hour in seconds
 		MaxInferenceRequestValidity: uint64(60 * 60 * 24 * 7 * 52), // 52 weeks approximately 1 year in seconds
 		MaxRequestCadence:           uint64(60 * 60 * 24 * 7 * 52), // 52 weeks approximately 1 year in seconds
+		Sharpness:                   uint64(20),                    // controls going from stake-weighted consensus at low values to majority vote of above-average stake holders at high values
+		BetaEntropy:                 float32(0.25),                 // controls resilience of reward payouts against copycat workers
+		DcoefAbs:                    float32(0.001),                // delta for numerical differentiation
+		LearningRate:                float32(0.05),                 // speed of gradient descent
+		MaxGradientThreshold:        float32(0.001),                // gradient descent stops when gradient falls below this
+		MinStakeFraction:            float32(0.5),                  // minimum fraction of stake that should be listened to when setting consensus listening coefficients
```
guilherme-brandao, line 28: Probably we want to use `cosmosMath.Dec` instead of `float32`.

guilherme-brandao: same thing for `uint64` -> `cosmosMath.Uint`

guilherme-brandao (dismissed):

relyt29:
I am philosophically against this PR, as discussed on slack

* Not necessary to encode subsidies directly in protocol
   * Subsidies can be paid either via direct token transfer by the foundation themselves
   * or via a claim mechanism akin to aidrops
* This should not be our problem, let the foundation figure out how to do subsidies after mainnet launch, we should not be wasting engineering resources code complexity, and software maintenance burden on this during this critical time while trying to launch

kpeluso:
@relyt29 you're correct. We'll commit to updating that before the next 2wk but only when we have a better understanding from research regarding how best to apply subsidies.

x/emissions/params.go
kpeluso: down, sg. Feel free to apply if you get to it before me.

relyt29:
[b58a48b](https://github.com/allora-network/allora-chain/pull/65/commits/b58a48bbb7f140be0fed546b80ca6e2aad767b3c) is failing because of the issue in #304 - when that's merged it should flip green again

br4e (approved):
its happening!!!

PR #66, opened 2024-03-10
add forecasts + relevant txs

kpeluso:


x/emissions/keeper/msg_server.go
```diff
@@ -221,6 +221,37 @@ func (ms msgServer) ProcessInferences(ctx context.Context, msg *state.MsgProcess
 	return &state.MsgProcessInferencesResponse{}, nil
 }
 
+// A tx function that accepts a list of forecasts and possibly returns an error
+// Need to call this once per forecaster per topic inference solicitation round because protobuf does not nested repeated fields
+func (ms msgServer) ProcessForecasts(ctx context.Context, msg *state.MsgProcessForecasts) (*state.MsgProcessForecastsResponse, error) {
```
guilherme-brandao, line 226: Should this come in a second tx? I thought it would be an opcional parameter in the Inference.

x/emissions/keeper/keeper.go
```diff
@@ -138,6 +138,9 @@ type Keeper struct {
 	// map of (topic, worker) -> inference
 	inferences collections.Map[collections.Pair[TOPIC_ID, sdk.AccAddress], state.Inference]
 
+	// map of (topic, worker) -> forecast[]
+	forecasts collections.Map[collections.Pair[TOPIC_ID, sdk.AccAddress], state.Forecast]
```
relyt29, line 142: why are we storing the same data twice in forecasts and allForecasts?

We should be able to use better data structures than this, some combination maybe triple indexed to only store data once

x/emissions/keeper/keeper.go
```diff
@@ -276,6 +284,30 @@ func (k *Keeper) InsertInferences(ctx context.Context, topicId TOPIC_ID, timesta
 	return k.allInferences.Set(ctx, key, inferences)
 }
 
+// Insert a complete set of inferences for a topic/timestamp. Overwrites previous ones.
```
relyt29, line 287: this should read "a complete set of inference _forecasts_" not inferences, confusing

I_ik not I_ij from the paper

PR #71, opened 2024-03-13
Fix bug carrying over pointer from loop

xmariachi:
Loop var reuses mem addr. 
Fix that making a copy when accum them.

### DIFF
--- a/x/emissions/keeper/query_server.go
+++ b/x/emissions/keeper/query_server.go
@@ -291,7 +291,8 @@ func (qs queryServer) GetAllExistingInferenceRequests(ctx context.Context, req *
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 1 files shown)


x/emissions/keeper/query_server.go
```diff
@@ -291,7 +291,8 @@ func (qs queryServer) GetAllExistingInferenceRequests(ctx context.Context, req *
 		if err != nil {
 			return nil, err
 		}
-		ret = append(ret, &state.InferenceRequestAndDemandLeft{InferenceRequest: &inferenceRequest, DemandLeft: demandLeft})
+		inferenceRequestCopy := inferenceRequest
```
relyt29, line 294: is this enough, do we need to maybe do a deep copy?

xmariachi: This is enough, just creating a transient new var. 
This has been tested locally

guilherme-brandao (approved):

relyt29 (approved):

PR #79, opened 2024-03-19
Guilherme/ora 993 rewards v2 calcs

guilherme-brandao:


x/emissions/module/rewards_v2_test.go
```diff
@@ -0,0 +1,201 @@
...
+	var reputersWorker2ReportedOneOutLosses []float64
+
+	for _, lossBundle := range lossBundles.LossBundles {
+		reputerAddr, err := sdk.AccAddressFromBech32(lossBundle.Reputer)
+		s.NoError(err, "Error getting reputerAddr")
+
+		reputerStake, err := s.emissionsKeeper.GetStakeOnTopicFromReputer(s.ctx, topicId, reputerAddr)
+		s.NoError(err, "Error getting reputerStake")
+
+		reputerStakeFloat, _ := reputerStake.BigInt().Float64()
```
relyt29, line 105: so this is where we need to be weary about truncation vs the correct encoding in float because if the value is large, it'll just get rounded to the closest value that the float can represent or +Inf

x/emissions/module/rewards_v2.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// GetWorkerScore calculates the worker score based on the losses and lossesCut.
+func GetWorkerScore(losses, lossesCut float64) float64 {
+	deltaLogLoss := math.Log10(lossesCut) - math.Log10(losses)
+	return deltaLogLoss
+}
+
+// GetStakeWeightedLoss calculates the stake-weighted average loss.
+func GetStakeWeightedLoss(reputersStakes, reputersReportedLosses []float64) (float64, error) {
```
relyt29, line 15: what formula does this correspond to in the paper? I think it's a bad idea to put the formula number because those might change as the whitepaper gets edited, but maybe copy the latex into the comment is my feeling

x/emissions/module/rewards_v2.go
```diff
@@ -0,0 +1,41 @@
...
+	var totalWeightedLoss float64 = 0
+	for i, loss := range reputersReportedLosses {
+		if loss <= 0 {
+			return 0, fmt.Errorf("loss values must be greater than zero")
+		}
+		weightedLoss := (reputersStakes[i] / totalStake) * math.Log10(loss)
+		totalWeightedLoss += weightedLoss
+	}
+
+	stakeWeightedLoss := math.Pow(10, totalWeightedLoss)
```
kpeluso, line 39: you might not have to do this. likely all the values could be log-transformed until they actually update stakes/rewards, at the end

kpeluso (changes requested):
+1 to what tyler said ab providing docs. should do more "Document driven design" than TDD imo so 1. copilot can help more and 2. to organize your thoughts better

x/emissions/module/rewards_v2.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"fmt"
+	"math"
+)
+
+// GetWorkerScore calculates the worker score based on the losses and lossesCut.
+// Consider the staked weighted inference loss and one-out loss to calculate the worker score.
+// T_ij / T_ik / T^-_ik / T^+_ik
+func GetWorkerScore(losses, lossesCut float64) float64 {
```
kpeluso, line 11: think losses and lossesCut are odd names for this

x/emissions/module/rewards_v2.go
```diff
@@ -0,0 +1,44 @@
...
+	if totalStake == 0 {
+		return 0, fmt.Errorf("total stake cannot be zero")
+	}
+
+	var stakeWeightedLoss float64 = 0
+	for i, loss := range reputersReportedLosses {
+		if loss <= 0 {
+			return 0, fmt.Errorf("loss values must be greater than zero")
+		}
+		weightedLoss := (reputersStakes[i] / totalStake) * math.Log10(loss)
```
kpeluso, line 39: would do multiplication first to minimize truncation

x/emissions/module/rewards_v2.go
```diff
@@ -0,0 +1,44 @@
...
+func GetWorkerScore(losses, lossesCut float64) float64 {
+	deltaLogLoss := math.Log10(lossesCut) - math.Log10(losses)
+	return deltaLogLoss
+}
+
+// GetStakeWeightedLoss calculates the stake-weighted average loss.
+// Consider the losses and the stake of each reputer to calculate the stake-weighted loss.
+// The stake weighted loss is used to calculate the network-wide losses.
+// L_i / L_ij / L_ik / L^-_i / L^-_il / L^+_ik
+func GetStakeWeightedLoss(reputersStakes, reputersReportedLosses []float64) (float64, error) {
```
kpeluso, line 20: no types on the first arg? same for function above

guilherme-brandao: Omitting the type of all but the last parameter in a list of parameters with the same type in Go indicates that these parameters share the same type as the one specified.

x/emissions/module/rewards_v2.go
```diff
@@ -0,0 +1,56 @@
...
+	if len(reputersStakes) != len(reputersReportedLosses) {
+		return 0, fmt.Errorf("slices must have the same length")
```
relyt29, line 34: excuse me ser these should all be cosmos errors living in error.go

guilherme-brandao: will do ser! was finishing the other formulas to batch all the errors and address them in one go

x/emissions/module/rewards_v2_test.go
```diff
@@ -0,0 +1,793 @@
...
+			}
+
+			if !slicesAreApproxEqual(got, tt.want, 1e-2) {
+				t.Errorf("GetAllConsensusScores() got = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestGetAllReputersOutput(t *testing.T) {
```
kpeluso, line 304: I think you should also test the number of iterations as well for diff input, and ensure it iterates the expected number of times. this is an algorithm not commonly if ever found on-chain and for good reason -- it could get quite expensive. so we should be extra sure we're properly bounding the computation here and testing that's the case

x/emissions/module/rewards_v2.go
```diff
@@ -0,0 +1,430 @@
...
+// the difference between each reputer's losses and the consensus losses, taking into account each reputer's stake.
+// returns:
+// T_im - reputer score (proximity to consensus)
+// a_im - listening coefficients
+func GetAllReputersOutput(allLosses [][]float64, stakes []float64, initialCoefficients []float64, numReputers int, enableListening bool) ([]float64, []float64, error) {
+	learningRate := 0.01
+	coefficients := make([]float64, len(initialCoefficients))
+	copy(coefficients, initialCoefficients)
+
+	if enableListening {
```
kpeluso, line 243: think instead of this control flow we agree to do it or not... there was control flow in the simulation cuz they were testing what happens with/without listening enabled. you can double check w diederik but quite sure the resolution was that it was needed => can remove this if-statement

x/emissions/module/rewards_v2.go
```diff
@@ -0,0 +1,430 @@
...
+	return scores, nil
+}
+
+// GetAllReputersOutput calculates the final scores and adjusted listening coefficients for all reputers.
+// This function iteratively adjusts the listening coefficients based on a gradient descent method to minimize
+// the difference between each reputer's losses and the consensus losses, taking into account each reputer's stake.
+// returns:
+// T_im - reputer score (proximity to consensus)
+// a_im - listening coefficients
+func GetAllReputersOutput(allLosses [][]float64, stakes []float64, initialCoefficients []float64, numReputers int, enableListening bool) ([]float64, []float64, error) {
```
kpeluso, line 238: plz make a follow up ticket for after you stitch things tg to test for "bounded iterations" (as I said in other comment) but also to simultaneously clean and optimize this function. "coeffs, cofficients, coeffs2..." plus all the copies and repeated compute (e.g. you should be able to call GetAllConsensusScores once not twice; you probs also have more `copy` calls than you need). you implemented the code in the simulation faithfully but perhaps too much so -- pandas in the simulation lets you get away with more compute than an on-chain environment, so we need to care about optimizations here more than the simulation authors had to

x/emissions/module/rewards_v2.go
```diff
@@ -0,0 +1,430 @@
...
+	for _, stake := range reputersAdjustedStakes {
+		totalStake += stake
+	}
+
+	// Ensure every loss array is non-empty and calculate geometric mean
+	stakeWeightedLoss := make([]float64, len(reputersReportedLosses[0]))
+	for j := 0; j < len(reputersReportedLosses[0]); j++ {
+		logSum := 0.0
+		for i, losses := range reputersReportedLosses {
+			logSum += (math.Log10(losses[j]) * reputersAdjustedStakes[i]) / totalStake
```
kpeluso, line 165: a super easy way to save on compute that applies here + anywherhe else where there is a normalization (plenty of places) is to calculate the sum as you're looping through this then divide the weighted sum in the end. So 2 for-loops turns into 1, a division at every iteration turns into just 1, and you just increment 2 variables as you go along: the weighted sum (numerator) and the normalization constant (denominator)

x/emissions/module/rewards_v2.go
```diff
@@ -0,0 +1,430 @@
...
+// T_im
+func GetConsensusScore(reputerLosses, consensusLosses []float64) (float64, error) {
+	fTolerance := 0.01
+	if len(reputerLosses) != len(consensusLosses) {
+		return 0, types.ErrInvalidSliceLength
+	}
+
+	var sumLogConsensusSquared float64
+	for _, cLoss := range consensusLosses {
+		sumLogConsensusSquared += math.Pow(cLoss, 2) //!
```
kpeluso, line 183: why the `//!` ?

kpeluso (commented):
will re-review after you address all these comments first

x/emissions/module/rewards_v2.go
guilherme-brandao: added during tests, gonna remove it

x/emissions/keeper/keeper.go
```diff
@@ -379,6 +379,16 @@ func (k *Keeper) GetLossBundlesAtOrAfterTimestamp(ctx context.Context, topicId T
 	return lossBundles, nil
 }
 
+// Get loss bundles for a topic/timestamp
+func (k *Keeper) GetLossBundles(ctx context.Context, topicId TOPIC_ID, timestamp uint64) (*types.LossBundles, error) {
```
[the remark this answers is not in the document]
guilherme-brandao: I added that during tests, probably will use the `GetNetworkLossBundleAtOrBeforeTime`

x/emissions/module/rewards_v2.go
```diff
@@ -0,0 +1,434 @@
...
+import (
+	"fmt"
+	"math"
+
+	errors "cosmossdk.io/errors"
+	"github.com/allora-network/allora-chain/x/emissions/types"
+)
+
+// SmoothAbsoluteX calculates the smooth absolute function.
+func SmoothAbsoluteX(x []float64, p float64) ([]float64, error) {
```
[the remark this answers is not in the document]
guilherme-brandao: No it shouldn't, just deleted

x/emissions/module/rewards_v2.go
```diff
@@ -0,0 +1,434 @@
...
+
+import (
+	"fmt"
+	"math"
+
+	errors "cosmossdk.io/errors"
+	"github.com/allora-network/allora-chain/x/emissions/types"
+)
+
+// SmoothAbsoluteX calculates the smooth absolute function.
```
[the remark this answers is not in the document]
guilherme-brandao: function deleted

x/emissions/module/rewards_v2.go
```diff
@@ -0,0 +1,430 @@
+package module
+
+import (
+	"math"
+
+	errors "cosmossdk.io/errors"
+	"github.com/allora-network/allora-chain/x/emissions/types"
+)
+
+// SmoothAbsoluteX calculates the smooth absolute function.
```
relyt29, line 10: is this phi? it looks like phi over an array

x/emissions/module/rewards_v2.go
```diff
@@ -0,0 +1,430 @@
...
+	result := make([]float64, len(x))
+	for i, v := range x {
+		expV := math.Exp(v)
+		logExpV := math.Log(1 + expV)
+		result[i] = math.Pow(logExpV, p)
+	}
+	return result, nil
+}
+
+// StdDev calculates the standard deviation of a slice of float64.
```
relyt29, line 21: humor me can you add a comment here:
```
// stdDev = sqrt((Σ(x - μ))^2/ N)
// where μ is mean and N is number of elements
```

x/emissions/module/rewards_v2.go
```diff
@@ -0,0 +1,430 @@
...
+		mean += v
+	}
+	mean /= float64(len(data))
+	for _, v := range data {
+		sd += math.Pow(v-mean, 2)
+	}
+	sd = math.Sqrt(sd / float64(len(data)))
+	return sd
+}
+
```
relyt29, line 34:
```
// flatten converts a double slice of float64 to a single slice of float64
```

x/emissions/module/rewards_v2.go
guilherme-brandao: yeah, deleted it to use phi

x/emissions/module/rewards_v2.go
guilherme-brandao: not sure if it's applicable here, since I'm calculated that stake weighted loss for each set of reported losses.

x/emissions/module/rewards_v2.go
guilherme-brandao: ticket created!

x/emissions/module/rewards_v2_test.go
guilherme-brandao: ticket created!

relyt29 (approved):

PR #81, opened 2024-03-19
Derek/ora 1006 release allorad binaries

dmikey:
needed to make some updates to the build script, and then added an install script


relyt29 (approved):
didn't test it, but this is trivially simplistic so i see no problem merging

PR #83, opened 2024-03-20
Implement functions for math formulas with unit tests

relyt29:
Implements equation 6 (phi), and 32-41 (Reputer intra-topic rewards)

Does not wire equations together. Just implements equations and provides unit tests to make sure the math works as expected.

Un-fucks the integration tests in the major-upgrade repo so that tests can actually pass.

Is built off an earlier commit in #79. At this point I would say it's not dependent on 79, and the squash+merge will make 79 fine to merge into major-upgrade (as long as guilherme pulls in changes)


x/emissions/module/rewards_v2.go
```diff
@@ -428,3 +428,386 @@ func GetAdjustedStake(
...
+
+// Used by Rewards fraction functions,
+// all the exponential moving average functions take the form
+// x_average=α*x_current + (1-α)*x_previous
+//
+// this covers the equations
+// Uij = αUij + (1 − α)Ui−1,j
+// ̃Vik = αVik + (1 − α)Vi−1,k
+// ̃Wim = αWim + (1 − α)Wi−1,m
+func exponentialMovingAverage(alpha float64, current float64, previous float64) (float64, error) {
```
kpeluso, line 440: should put all this kinda stuff in a common math lib/file

x/emissions/module/rewards_v2.go
```diff
@@ -428,3 +428,386 @@ func GetAdjustedStake(
...
+			"entropyInference: %f, entropyForecasting: %f, forecastingUtility: %f",
+			entropyInference,
+			entropyForecasting,
+			forecastingUtility,
+		)
+	}
+	numerator := entropyInference + entropyForecasting
+	denominator := (1-forecastingUtility)*entropyInference + forecastingUtility*entropyForecasting
+	ret := numerator / denominator
+	if math.IsInf(ret, 0) {
```
kpeluso, line 789: is this pattern better/recommended than checking if denominator is 0 or close to it before the division on the line above?

kpeluso (approved):

guilherme-brandao (approved):

PR #86, opened 2024-03-21
Stand up an IBC relayer to test out the bridging functionality

RedBird96:
Before testing you have to download relayer app from this link
https://github.com/cosmos/relayer/releases

You can test as following commands

- make init-allora
- make init-axelar
- init-relayer
- start-relayer


x/ibc/scripts/init-allorad.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+BINARY=./allorad
+CHAIN_DIR=./data
+CHAINID=allora_demo
+RLY_MNEMONIC="alley afraid soup fall idea toss can goose become valve initial strong forward bright dish figure check leopard decide warfare hub unusual join cart"
```
relyt29, line 6: I don't know how great an idea it is to store mnemonics in code, even for testing purposes, can we can do something better here?

x/ibc/scripts/init-allorad.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+BINARY=./allorad
```
relyt29, line 3: probably better to do 

```
BINARY=$(which allorad)
```

x/ibc/scripts/init-allorad.sh
relyt29: e.g. create a new key like alice or bob and then use that instead of hardcoding a mnemonic

x/ibc/scripts/relayer/chains/alloard.json
```diff
@@ -0,0 +1,22 @@
...
+  "value": {
+    "key": "testkey",
+    "chain-id": "allora_demo",
+    "rpc-addr": "http://localhost:16657",
+    "grpc-addr": "",
+    "account-prefix": "cosmos",
+    "keyring-backend": "test",
+    "default-denom": "stake",
+    "gas-adjustment": 1.5,
+    "gas-prices": "0.025stake",
```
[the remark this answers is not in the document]
guilherme-brandao: opcional: also we can use similar chain configs to the original allorad e.g account-prefix: allo

x/ibc/scripts/start-icad.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+BINARY=./icad
+CHAIN_DIR=./data
+CHAINID_1=alloard
```
guilherme-brandao, line 5: typo:  allorad*

guilherme-brandao (approved):

PR #87, opened 2024-03-21
Kenny/ora 1049 unify cadences add ground truth lag topic param global

kpeluso:
- unify cadences -> epoch length
- add ground_truth_lag topic param;
- add max num workers and reputers per topic request global params

Ticket:]
https://linear.app/upshot/issue/ORA-1049/unify-cadences-add-ground-truth-lag-topic-param-global-params-num-top


x/emissions/module/module.go
```diff
@@ -194,32 +194,18 @@ func (am AppModule) EndBlock(ctx context.Context) error {
...
-
-			// Check the cadence of loss calculations
-			if currentTime-topic.LossLastRan >= topic.LossCadence {
-				fmt.Printf("Loss cadence met for topic: %v metadata: %s default arg: %s \n",
-					topic.Id,
-					topic.Metadata, topic.
-						DefaultArg)
-
-				// Update the last loss update ran
-				err = am.keeper.UpdateTopicLossUpdateLastRan(sdkCtx, topic.Id, currentTime)
```
relyt29, line 218: we're not going to update this now?

x/emissions/types/params.go
```diff
@@ -26,6 +25,8 @@ func DefaultParams() Params {
 		LearningRate:                  float32(0.05),                             // speed of gradient descent
 		MaxGradientThreshold:          float32(0.001),                            // gradient descent stops when gradient falls below this
 		MinStakeFraction:              float32(0.5),                              // minimum fraction of stake that should be listened to when setting consensus listening coefficients
+		MaxWorkersPerTopicRequest:     uint64(20),                                // maximum number of workers that can be assigned to a single inference request
```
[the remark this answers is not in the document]
kpeluso: even if we get responses from more than that many, we simply ignore if they're not performant workers as of the last round

x/emissions/module/module.go
kpeluso: inference + loss epochs are being unified in this PR into 1 "topic epoch" => only 1 thing needs to be updated

PR #88, opened 2024-03-22
Added slashing module

guilherme-brandao:


relyt29 (approved):
2 things:

1. fix the formatting, trivial - run `gofmt` in the root of the repo (or alternatively on just the app directory)
2. This looks good to me, but just sanity check that the module order is ok? can maybe check osmosis or other live chains for their module order with regards to the slashing module

otherwise good catch, good pr, :+1: from me dawg

PR #89, opened 2024-03-24
add eq3-8

kpeluso:


relyt29:
same comment as I had on PR 90, sorry I read that one first and this one second

Is there a reason we're using a timestamps in these functions and not block heights? Can we use block heights instead without breaking anything? I'm reminded of the consensus bug we had due to differing timestamps among validators

x/emissions/keeper/keeper.go
```diff
@@ -438,25 +429,6 @@ func (k *Keeper) GetParamsRewardCadence(ctx context.Context) (int64, error) {
 	return params.RewardCadence, nil
 }
 
-// Gets the total sum of all stake in the network across all topics
-func (k *Keeper) GetTotalStake(ctx context.Context) (Uint, error) {
```
relyt29, line 442: I question why we're deleting this, I would think this is something we would want to expose via query even if we don't use it ourselves

x/emissions/module/network_summaries.go
```diff
@@ -0,0 +1,141 @@
...
+	"math"
+
+	"github.com/allora-network/allora-chain/x/emissions/keeper"
+	emissions "github.com/allora-network/allora-chain/x/emissions/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+/**
+ * This file contains the logic from Section 3 of the litepaper,
+ * which churns current inferences, forecasts, previous losses and regrets
```
relyt29, line 14: churn is definitely not the correct word here

it combines

x/emissions/module/network_summaries.go
```diff
@@ -0,0 +1,141 @@
...
+ * into current network losses and regrets
+ */
+
+const p = 2.0
```
relyt29, line 18: these should be module params, made updateable, I think

relyt29:
copying comments from PR 90 here, as they apply here as well:



I think you should never refer to these equations by number because the number of an equation in the paper may change based on some edits to the paper over time. In my PRs I copy the actual equation math and put it directly in the comment. That way folks referencing the paper after the fact can look for the equation in the paper for the relevant information without any possibility that equation 9-11 may be totally different equations in the paper 3 months from now

It also makes my job easier as your PR reviewer, I don't have to go look at the paper to figure out wtf equations 9-11 are to check your implementation of them isn't broken

relyt29 (changes requested):
This PR should have unit tests for the math functions and should also pass the github actions, right now theres some issue with having deleted `GetAllInferences` probably you need to make-proto

kpeluso:
updated in branch off this PR. will merge in as soon as that passes. then will update branch to integrate merge confs

x/emissions/types/params.go
```diff
@@ -34,7 +36,7 @@ func DefaultParamsVersion() string {
 	return DefaultParams().Version
 }
 
-func DefaultParamsEpochLength() int64 {
+func DefaultParamsEpochLength() BLOCK_NUMBER {
```
relyt29, line 39: technically speaking this isn't a block number (in the sense of "block 62") but rather a delta of blocks (like "5 blocks") ...

x/emissions/types/keys.go
```diff
@@ -31,8 +31,8 @@ var (
 	AllInferencesKey              = collections.NewPrefix(20)
 	AllForecastsKey               = collections.NewPrefix(21)
 	AllLossBundlesKey             = collections.NewPrefix(22)
-	NetworkLossBundlesKey         = collections.NewPrefix(23)
-	StakeRemovalQueueKey          = collections.NewPrefix(24)
+	StakeRemovalQueueKey          = collections.NewPrefix(23)
+	StakeByReputerAndTopicId      = collections.NewPrefix(24)
```
relyt29, line 35: why is this here when 5 lines below it we have `StakeByReputerAndTopicIdKey`

x/emissions/types/inference_request.go
```diff
@@ -45,8 +45,8 @@ func CreateNewInferenceRequestFromListItem(sender string, item *RequestInference
 		Cadence:              item.Cadence,
 		MaxPricePerInference: item.MaxPricePerInference,
 		BidAmount:            item.BidAmount,
-		TimestampValidUntil:  item.TimestampValidUntil,
-		LastChecked:          0,
+		BlockValidUntil:      item.BlockValidUntil,
+		BlockLastChecked:     0,
```
relyt29, line 49: this is going to break rewards.sh in the integration folder fyi, but I agree with the change

x/emissions/module/rewards_calc.go
```diff
@@ -182,14 +182,14 @@ func MaskWeightsIfInsufficientLiveness(
 			if err != nil {
 				return nil, err
 			}
-			maxPossibleInferencesInRewardEpoch := uint64(rewardCadence) / topic.EpochLength
+			maxPossibleInferencesInRewardEpoch := rewardCadence / topic.EpochLength
```
relyt29, line 185: this entire file is deleted in the branch you are trying to merge into so probably this goes away when you go to merge

x/emissions/module/network_summaries.go
```diff
@@ -0,0 +1,271 @@
...
+	epsilon, err := k.GetParamsEpsilon(ctx)
+	if err != nil {
+		fmt.Println("Error getting epsilon: ", err)
+		return float64(0), err
+	}
+
+	pInferenceSynthesis, err := k.GetParamsPInferenceSynthesis(ctx)
+	if err != nil {
+		fmt.Println("Error getting epsilon: ", err)
+		return float64(0), err
```
guilherme-brandao, line 267: why this `float64(0)` in all returns? I can see other examples above that doesn't require converting 0 to float64

relyt29 (approved):
No more commit conflicts please for the love of god

x/emissions/types/inference_request.go
relyt29: fixed it

x/emissions/module/network_summaries.go
relyt29: not kenny, but usually depends on whether the compiler is smart enough to figure out the type by itself

but could also be written `return 0.0, err`

PR #90, opened 2024-03-24
add eq9-11

kpeluso:


relyt29:
Is there a reason we're using a timestamps in these functions and not block heights? Can we use block heights instead without breaking anything? I'm reminded of the consensus bug we had due to differing timestamps among validators

x/emissions/module/network_summaries.go
```diff
@@ -132,10 +117,132 @@ func CalcForcastImpliedInferencesAtTime(
...
+	networkLossBundle, err := k.GetNetworkLossBundleAtOrBeforeTime(ctx, topicId, time)
+	if err != nil {
+		fmt.Println("Error getting network losses: ", err)
+		return nil, err
+	}
+
+	return CalcForcastImpliedInferencesAtTime(ctx, k, topicId, inferences, forecasts, networkLossBundle)
+}
+
+// Calculates eq9-11 from the litepaper using the given inferences, forecast-implied inferences, and network regrets
```
relyt29, line 161: I think you should never refer to these equations by number because the number of an equation in the paper may change based on some edits to the paper over time. In my PRs I copy the actual equation math and put it directly in the comment. That way folks referencing the paper after the fact can look for the equation in the paper for the relevant information without any possibility that equation 9-11 may be totally different equations in the paper 3 months from now

It also makes my job easier as your PR reviewer, I don't have to go look at the paper to figure out wtf equations 9-11 are to check your implementation of them isn't broken

relyt29:
`CalcNetworkCombinedInference` and the various other stuff added in this PR would probably benefit from some unit tests prior to a merge

x/emissions/module/network_summaries.go
kpeluso: done

PR #93, opened 2024-03-27
[ORA-1050] PR 89 but Rebased: create eq3 8 on chain rebase

relyt29:
I despise git rebase, what a waste of my life


relyt29:
Ok this is the same as #89, nothing is fixed from the rebase

PR #94, opened 2024-03-27
[ORA-1091] Tokens to Mint Each Block (inflation schedule)

relyt29:
implement section 5.1 of the paper


x/mint/proto/mint/v1beta1/types.proto
```diff
@@ -7,72 +7,45 @@ import "gogoproto/gogo.proto";
 import "cosmos_proto/cosmos.proto";
 import "amino/amino.proto";
 
-// Minter represents the minting state.
-message Minter {
-  // current annual inflation rate
-  string inflation = 1 [
```
guilherme-brandao, line 13: fyi we are using this inflation parameter in the explorer, but I'm not sure if we still want to calculate a inflation rate on chain and update it frequently just for query purposes. 

cc @kpeluso

relyt29: yeah i deleted all that code

we can make some kind of workaround to support the explorer that's a read only query call

relyt29: alright I put it back, might want to check the math but yea

x/mint/keeper/emissions.go
```diff
@@ -0,0 +1,185 @@
...
+// controls the inflation rate of the token issuance
+//
+// ^e_i = (f_e*T_{total,i}) / N_{staked,i}) * (N_{circ,i} / N_{total,i})
+//
+// where T_{total,i} is the total number of tokens held by the ecosystem
+// treasury, N_{total,i} is the total token supply, N_{circ,i} is the
+// circulating supply, and N_{staked,i} is the staked supply. The
+// factor f_e = 0.015 month^{−1} represents the fraction of the
+// ecosystem treasury that would ideally be emitted per unit time.
+func TargetRewardEmissionPerUnitStakedToken(
```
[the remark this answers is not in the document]
relyt29: done

x/mint/keeper/emissions.go
```diff
@@ -0,0 +1,185 @@
...
+	"cosmossdk.io/errors"
+	"cosmossdk.io/math"
+	"github.com/allora-network/allora-chain/x/mint/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// set once at genesis and never changed after
+// 0.3675
+const EcosystemTreasuryPercentOfTotalSupply = 3675
+const EcosystemTreasuryPercentOfTotalSupplyPrecision = 4
```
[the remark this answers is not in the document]
relyt29: done

x/mint/keeper/emissions.go
```diff
@@ -0,0 +1,131 @@
...
+	"context"
+
+	"cosmossdk.io/errors"
+	"cosmossdk.io/math"
+	"github.com/allora-network/allora-chain/x/mint/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// return the uncirculating supply, i.e. tokens on a vesting schedule
+// latest discussion on how these tokens should be handled lives in ORA-1111
```
br4e, line 13: Unfortunately others who access only to the open source code won't have access to ORA-1111, so is there another way we can write an informative comment?

x/mint/keeper/emissions.go
```diff
@@ -0,0 +1,131 @@
...
+			types.ErrNegativeTargetEmissionPerToken,
+			"target emission per token is negative: %s | %s | %s",
+			ratioCirculating.String(),
+			ratioEcosystemToStaked.String(),
+			ret.String(),
+		)
+	}
+
+	return ret, nil
+}
```
br4e, line 98: its insane to me how much code there is for essentially
```
if (C == 0 || E == 0 || F == 0) throw 'division by zero'
result = (A * (B / C) * (D / E)) / F
if (result < 0) throw 'negative emissions'
return result
```

br4e: is there any way to simplify this code? or is go just super crufty

x/mint/keeper/genesis.go
```diff
@@ -8,11 +8,24 @@ import (
...
+		data.PreviousRewardEmissionPerUnitStakedToken,
+	)
+	if err != nil {
+		panic(err)
+	}
+
+	err = keeper.PreviousBlockEmission.Set(ctx, data.PreviousBlockEmission)
+	if err != nil {
+		panic(err)
+	}
```
br4e, line 26: also why are all these errors passed around like variables? can't we just panic as soon as there's an error instead of passing it around and checking error states everywhere? isn't it riskier to rely on remembering to check it? also it makes the code very cluttered

x/mint/keeper/genesis.go
```diff
@@ -21,15 +34,27 @@ func (keeper Keeper) InitGenesis(ctx context.Context, ak types.AccountKeeper, da
...
+	params, err := keeper.Params.Get(ctx)
 	if err != nil {
 		panic(err)
 	}
 
-	params, err := keeper.Params.Get(ctx)
+	previousRewardEmissionPerUnitStakedToken, err := keeper.PreviousRewardEmissionPerUnitStakedToken.Get(ctx)
+	if err != nil {
+		panic(err)
+	}
```
br4e, line 45: 🤯

I know you are following the pattern but the pattern blows my mind.

x/mint/keeper/genesis_test.go
```diff
@@ -52,42 +52,59 @@ func (s *GenesisTestSuite) SetupTest() {
...
+	s.keeper = keeper.NewKeeper(s.cdc, runtime.NewKVStoreService(key), stakingKeeper, accountKeeper, bankKeeper, emissionsKeeper, "", "")
 }
 
 func (s *GenesisTestSuite) TestImportExportGenesis() {
 	genesisState := types.DefaultGenesisState()
-	genesisState.Minter = types.NewMinter(math.LegacyNewDecWithPrec(20, 2), math.LegacyNewDec(1))
+	maxSupply, ok := math.NewIntFromString("1000000000000000000000000000")
+	if !ok {
+		panic("invalid number")
+	}
```
br4e, line 68: yay another pattern for handling failures

br4e: also like shouldn't this always either be ok or not ok?

x/mint/keeper/keeper.go
```diff
@@ -101,14 +106,65 @@ func (k Keeper) MintCoins(ctx context.Context, newCoins sdk.Coins) error {
...
+		types.EcosystemModuleName,
+		mintedCoins,
+	)
+}
+
+// PayCosmosValidatorRewardFromEcosystemAccount sends funds from the ecosystem
+// treasury account to the validator rewards account (fee collector)
+// PayCosmosValidatorRewardFromEcosystemAccount to be used in BeginBlocker.
+func (k Keeper) PayCosmosValidatorRewardFromEcosystemAccount(ctx context.Context, rewards sdk.Coins) error {
+	fmt.Println("PayCosmosValidatorRewardFromEcosystemAccount, paying: ", rewards)
```
br4e, line 129: we dont actually want print statements right, can we remove them all?

x/mint/keeper/msg_server_test.go
```diff
@@ -8,6 +8,10 @@ import (
 )
 
 func (s *IntegrationTestSuite) TestUpdateParams() {
+	maxSupply, ok := sdkmath.NewIntFromString("1000000000000000000000000000")
+	if !ok {
+		panic("invalid number")
+	}
```
br4e, line 14: same comment as before

x/mint/keeper/query_server.go
```diff
@@ -26,22 +27,29 @@ func (q queryServer) Params(ctx context.Context, _ *types.QueryParamsRequest) (*
...
 func (q queryServer) Inflation(ctx context.Context, _ *types.QueryInflationRequest) (*types.QueryInflationResponse, error) {
-	minter, err := q.k.Minter.Get(ctx)
+	// as a crude approximation we take the last blockEmission
+	// multiply by the amount of blocks in a year,
+	// then use that relative to the current supply as "inflation"
+	// Inflation Rate = ((B-A)/A) x 100
+	blockEmission, err := q.k.PreviousBlockEmission.Get(ctx)
 	if err != nil {
 		return nil, err
 	}
```
br4e, line 40: 😢

there must be a better way

x/mint/module/abci.go
```diff
@@ -2,80 +2,180 @@ package mint
...
+	ctx sdk.Context,
+	k keeper.Keeper,
+	params types.Params,
+	ecosystemBalance math.Int,
+) (
+	emissionPerTimestep math.Int,
+	emissionPerUnitStakedToken math.LegacyDec,
+	err error,
+) {
+	fmt.Println("Updating emission rate")
```
br4e, line 24: remove all the print statements?

x/mint/module/abci.go
```diff
@@ -2,80 +2,180 @@ package mint
...
+	// Get the expected amount of emissions this block
+	networkStaked, err := keeper.GetNumStakedTokens(ctx, k)
+	fmt.Println("Network staked", networkStaked)
+	if err != nil {
+		return math.Int{}, math.LegacyDec{}, err
+	}
```
br4e, line 30: 😢

x/mint/module/abci.go
```diff
@@ -2,80 +2,180 @@ package mint
...
+			lockedSupply.String(),
+		)
+	}
+	fmt.Println("Total supply", totalSupply)
+	fmt.Println("Locked supply", lockedSupply)
+	fmt.Println("Circulating supply", circulatingSupply)
+	fmt.Println("FEmissionNumerator", params.FEmissionNumerator)
+	fmt.Println("FEmissionDenominator", params.FEmissionDenominator)
+	targetRewardEmissionPerUnitStakedToken,
+		err := keeper.GetTargetRewardEmissionPerUnitStakedToken(
```
br4e, line 48: indentation is off

x/mint/module/abci.go
```diff
@@ -2,80 +2,180 @@ package mint
...
-
-	// Calculate the new total supply if coins were minted this block
-	newTotalCirculatingSupply := totalCirculatingSupply.Add(cosmosMath.Int(currentBlockProvision))
-
-	// Only mint new coins if new total supply would not exceed max supply
-	if newTotalCirculatingSupply.LTE(cosmosMath.Int(params.MaxSupply)) {
-		mintedCoin := sdk.NewCoin(params.MintDenom, cosmosMath.Int(currentBlockProvision))
-		mintedCoins := sdk.NewCoins(mintedCoin)
+	// find out if we need to update the block emissions rate
+	emissionRateUpdateCadence := params.BlocksPerMonth / params.EmissionCalibrationsTimestepPerMonth
```
br4e, line 98: bold! a raw division. does this need to be protected like all of the other divisions?

x/mint/keeper/emissions.go
dmikey: I'd probably rewrite it like, biggest goal of idiomatic golang is to reduce mental unrolling when we come back to it in 6 months.

```go
// CalculateTargetRewardEmission calculates the target reward emission per unit of staked token.
func CalculateTargetRewardEmission(
	emissionNumerator, emissionDenominator, ecosystemBalance,
	networkStaked, circulatingSupply, totalSupply math.Int,
) (math.LegacyDec, error) {
	if emissionDenominator.IsZero() || networkStaked.IsZero() || totalSupply.IsZero() {
		return math.LegacyDec{}, fmt.Errorf(
			"invalid input: denominator, network staked, or total supply is zero",
		)
	}

	// Calculate the ratios needed for the target emission calculation
	circulatingRatio := circulatingSupply.ToLegacyDec().Quo(totalSupply.ToLegacyDec())
	ecosystemStakedRatio := ecosystemBalance.ToLegacyDec().Quo(networkStaked.ToLegacyDec())

	// Calculate the target emission per unit of staked token
	targetEmission := emissionNumerator.ToLegacyDec().
		Mul(ecosystemStakedRatio).
		Mul(circulatingRatio).
		Quo(emissionDenominator.ToLegacyDec())

	if targetEmission.IsNegative() {
		return math.LegacyDec{}, fmt.Errorf(
			"negative target emission: circulating ratio %s, ecosystem to staked ratio %s, emission %s",
			circulatingRatio.String(), ecosystemStakedRatio.String(), targetEmission.String(),
		)
	}

	return targetEmission, nil
}
```

x/mint/keeper/emissions.go
relyt29: I hope that this comment will not survive the next 2 weeks and would be replaced before any outside eyes look at us seriously

that said, I can try to make this comment more informative now and remove the reference

x/mint/keeper/emissions.go
relyt29: yeah go is extremely verbose it sucks

x/mint/keeper/genesis.go
relyt29: I agree, I don't like this about golang

x/mint/keeper/keeper.go
relyt29: yes good catch

x/mint/keeper/query_server.go
relyt29: there isn't

x/mint/module/abci.go
relyt29: it should be auto indented by the go formatter

x/mint/module/abci.go
relyt29: I think it's fine, truncation is fine with me in general for this

x/mint/keeper/genesis.go
dmikey: another go convention. You pass errs back to the caller, so that deep yielding doesn't happen. no `try catch`

x/mint/module/abci.go
relyt29: we care about other divisions where we're dealing with decimal-like types or financial coin balances, but this is just to figure out how often to do something

relyt29:
I'm going to replace all the XNumerator XDenominator code with LegacyDec code, fuck it yolo we basically have to use the deprecated class anyways for other parts of the codebase at this point


if and when it becomes a problem we can move to cockroach DB's apd Decimal types like me and @kpeluso talked about, for now using legacyDec should be fine, I think (other cosmos-sdk modules like /x/staking are still using it soooooo)

br4e:
Thank you for entertaining my comments that were primarily concerned with go works. Ill take another pass today to get more into the details of what you wrote @relyt29

x/mint/keeper/emissions.go
```diff
@@ -0,0 +1,131 @@
...
+// just return the circulating supply based off of what the agreements off chain
+// were supposed to be at time of chain-genesis
+func GetLockedTokenSupply() math.Int {
+	return math.ZeroInt()
+}
+
+// helper function to get the number of staked tokens on the network
+// includes both tokens staked by cosmos validators (cosmos staking)
+// and tokens staked by reputers (allora staking)
+func GetNumStakedTokens(ctx context.Context, k Keeper) (math.Int, error) {
```
dmikey, line 24: are both of these tokens goveren'd by the same bonding time? Might be worth while noting the state of the token?

scripts/init.sh
```diff
@@ -21,6 +21,8 @@ $ALLORAD_BIN genesis add-genesis-account allo1q4fa4tqzng2lshfhjaklx90hzfnfennxt0
 $ALLORAD_BIN genesis add-genesis-account allo1zy5akp9grwfp3x6rqd40x0g4agpzjaskxr9lnn 10000000allo
 $ALLORAD_BIN genesis add-genesis-account allo1ywhj2svg67mn7ylr9mu5kz9f668z2xejnp9w9y 10000000allo
 $ALLORAD_BIN genesis add-genesis-account allo1r7hqeqdmf6jg9v9px0gh5l6n7tlr0tlxt86plc 10000000allo
+# Ecosystem account gets 36.75% of total supply
+$ALLORAD_BIN genesis add-genesis-account allo12uxa8rw9hte3z2nuswjzpmfen289n30uag6agp 367500000allo
```
dmikey, line 25: Might be more of a business question, but the optics around this are kind bad IMO. These should be put into a vesting account that unlocks community funds over time. Tossing 1/3 into a org controlled wallet will bring a lot of FUD. Maybe this is just pre production?

dmikey (approved):
Looks sound to me, I don't like many of the variables but a lot of this is math, and so being true to the equations probably matters more over the long term, than verbosity in descriptive names.

scripts/init.sh
relyt29: a) this is a testing environment
b) this address is not human controlled, it's the address of the module account which is exactly the vesting account you described

x/mint/keeper/emissions.go
relyt29: I don't exactly know what you mean by bonding time in this context, but I do believe they have different withdrawal delay periods if that's what you're asking

I feel like making any decision about that is more of a paper question rather than code - i.e. go convince diederik

x/mint/keeper/msg_server_test.go
relyt29: same response as before

scripts/init.sh
dmikey: thanks for the clarification. Is this key constant, shouldn't this be in the modules genesis initializer?

x/mint/module/abci.go
```diff
@@ -2,80 +2,176 @@ package mint
...
 		if err != nil {
 			return err
 		}
 	}
-
+	// pay out the computed block emissions from the ecosystem account
+	// if it came from collected fees, great, if it came from minting, also fine
+	// we pay both reputers and cosmos validators, so each payment should be
+	// half as big (divide by two). Integer division truncates, and that's fine.
+	coins := sdk.NewCoins(sdk.NewCoin(params.MintDenom, blockEmission.Quo(math.NewInt(2))))
```
kpeluso, line 161: global param

actually default it to 1/4 (so, 4)

kpeluso: see: `PercentRewardsReputersWorkers`

x/mint/module/abci.go
```diff
@@ -2,80 +2,176 @@ package mint
...
+	fmt.Println("Paying coins", coins)
+	err = k.PayCosmosValidatorRewardFromEcosystemAccount(sdkCtx, coins)
+	if err != nil {
+		return err
+	}
+	err = k.PayReputerRewardFromEcosystemAccount(sdkCtx, coins)
```
kpeluso, line 167: rename: `PayEmissionsModule` cuz workers still being paid from this too

kpeluso (changes requested):

x/emissions/keeper/keeper.go
```diff
@@ -932,7 +932,7 @@ func (k *Keeper) RemoveDelegatedStake(
 }
 
 // Gets the total sum of all stake in the network across all topics
-func (k *Keeper) GetTotalStake(ctx context.Context) (Uint, error) {
+func (k Keeper) GetTotalStake(ctx context.Context) (Uint, error) {
```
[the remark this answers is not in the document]
relyt29:
```
cannot use emissionsKeeper (variable of type "github.com/allora-network/allora-chain/x/emissions/keeper".Keeper) as "github.com/allora-network/allora-chain/x/mint/types".EmissionsKeeper value in argument to keeper.NewKeeper: "github.com/allora-network/allora-chain/x/emissions/keeper".Keeper does not implement "github.com/allora-network/allora-chain/x/mint/types".EmissionsKeeper (method GetTotalStake has pointer receiver)compiler[InvalidIfaceAssign](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#InvalidIfaceAssign)
```

some weird shit with interfaces in golang and pointers

x/mint/module/abci.go
```diff
@@ -2,80 +2,176 @@ package mint
...
 		if err != nil {
 			return err
 		}
-
-		// Halving logic: Check if the current block height is a multiple of the halving interval.
-		if sdkCtx.BlockHeight()%int64(params.HalvingInterval) == 0 {
-			currentBlockProvision = currentBlockProvision.QuoUint64(2)
-			params.CurrentBlockProvision = currentBlockProvision
-			minter.AnnualProvisions = minter.AnnualProvisions.QuoInt64(2)
+		ecosystemMaxSupply := math.LegacyNewDecFromInt(params.MaxSupply).
```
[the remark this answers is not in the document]
relyt29: i think this code got deleted after more changes, resolving

scripts/init.sh
relyt29: no i think it belongs to the process making the genesis.json

also this specific line of code deleted after some discussion with diederik

dmikey: > no i think it belongs to the process making the genesis.json
> 
> also this specific line of code deleted after some discussion with diederik

Hmm, right I thought there is a genesis initializer for each module?

x/mint/module/abci.go
```diff
@@ -2,80 +2,180 @@ package mint
...
-
-	// Only mint new coins if new total supply would not exceed max supply
-	if newTotalCirculatingSupply.LTE(cosmosMath.Int(params.MaxSupply)) {
-		mintedCoin := sdk.NewCoin(params.MintDenom, cosmosMath.Int(currentBlockProvision))
-		mintedCoins := sdk.NewCoins(mintedCoin)
+	// find out if we need to update the block emissions rate
+	emissionRateUpdateCadence := params.BlocksPerMonth / params.EmissionCalibrationsTimestepPerMonth
+	if emissionRateUpdateCadence == 0 {
+		return errors.Wrapf(
+			types.ErrZeroDenominator,
```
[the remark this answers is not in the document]
relyt29: i'm removing the error. if you want to never update the emission rate, that should be permitted

relyt29:
observing 45% inflation in the first few blocks with parameters tweaked where they are

relyt29:
2024-04-06-161315_181x55_scrot






2024-04-06-161434_411x304_scrot

PR #96, opened 2024-03-28
[WIP] Worker Rewards Integration

guilherme-brandao:
Adds two new functions that will be used to calculate the amount of rewards that should be paid to each worker for inference task and forecast task.


x/emissions/module/worker_rewards.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	// Convert scores to float64
+	var scoresFloat64 [][]float64
+	for _, workerScoreArray := range scores {
+		var scoresOneOutFloat64 []float64
+		for _, score := range workerScoreArray {
+			scoresOneOutFloat64 = append(scoresOneOutFloat64, score.Score)
+		}
+		scoresFloat64 = append(scoresFloat64, scoresOneOutFloat64)
```
kpeluso, line 121: can this be done in the loop above to save an extra loop?

x/emissions/module/worker_rewards.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// Get Worker Reward Fractions
+	workersRewardsFractions, err := GetWorkerRewardFractions(scoresFloat64, preward)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get Worker Rewards
+	var workerRewards []TaskRewards
+	for i, rewardFraction := range workersRewardsFractions {
```
kpeluso, line 67: if you pass in the total rewards into the function `GetWorkerRewardFractions` then you could save a whole loop. True here and in the function below. Function should probs then be renamed to `GetWorkerPortionOfRewards` or something

x/emissions/module/rewards_internal.go
```diff
@@ -0,0 +1,795 @@
...
+	}
+
+	stdDev := StdDev(flatten(lastScores))
+	var normalizedScores []float64
+	for _, score := range lastScores {
+		normalizedScores = append(normalizedScores, score[len(score)-1]/stdDev)
+	}
+
+	smoothedScores := make([]float64, len(normalizedScores))
+	for i, v := range normalizedScores {
```
kpeluso, line 55: this for-loop and the one above it can easily be combined.

x/emissions/module/rewards_internal.go
```diff
@@ -0,0 +1,795 @@
...
+		res, err := Phi(preward, v)
+		if err != nil {
+			return nil, err
+		}
+		smoothedScores[i] = res
+	}
+
+	total := 0.0
+	for _, score := range smoothedScores {
```
kpeluso, line 64: the body of this loop could be computed as part of the body of the 2 loops above

x/emissions/keeper/keeper.go
```diff
@@ -57,6 +57,14 @@ type Keeper struct {
...
 	addressTopics collections.Map[sdk.AccAddress, []uint64]
 
+	/// SCORES
+	// map of (topic, block_number, worker) -> score
+	workerInferenceScores collections.Map[collections.Pair[TOPIC_ID, BLOCK_NUMBER], types.Scores]
+	// map of (topic, block_number, worker) -> score
+	workerForecastScores collections.Map[collections.Pair[TOPIC_ID, BLOCK_NUMBER], types.Scores]
+	// map of (topic, block_number, reputer) -> score
+	reputerScores collections.Map[collections.Pair[TOPIC_ID, BLOCK_NUMBER], types.Scores]
+
```
relyt29, line 67: maybe we can save some letters and just call this inferenceScores, forecastScores, reputerScores?  

or not, just seems a bit verbose

x/emissions/module/reputer_rewards.go
```diff
@@ -0,0 +1 @@
+package module
```
relyt29, line 1: maybe at this point it would be appropriate to put all things "rewards" in its own folder

x/emissions/keeper/keeper.go
```diff
@@ -1693,6 +1710,115 @@ func (k *Keeper) ResetNumInferencesInRewardEpoch(ctx context.Context) error {
...
+		} else {
+			return err
+		}
+	}
+	scores.Scores = append(scores.Scores, &score)
+
+	return k.reputerScores.Set(ctx, key, scores)
+}
+
+func (k *Keeper) GetLastWorkerInferenceScores(ctx context.Context, topicId TOPIC_ID, blockNumber BLOCK_NUMBER, worker WORKER) ([]*types.Score, error) {
```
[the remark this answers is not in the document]
guilherme-brandao: I think `UntilBlock` will be better since it's not just the rewards at this specific block, but the last 10.

relyt29:
I would like to merge this

relyt29 (approved):

PR #98, opened 2024-03-31
add eq14

kpeluso:
All network losses as per addendum doc (below) and fix some earlier in Section 3 equation implementations.

image


x/emissions/proto/emissions/v1/types.proto
```diff
@@ -98,18 +113,22 @@ message ValueBundle {
...
   double naive_value = 6;
-  // L^-_lim or I^-_li
-  repeated WorkerAttributedValue one_out_values = 7;
-  // L^+_kim or I^+_ki
-  repeated WorkerAttributedValue one_in_naive_values = 8;
+  // R^-_ilkm || log10 L^-_ilkm || I^-_ilk where l = any j
+  // An extra index is necessary because forecast-implied inferences are recomputed for each left-out inference
+  repeated WorkerAttributedValuesOfWithheldWorker one_out_inferer_values = 7;
+  // R^-_ilm || log10 L^-_ilm || I^-_il where l = any k
+  repeated WithheldWorkerAttributedValue one_out_forecaster_values = 8;
```
guilherme-brandao, line 128: What `L^-_ilkm` mean? 

and I guess it's the `one_out_forecaster_values` that needs to have `WorkerAttributedValuesOfWithheldWorker` type.

guilherme-brandao: Also, considering diederik comment in the gc: 
```
reputer combinator loss oneout
L^-_lim where l = some j or k (l goes from 1 to n_inferences + n_forecasts,  j goes from 1 to n_inferences , and k from 1 to n_forecasts)
````

Prob we should use a new type for `one_out_inferer_values` since it will be an array that includes both withheld workers inference losses and forecast losses. Perhaps it should be named `one_out_values`.

guilherme-brandao (changes requested):

x/emissions/proto/emissions/v1/types.proto
kpeluso: talked ab this on call. remedied in later commit.

PR #104, opened 2024-04-02
Reputer Rewards Integration

guilherme-brandao:
Implement the flow to calculate reputer rewards, generate the exact values that will be sent to reputer's account.


x/emissions/module/rewards/reputer_rewards_test.go
```diff
@@ -0,0 +1,95 @@
...
+		1,
+		1017.5559072418691,
+	)
+	s.Require().NoError(err)
+	s.Require().Equal(5, len(reputerRewards))
+
+	expectedRewards := []float64{456.49, 172.71, 211.93, 52.31, 124.10}
+
+	for i, reputerReward := range reputerRewards {
+		if math.Abs(reputerReward.Reward-expectedRewards[i]) > 1e-2 {
```
br4e, line 35: can we tighten this epsilon value at all? a 1% difference is pretty significant

br4e (approved):
This looks reasonable to me, it seems to be correctly working towards 
"reputer receives rewards proportional both to its stake and the consensus between its evaluations and those of other reputers"
and
"As discussed above, reputers require scoring according to their consensus in reporting the ground truth. The naive way of doing this is to calculate the stake-weighted average of all losses reported by a reputer, and to add the rewards to their stakes. However, this creates a runaway effect towards increased centralization, where the reputer with the highest stake has the largest weight in setting the consensus, thereby receiving the highest rewards and further increasing their stake advantage. This can be remedied by using an adjusted stake to set the weight of each reputer when calculating the consensus, where the weight saturates above a certain fraction of the stake. Specifically,"

where did you get the numbers you used in your test cases?

x/emissions/module/rewards/reputer_rewards.go
```diff
@@ -1 +1,72 @@
...
+)
+
+// The reputer rewards are calculated based on the reputer stake and the reputer score.
+// The reputer score is defined right after the network loss is generated.
+func GetReputerRewards(
+	ctx sdk.Context,
+	keeper keeper.Keeper,
+	topicId uint64,
+	block int64,
+	preward float64,
```
br4e, line 15: is it possible to name this something more descriptive?

br4e: or is it generally known what this is?

kpeluso (approved):

PR #106, opened 2024-04-02
Added scores generation flow

guilherme-brandao:
These functions will be used right after a lead reputer insert all loss bundles and the network loss is generated.

### DIFF
--- a/x/emissions/module/rewards/scores.go
+++ b/x/emissions/module/rewards/scores.go
@@ -1,0 +1,133 @@
+package rewards
+
+import (
+	"github.com/allora-network/allora-chain/x/emissions/keeper"
+	"github.com/allora-network/allora-chain/x/emissions/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func GenerateReputerScores(ctx sdk.Context, keeper keeper.Keeper, topicId uint64, block int64, reportedLosses types.ReputerValueBundles) ([]types.Score, error) {
+	// Get reputers informations
+	var reputerAddresses []sdk.AccAddress
+	var reputerStakes []float64
+	var reputerListeningCoefficients []float64
+	var losses [][]float64
+	for _, reportedLoss := range reportedLosses.ReputerValueBundles {
+		reputerAddr, err := sdk.AccAddressFromBech32(reportedLoss.Reputer)
+		if err != nil {
+			return []types.Score{}, err
+		}
+		reputerAddresses = append(reputerAddresses, reputerAddr)
+
+		// Get reputer topic stake
+		reputerStake, err := keeper.GetStakeOnTopicFromReputer(ctx, topicId, reputerAddr)
+		if err != nil {
+			return []types.Score{}, err
+		}
+		reputerStakes = append(reputerStakes, float64(reputerStake.BigInt().Int64()))
+
+		// Get reputer listening coefficient
+		res, err := keeper.GetListeningCoefficient(ctx, topicId, reputerAddr)
+		if err != nil {
+			return []types.Score{}, err
+		}
+		reputerListeningCoefficients = append(reputerListeningCoefficients, res.Coefficient)
+
+		// Get all reported losses from bundle
+		reputerLosses := ExtractValues(reportedLoss.ValueBundle)
+		losses = append(losses, reputerLosses)
+	}
+
+	// Get reputer output
+	scores, newCoefficients, err := GetAllReputersOutput(losses, reputerStakes, reputerListeningCoefficients, len(reputerStakes))
+	if err != nil {
+		return []types.Score{}, err
+	}
+
+	// Insert new coeffients and scores
+	var newScores []types.Score
+	for i, reputerAddr := range reputerAddresses {
+		err := keeper.SetListeningCoefficient(ctx, topicId, reputerAddr, types.ListeningCoefficient{Coefficient: newCoefficients[i]})
+		if err != nil {
+			return []types.Score{}, err
+		}
+
+		newScore := types.Score{
+			TopicId:     topicId,
+			BlockNumber: block,
+			Address:     reputerAddr.String(),
+			Score:       scores[i],
+		}
+		err = keeper.InsertReputerScore(ctx, topicId, block, newScore)
+		if err != nil {
+			return []types.Score{}, err
+		}
+		newScores = append(newScores, newScore)
+	}
+
+	return newScores, nil
+}
+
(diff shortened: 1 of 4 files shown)


x/emissions/module/rewards/scores.go
```diff
@@ -0,0 +1,134 @@
+package rewards
+
+import (
+	"github.com/allora-network/allora-chain/x/emissions/keeper"
+	"github.com/allora-network/allora-chain/x/emissions/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func GenerateReputerScores(ctx sdk.Context, keeper keeper.Keeper, topicId uint64, block int64, reportedLosses types.ReputerValueBundles) ([]types.Score, error) {
```
br4e, line 9: Where is this used? It's not being called anywhere besides in the test

x/emissions/module/rewards/rewards_internal_test.go
```diff
@@ -964,7 +960,7 @@ func TestGetAllReputersOutput(t *testing.T) {
 				return
 			}
 
-			if !slicesAreApproxEqual(gotScores, tt.wantScores, 1e-4) {
+			if !slicesAreApproxEqual(gotScores, tt.wantScores, 1e-1) {
```
br4e, line 963: Doesn't this just allow any difference between the two scores? Or does this mean that any score within 100% of the correct score is accepted? 

Both feel not ideal to me

x/emissions/module/rewards/scores.go
```diff
@@ -0,0 +1,134 @@
...
+		err = keeper.InsertWorkerInferenceScore(ctx, topicId, block, newScore)
+		if err != nil {
+			return []types.Score{}, err
+		}
+		newScores = append(newScores, newScore)
+	}
+	return newScores, nil
+}
+
+func GenerateForecastScores(ctx sdk.Context, keeper keeper.Keeper, topicId uint64, block int64, networkLosses types.ValueBundle) ([]types.Score, error) {
```
br4e, line 97: same for all of these functions, will they be used in a future pr?

x/emissions/module/rewards/scores_test.go
```diff
@@ -0,0 +1,191 @@
...
+	scores, err := rewards.GenerateReputerScores(
+		s.ctx,
+		s.emissionsKeeper,
+		topidId,
+		block,
+		reportedLosses,
+	)
+	s.Require().NoError(err)
+
+	expectedScores := []float64{17.98648, 20.32339, 26.44637, 11.17804, 14.93222}
```
br4e, line 30: how did you calculate the expected scores? For these tests to be meaningful the scores would need to be calculated outside of the context of the tests, using the expected whitepaper behavior

br4e (commented):
its difficult to determine if the math exactly matches what is found in the litepaper given the complexity of the math here and in the litepaper. That's why it's so important to understand where you are getting the expected test values, if we are seeing the correct end to end values according to the whitepaper logic then we should be good to go. 

Also will there be another pr to use the functions introduced here?

x/emissions/module/rewards/rewards_internal_test.go
guilherme-brandao: For context, I’m using Diederik’s Python simulation values as a reference for I/O. In this particular case, he is using some values in the simulation that we will not be using anymore to calculate the consensus score (he clarified this in our last conversation), which leads to slightly different expected outputs. In ‘rewards_internal_test.go,’ I used all the values from the simulator in the isolated GetAllConsensusScores function and was able to get a more similar output. Probably in this particular case, I can use different expected values.

x/emissions/module/rewards/scores.go
guilherme-brandao: It will be attached to @kpeluso work on the losses input. The scores will be generated right after network losses generation.

x/emissions/module/rewards/scores.go
```diff
@@ -0,0 +1,144 @@
...
+
+/*
+ These functions will be used immediately after the network loss for the relevant time step has been generated.
+ Using the network loss and the sets of losses reported by each repeater, the scores are calculated. In the case
+ of workers (who perform the forecast task and network task), the last 10 previous scores will also be taken into
+ consideration to generate the score at the most recent time step.
+*/
+
+// GenerateReputerScores calculates and persists scores for reputers based on their reported losses.
+func GenerateReputerScores(ctx sdk.Context, keeper keeper.Keeper, topicId uint64, block int64, reportedLosses types.ReputerValueBundles) ([]types.Score, error) {
```
br4e, line 17: looks like this corresponds to equation 27 in the whitepaper?

br4e (approved):
Given that you are thoroughly testing against the specific simulation results provided by diederik, and that these are unit tested functions that will be tested again when they are integrated, and looking through the code myself and not finding issues, I believe in this! 

Awesome work. 

If there is anything specific in this pr that you want me to focus some more time on, let me know.

x/emissions/module/rewards/scores.go
```diff
@@ -0,0 +1,144 @@
...
+
+		// Get reputer listening coefficient
+		res, err := keeper.GetListeningCoefficient(ctx, topicId, reputerAddr)
+		if err != nil {
+			return []types.Score{}, err
+		}
+		reputerListeningCoefficients = append(reputerListeningCoefficients, res.Coefficient)
+
+		// Get all reported losses from bundle
+		reputerLosses := ExtractValues(reportedLoss.ValueBundle)
```
kpeluso, line 45: generally i think you're making it harder for yourself to debug when you index by "index" as opposed to reputer address in this case

also looking at `GetStakeWeightedLossMatrix()`, which is downstream of `GetAllReputersOutput` invoked here... i see you're not dealing with the case where reputers report on a different set of workers with potentially a different distribution regarding "who sent inference and/or forecast"... right now you're assuming they're all reporting on the same thing...

surely our own reputers will eb able to deal with this and always report on a uniform set, but we're just opening the door for malicious actors if we're not taking that into account -- cuz obvi we won't be the only people running reputers

kpeluso (changes requested):
found one critical thing, will continue reviewing after that is fixed.

kpeluso (approved):
the use of arrays overall isn't my favorite cuz it makes it unclear what values belong to whom, but to balance expediency, it'll do. Furthermore, we'll be adding checks that come before this code is invoked, which will check that all values come from valid reputers, that their reports are generated from valid workers, that signatures are verified, and that a stake-weighted vote is done to connclude which workers should actually have their losses scored

PR #109, opened 2024-04-03
[ORA-1053] Test eq 3-8, 9-11, 12, 14

br4e:
These are test harnesses to test the code for calculating most of the equations from the litepaper. This will be ready for review once we have actual values to test derived from the equations in the litepaper.


br4e:
closing in favor of https://github.com/allora-network/allora-chain/pull/111

PR #113, opened 2024-04-07
[ORA-1092] Decide how much reward each topic gets given sum of rewards

relyt29:
Implements paper section 5.2: deciding how much rewards each topic should get


x/emissions/keeper/keeper.go
```diff
@@ -550,6 +586,24 @@ func (k *Keeper) GetLatestForecastsFromTopic(ctx context.Context, topicId TOPIC_
 	return forecasts, nil
 }
 
+// get the previous weight during rewards calculation for a topic
+func (k *Keeper) GetPreviousTopicWeight(ctx context.Context, topicId TOPIC_ID) (types.PreviousTopicWeight, error) {
+	topicWeight, err := k.previousTopicWeight.Get(ctx, topicId)
+	if errors.Is(err, collections.ErrNotFound) {
+		ret := types.PreviousTopicWeight{
```
dmikey, line 593: could just return the value?

```
if errors.Is(err, collections.ErrNotFound) {
    return types.PreviousTopicWeight{Weight: 0.0, Epoch: 0}, nil
}
```

x/mint/proto/mint/v1beta1/types.proto
```diff
@@ -22,67 +22,57 @@ message Params {
...
-  string validators_vs_allora_percent_reward = 4 [
-    (cosmos_proto.scalar) = "cosmos.LegacyDec",
-    (gogoproto.customtype) = "cosmossdk.io/math.LegacyDec",
-    (gogoproto.nullable) = false,
-    (amino.dont_omitempty) = true
-  ];
-
   // maximum total supply of the coin
-  string max_supply = 5 [
+  string max_supply = 4 [
```
dmikey, line 26: we should be careful about re-indexing, this would prob cause a state halt, requiring us to do something like

"allorad version 0.0.1 until block 54. then switch clients and continue to sync until"

x/emissions/keeper/keeper.go
relyt29: Don't know if the value is initialized to zero correctly, this is safer

x/mint/proto/mint/v1beta1/types.proto
relyt29: Good to know 

I think this is a major change and will require a new genesis for the major-upgrade PR but after that I will stop reindexing

kpeluso: we def were gonna have to do that anyway so you're clear to do that up until monday

kpeluso (approved):

PR #116, opened 2024-04-08
[ORA-1166] Migrate float64 type to CockroachDB Arbitrary Precision Decimal type

relyt29:
float64 are handled differently on different CPU architecture so we cannot use it for consensus when we expect different types of CPU machines to come to consensus e.g. x86 and ARM

Fortunately underlying cockroachDB that cosmos-sdk already relies on provides a complex arbitrary-precision decimal package for go called `Dec.`

This PR migrates us to use `Dec` in place of float64.


relyt29:
We can send the APD Decs through the CLI:
Help text
```
➜  allora-chain git:(APD-Dec) ✗ allorad tx emissions test-dec-message --help
test sending a dec message via the cli

Usage:
  allorad tx emissions test-dec-message [value] [flags]
```
Sending the transaction
```
➜  allora-chain git:(APD-Dec) ✗ allorad tx emissions test-dec-message "allo17e5ge56zvzcrw0pehv0y7hyuakwwv7ypyxk7j9" "0.
12345"                                             
auth_info: 
  fee:       
    amount: []
    gas_limit: "200000"                                                                                                
    granter: ""                                                                                                        
    payer: ""
  signer_infos: []
  tip: null
body:
  extension_options: []
  memo: ""
  messages:
  - '@type': /emissions.v1.DecMessage
    sender: allo17e5ge56zvzcrw0pehv0y7hyuakwwv7ypyxk7j9
    value: "0.12345"
  non_critical_extension_options: []
  timeout_height: "0"
signatures: []
confirm transaction before signing and broadcasting [y/N]: y
code: 0
codespace: ""
data: ""
events: []
gas_used: "0"
gas_wanted: "0"
height: "0"
info: ""
logs: []
raw_log: ""
timestamp: ""
tx: null
txhash: 8C66276C1DED0DA8BF9C595EC9982D4EF56BAA6B574BACCD84329D932E2EB889
```
In another tab where the CLI is running we have a simple printf statement in the msg_handler.go:

```
11:43AM INF committed state block_app_hash=A7D971D27E177737E28E8AB87B631E6DD45F0751120B4530DB19739C3BF1F954 height=38 module=state

 ---------------- Emissions BeginBlock ------------------- 
TestDecMessage
0.12345


 ---------------- Emissions EndBlock ------------------- 
```

RedBird96:
From now on, we should use this `Dec` instead of `float64`, right?

relyt29:
> From now on, we should use this `Dec` instead of `float64`, right?

yes

relyt29:
this is as good as this is going to get for now

x/emissions/module/rewards/rewards_internal.go
```diff
@@ -229,129 +443,233 @@ func GetAllConsensusScores(allLosses [][]float64, stakes []float64, allListening
...
+	numReputers int64,
+) ([]alloraMath.Dec, []alloraMath.Dec, error) {
 	learningRate := types.DefaultParamsLearningRate()
-	coefficients := make([]float64, len(initialCoefficients))
+	coefficients := make([]alloraMath.Dec, len(initialCoefficients))
 	copy(coefficients, initialCoefficients)
 
-	oldCoefficients := make([]float64, numReputers)
-	maxGradientThreshold := 0.001
-	imax := int(math.Round(1.0 / learningRate))
```
relyt29, line 239: @kenny can you double check this code

kpeluso (approved):

PR #118, opened 2024-04-09
Michaelc/ora 1130 set tax on workers to be a function of average rewards and

RedBird96:
add tax for only worker reward
[ORA-1130]([ORA-1130 : Set tax on workers to be a function of average rewards and parameters](https://linear.app/upshot/issue/ORA-1130/set-tax-on-workers-to-be-a-function-of-average-rewards-and-parameters))


dmikey:
so many unrelated commits, makes me sad.

RedBird96:
> so many unrelated commits, makes me sad.

Please review only last commit...

x/emissions/proto/emissions/v1/types.proto
```diff
@@ -269,6 +271,13 @@ message Scores {
   repeated Score scores = 1;
 }
 
+/// AVERAGE REWARD
+
+message AvgReward {
```
kpeluso, line 276: should be consistent -- you say "AverageReward" everywhere but here, where you use shorthand

x/emissions/types/keys.go
```diff
@@ -49,7 +49,8 @@ var (
 	InferenceScoresKey             = collections.NewPrefix(38)
 	ForecastScoresKey              = collections.NewPrefix(39)
 	ReputerListeningCoefficientKey = collections.NewPrefix(40)
-	FeeRevenueEpochKey             = collections.NewPrefix(41)
-	TopicFeeRevenueKey             = collections.NewPrefix(42)
-	PreviousTopicWeightKey         = collections.NewPrefix(43)
+	AverageRewardKey               = collections.NewPrefix(41)
```
kpeluso, line 52: There are a few actors for which an avg reward can be calculated. this is for workers tho, so should be specified

kpeluso (changes requested):
this is missing 1 key thing per the ticket:
> These funds should be sent to the ecosystem treasury, where all the inference request fees are already paid today.

kpeluso (approved):
nvm actually... because you're subtracting the tax here, then the rewards will never be sent from the ecosystem bucket => it's equivalent to doing what I said above, just simpler!

great!

x/emissions/proto/emissions/v1/types.proto
kpeluso: fixed this myself

x/emissions/types/keys.go
kpeluso: fixed this myself

PR #123, opened 2024-04-10
Michaelc/ora 1128 update topic registration to cost a fixed global fee that is

RedBird96:
create topic fee
[ORA-1128] https://linear.app/upshot/issue/ORA-1128/update-topic-registration-to-cost-a-fixed-global-fee-that-is-sent-to


x/emissions/types/keys.go
```diff
@@ -6,6 +6,7 @@ const ModuleName = "emissions"
 const AlloraStakingAccountName = "allorastaking"
 const AlloraRequestsAccountName = "allorarequests"
 const AlloraRewardsAccountName = "allorarewards"
+const EcosystemModuleName = "ecosystem"
```
kpeluso, line 9: @relyt29 odd we didn't have this already...

kpeluso (changes requested):
Some extra places you have to add the global param. Will make a post in slack because everyone should be aware

kpeluso (approved):

PR #127, opened 2024-04-11
CalcOneOutInferences bug fixes and tests

br4e:
Take a look. In addition to passing tests, there are a couple of significant architectural updates. 

1) ignore forecasts that don't have a corresponding inference in `CalcForcastImpliedInferences` to enable one-out inferences without messing up the max values
2) remove `forecastImpliedInferenceByWorker ` as an argument from `CalcOneOutInferences` since this value can instead be calculated with the `inferenceByWorker` and `forecasts` variables


br4e:
Fix merge conflicts

x/emissions/keeper/inference_synthesis/forecast_implied_inferences.go
```diff
@@ -74,7 +74,14 @@ func CalcForcastImpliedInferences(
 			// Map inferer -> forecast element => only one (latest in array) forecast element per inferer
 			forecastElementsByInferer := make(map[Worker]*emissions.ForecastElement, 0)
 			for _, el := range forecast.ForecastElements {
-				forecastElementsByInferer[el.Inferer] = el
+				for worker := range inferenceByWorker {
```
kpeluso, line 77: was thinking it'd solved when the payload of inferences is committed to the chain but think you're right -- it's safer to have these checks anyway

x/emissions/keeper/inference_synthesis/network_inferences.go
```diff
@@ -346,7 +392,7 @@ func CalcNetworkInferences(
 	}
 
 	// Calculate the one-out inference I^-_li
-	oneOutInferences, oneOutImpliedInferences, err := CalcOneOutInferences(ctx, k, topicId, inferenceByWorker, forecastImpliedInferenceByWorker, forecasts, maxCombinedRegret, networkCombinedLoss, epsilon, pInferenceSynthesis)
+	oneOutInferences, oneOutImpliedInferences, err := CalcOneOutInferences(ctx, k, topicId, inferenceByWorker, forecasts, maxCombinedRegret, networkCombinedLoss, epsilon, pInferenceSynthesis)
```
kpeluso, line 395: I don't agree with removing the forecastImpliedInference argument for this reason -- computation.

It's already computed and in memory => should pass it in. Yes it could be recreated, but it shouldn't be if it already exists.

Maybe this admits slightly less sexy code, but this is an expensive computation as far as on-chain computations go, so I think we need to be conservative.

kpeluso (changes requested):

kpeluso (approved):

PR #130, opened 2024-04-11
Tobi/ora 1141 reset whitelisted account and save properly on 1password

okedeji:
- Update to all whitelisted accounts. These are new ones now and they are saved both in SSM and 1Password in the infra vault
- added telnet just so we can use for debug when needed


kpeluso (changes requested):
spoke on slack

PR #135, opened 2024-04-12
Test coverage viewer

br4e:
Create a new test coverage tool that displays data in an easy-to-parse format (lcov.info). lcov.info is not native to go. The native go coverage viewer is not good. An external tool is used to convert the go coverage file to lcov to enable better display: `gcov2lcov` which can be installed with 
`go install github.com/jandelgado/gcov2lcov@latest`

Screenshot 2024-04-12 at 12 21 13 PM

I had to comment out a bunch of failing tests, next step is to get those tests back online. 

After removing generated files and libraries, we sit at about 26% test coverage (!!!)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ go.work
 data/
 # Account info
 *.account_info
+
+# coverage
+coverage
--- a/scripts/coverage.sh
+++ b/scripts/coverage.sh
@@ -1,0 +1,35 @@
+# SPDX-License-Identifier: AGPL-3.0
+
+# if any command fails then immediately exit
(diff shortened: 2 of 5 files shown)


relyt29 (approved):
I can confirm that after installing the package I was also able to view our atrociously bad coverage

br4e:
Ok guys, this is becoming a mind-fuck to deal with these broken tests in this pr, so I'm going to land this pr and then figure out the best way to get to 90% coverage whether that is fixing the commented out tests or not. I need to move forward instead of focusing just on these tests.

PR #137, opened 2024-04-12
Michaelc/ora 1167 ensure nonces and registration are checked on chain before

RedBird96:


dmikey (approved):

x/emissions/proto/emissions/v1/types.proto
```diff
@@ -244,6 +246,7 @@ message ForecastElement {
       [(gogoproto.customtype) = "github.com/allora-network/allora-chain/math.Dec", (gogoproto.nullable) = false];
   bytes extra_data = 4;
   string proof = 5;
+  bytes signature = 6;
```
kpeluso, line 249: Just to ensure we're on the same page:
Each forecast is a set of predictions => each forecaster is making a `Forecast` which contains a list of these signature elements.

Therefore, could we not have this signature on the Forecast instead of forecast element?
If so => way fewer signatures have to be checked

Or, is there some technical limitation preventing that?

kpeluso (changes requested):
1. The signature on each ForecastElement should be moved to the Forecast itself.
    a. This is because the Forecaster makes all Forecast elements at once. The b7s leader then takes all signed Forecasts (which each contain forecast elements for all inferers), and puts that on chain => just the signature of each Forecast (not element) needs to be checked.
2. You need to sign the full inference or forecast value (looks like you got this covered for reputer reports already)
    a. This is because if I sign just the value, and that's all that's checked, then someone can use my inference later, when perhaps it's opportune for them to do so, and essentially say that I said the inference was some value at the wrong time. The protocol would be none the wiser. => the full payload from each actor needs to be signed.
    b. Side note related to but not technically part of this PR: You need to verify that the way the off-chain node signs data exactly matches how it's done here.
3. Need to pull in `major-upgrade`.
    a. When you do, look for my "todo" comments pointing you to where to put this additional filter (the signature checks)

kpeluso (approved):

PR #139, opened 2024-04-13
Keeper Part 0: Initial tests

br4e:
There were a few bugs in the keeper that I have been testing and fixing

Comment out failing tests and introduce the "TODO_RESTORE_TESTS @handle" pattern so people can fix their tests. This is much more efficient than me trying to figure out wth people were trying to test and fix it for them, I can go focus on writing tests that don't exist at all.


x/emissions/keeper/keeper.go
```diff
@@ -754,6 +775,7 @@ func (k *Keeper) InsertForecasts(ctx context.Context, topicId TopicId, nonce typ
 		if err != nil {
 			return err
 		}
+		// TODO
```
kpeluso, line 778: nope

PR #154, opened 2024-04-16
Fix TestMsgInsertBulkReputerPayload

br4e:


x/emissions/keeper/msgserver/msg_server_losses.go
```diff
@@ -47,7 +48,7 @@ func (ms msgServer) InsertBulkReputerPayload(
 		return nil, err
 	}
 	// Throw if already fulfilled -- can't return a response twice
-	if !nonceUnfulfilled {
+	if nonceUnfulfilled {
```
br4e, line 51: I still need to deal with this.

RedBird96 (approved):

PR #157, opened 2024-04-17
Diego/ora 1215 topichandler not updating state

xmariachi:
* Move change of state from topicsHandler to EndBlocker (otherwise it doesnt get persisted)
* e2e worker full cycle tested (incl unfulfilled & fulfilling nonces)


x/emissions/module/abci.go
```diff
@@ -90,5 +91,32 @@ func EndBlocker(ctx context.Context, am AppModule) error {
...
+						if err != nil {
+							fmt.Println("Error adding worker nonce: ", err)
+							return
+						}
+					}
+				}(topic)
+			}
+		}
+	}
+	wgAddingNonces.Wait()
```
guilherme-brandao, line 119: there's no need to query churnReadyTopics again + iterate through then. You already have everything available above in `topTopicsActiveWithDemand`. I guess you just need add the `AddWorkerNonce` right below `UpdateTopicEpochLastEnded`.

guilherme-brandao (changes requested):

x/emissions/module/abci.go
xmariachi: Good catch, changing it now.

integration/postinit.sh
```diff
@@ -23,5 +23,6 @@ mv $GENESIS.tmp $GENESIS
 
 
 echo "Starting allorad daemon and sleep for 3 seconds to let it start"
-$ALLORAD_BIN start & disown;
-sleep 3
+# $ALLORAD_BIN start & disown;
+# sleep 3
+$ALLORAD_BIN start
```
relyt29, line 28: why are you starting this without disowning, also the sleeping affects stability as well

xmariachi: sorry - that slipped in. I used that for my local testing without it. I'll change it. good catch.

xmariachi: done

kpeluso (approved):

PR #160, opened 2024-04-17
Delete rewards integration workflow

br4e:
forgot to delete the trigger


relyt29 (approved):
can delete inference request tbh 

just delete the whole integration task if you want

PR #162, opened 2024-04-18
Add pagination to rpc queries and some keeper methods

kpeluso:
### TRACK MY PROGRESS HERE

https://linear.app/upshot/issue/ORA-715/add-pagination-and-limits-for-emissions-module-queryserver-responses#comment-e116c215

### TICKET

https://linear.app/upshot/issue/ORA-715/add-pagination-and-limits-for-emissions-module-queryserver-responses


kpeluso:
Still todo:
1. Currently we value requests in a time order (first in, first active). This should be replaced with a "best price" order, though we must be sure that the `MaxRequestsPerTopic` param isn't too large so as to break things when a request is made.
  a. Perhaps we can have the consumer bring a request id if worse value => a new consumer can steal the spot of a previous one. (but then, really why would someone pay more if the inference is already covered... => maybe this doesn't matter and we could keep the current "time order")
3. Fix integration tests

x/emissions/keeper/keeper.go
```diff
@@ -1509,10 +1459,6 @@ func (k *Keeper) InsertReputer(ctx context.Context, topicId TopicId, reputer sdk
 	if err != nil {
 		return err
 	}
-	err = k.AddAddressTopic(ctx, reputer, topicId)
```
relyt29, line 1512: this line of code is why integration tests are breaking

PR #163, opened 2024-04-18
Fix network loss calc

guilherme-brandao:


kpeluso (approved):
just add exp10 to `CalcCombinedNetworkLoss` then gtg. preemptive approval

PR #164, opened 2024-04-18
Diego/ora 1222 top n workerreputer nonces retry to fulfil

xmariachi:
* Add Params `MaxRetriesToFulfilNoncesWorker` , `MaxRetriesToFulfilNoncesWorker`. Number of latest unfulfilled nonces that will be attempted to be resolved if not fulfilled in previous blocks. Can be seen as number of retries for an unfulfilled nonce to be atetmpted to be fulfilled.
   *  default to 10
* Use it in topicsHandler. 
* Adds test for params
* Tested in integration


app/topics_handler.go
```diff
@@ -41,12 +43,48 @@ func (th *TopicsHandler) calculatePreviousBlockApproxTime(ctx sdk.Context, block
...
+	var topN []*emissionstypes.ReputerRequestNonce
+	if len(reputerRequestNonces.Nonces) <= N {
+		topN = reputerRequestNonces.Nonces
+	} else {
+		topN = reputerRequestNonces.Nonces[:N]
+	}
+	return topN
+}
+
+func sortAndSelectTopNWorkerNonces(workerNonces emissionstypes.Nonces, N int) []*emissionstypes.Nonce {
```
kpeluso, line 65: maybe we don't need to sort? per our convo

kpeluso (approved):
just remove sorting on worker + reputer in topic_handler -- unneeded

preemptive approvval

app/topics_handler.go
xmariachi: changes done

PR #170, opened 2024-04-19
Diego/add inf functions to autocli

xmariachi:
Add get inference, forecast, losses at topic and block to autocli


relyt29 (approved):

relyt29:
just make sure make init doesn't break, otherwise good

PR #174, opened 2024-04-19
Guilherme/ora 1212 create integration test for emissions module emit rewards

guilherme-brandao:
This is adding an integration test in the emissions module and the goal is to simulate the happy path of our rewards cycle. It goes from topic creation until rewards distribution. 

To Consider:

- Rewards for workers are pending until the worker is fixed; right now, they are receiving 0 rewards since the tax is higher than the reward.
- Added some TODOs. The main ones are related to deciding which block we need to consider when querying for network losses; probably, it's the latest one for all cases.
- The test is just ensuring that things are not breaking and rewards are being generated, not checking values until we have the tax and the reputers' scores calculation update (it will come in subsequent PRs).


x/emissions/module/demand.go
```diff
@@ -158,7 +152,7 @@ func GetRequestsThatMaxFees(
 			fmt.Println("Error checking if request is valid at price: ", err)
 			return cosmosMath.Uint{}, cosmosMath.Uint{}, nil, err
 		}
-		//fmt.Println("Req id ", req.TopicId, " is valid at price ", req.MaxPricePerInference, " : ", isValidAtPrice)
+		fmt.Println("Req id ", req.TopicId, " is valid at price ", req.MaxPricePerInference, " : ", isValidAtPrice)
```
relyt29, line 155: do we want to leave this in before merge?

relyt29:
what's with all the blockHeight - 1 stuff

relyt29 (approved):
great work

kpeluso:
Per latest convo w Diederik, the tax needs to be removed now.

I also could use an explanation wrt the "blockHeight-1" stuff. Guess because you need the scores prior to that block, not as of that block? If so, could adopt the "GetXAtOrBeforeBlock" paradigm.

Since you started with your rewards work we have also developed a means of id'ing bundles from reputers and workers -- nonces -- so those could be used as well to get exactly the scores as of the time you want

I'd double check w Diederik and the litepaper to see if you could just use latest scores like we do with regrets though... That'd be simplest, which is probs why you added those todo comments -- just wanted to lay out all the options and conditions for which using the latest can be considered acceptable

x/emissions/module/rewards/reputer_rewards.go
```diff
@@ -13,22 +13,22 @@ func GetReputerTaskEntropy(
...
+	blockHeight int64,
 ) (
 	entropy alloraMath.Dec,
 	modifiedRewardFractions []alloraMath.Dec,
 	reputers []sdk.AccAddress,
 	err error,
 ) {
-	// TODO: Check if we need to get the block height - 1 (we might need to create GetLatestReputersScores)
-	scoresAtBlock, err := k.GetReputersScoresAtBlock(ctx, topicId, ctx.BlockHeight()-1)
+	latestScores, err := k.GetReputersScoresAtBlock(ctx, topicId, blockHeight)
```
kpeluso, line 23: minor: revert name

kpeluso (commented):
makes sense after a quick pass apart from the map `latestReputerScoresByReputer`. could use an explanation on that, esp per our discussion on indexing by block. will review deeper after merge confs

kpeluso (approved):

PR #175, opened 2024-04-19
simplify v2 stake removal

kpeluso:


x/emissions/keeper/msgserver/msg_server_stake.go
```diff
@@ -57,34 +57,29 @@ func (ms msgServer) AddStake(ctx context.Context, msg *types.MsgAddStake) (*type
...
-			return nil, types.ErrInsufficientStakeToRemove
-		}
 
-		// Push to the stake removal object
-		stakeRemoval.Placements = append(stakeRemoval.Placements, &types.StakePlacement{
-			TopicId: stakePlacement.TopicId,
-			Amount:  stakePlacement.Amount,
-		})
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	stakeToRemove := types.StakeRemoval{
```
br4e, line 72: wait you can only remove one stake at a time?

x/emissions/keeper/msgserver/msg_server_stake.go
```diff
@@ -118,22 +110,17 @@ func (ms msgServer) ConfirmRemoveStake(ctx context.Context, msg *types.MsgConfir
...
+	coins := sdk.NewCoins(sdk.NewCoin(params.DefaultBondDenom, amountInt))
+	ms.k.SendCoinsFromModuleToAccount(ctx, types.AlloraStakingAccountName, sender, coins)
 
-		// 7. update the stake data structures
-		err = ms.k.RemoveStake(ctx, stakePlacement.TopicId, senderAddr, stakePlacement.Amount)
-		if err != nil {
-			return nil, err
-		}
+	// Update the stake data structures
+	err = ms.k.RemoveStake(ctx, stakeRemoval.Placement.TopicId, sender, stakeRemoval.Placement.Amount)
```
br4e, line 121: Do we want to set `SetStakeRemoval` to empty after fully removing the stake?

x/emissions/keeper/msgserver/msg_server_stake.go
br4e: like what if someone wants to just unstake everywhere?

x/emissions/keeper/msgserver/msg_server_stake.go
kpeluso: think you mean "set the map `stakeRemoval`/`delegateStakeRemoval` to empty after fully removing the stake", to which I'd say "yes"

x/emissions/keeper/msgserver/msg_server_stake.go
kpeluso: gas is cheap here, should be fine. I don't imagine people to be participating in many topics as each requires specialized knowledge

PR #176, opened 2024-04-20
Automate run local network for integration

vladupshot:


relyt29:
```
+ docker build --pull -t allorad -f ./Dockerfile.development .
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/

unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /<REDACTED>/Dockerfile.development: no such file or directory
```

one strange thing is that the redacted path is one directory up (`../`) of the root of the entire workspace path

integration/local_testnet.sh
```diff
@@ -0,0 +1,210 @@
...
+VALIDATORS_IP_START=10
+HEADS_IP_START=20
+CHAIN_ID="testnet"
+LOCALNET_DATADIR="./$CHAIN_ID"
+# VALIDATORS_IP_START=10
+# PEERS=peers.txt
+# ALLORA_RPC="http://localhost:26657"
+ACCOUNTS_TOKENS=1000000
+
+ALLORA_RPC="http://172.20.0.10:26657"  # Take validator0
```
relyt29, line 17: doesn't necessarily require any additional work in this PR but something I want to make note of:

In the integration test (the actual golang code), I'd like to be able to hit all of the RPC addresses of all three nodes in (either) round-robin or random fashion to make sure that all three nodes are still up and running, consensus etc

We can hardcode values into the golang or have this bash script additionally write out the IP addresses of the validators on the local network to a (csv, json, etc) file so that the golang script can hit them

vladupshot: Just added https://github.com/allora-network/allora-chain/pull/176/files#diff-4cb83b8d20d1f4f6ba66763c0c11783c1312bc98ba4131946e2511830b4338d4R113

So you can get all validators IPs in `testnet/validators.json`

relyt29:
was able to get commit b2b1c087b1a02b8bdebb3443499275b6ce9c9839 to launch three validators

relyt29 (approved):
Launches three validators on localhost using docker

PR #177, opened 2024-04-21
Make reputer request be made if incomplete data

xmariachi:
* CalcNetworkInferences to still return incomplete set if any of the calculations for ValueBundle components have an error, providing empty arrays instead (affects OneIn/Out mostly)
* Chain to make a request to reputers with incomplete ValueBundles if that's the case 
* Ensure requests to workers and reputer are made correctly 
     * workers block height `t`, reputers blocks `(t - EpochLength, t - 2*EpochLength)`, checking against subzero heights.
* Added two CalcNetworkInferences tests.
     * Incomplete data to still return data. 
          * **Here's a catch**: currently if no regrets are set, the `CalcOneInForecasterValues` fails bc regrets are no higher 
 than epsilon, thus having to use empty array. Is that the correct behaviour, or do we want to change the behaviour inside `CalcOneInForecasterValues` ?
     * Same inferer and forecaster test (prev test was done with different addreses, but may be the same) 
* Added logging - some of it may be eventually removed.

### DIFF
--- a/app/topics_handler.go
+++ b/app/topics_handler.go
@@ -132,19 +132,15 @@ func (th *TopicsHandler) PrepareProposalHandler() sdk.PrepareProposalHandler {
 						reputerValueBundle, inferencesBlockHeight, err := synth.GetNetworkInferencesAtBlock(ctx, th.emissionsKeeper, topic.Id, nonceCopy.ReputerNonce.BlockHeight)
 						if err != nil {
 							fmt.Println("Error getting latest inferences at block: ", nonceCopy.ReputerNonce.BlockHeight, ", error: ", err)
-							continue
 						}
-						if reputerValueBundle == nil {
-							fmt.Println("Reputer value bundle is nil, skipping")
-							continue
-						}
+
 						blockDifference := currentBlockHeight - inferencesBlockHeight
 						previousBlockApproxTime, err := th.calculatePreviousBlockApproxTime(ctx, blockDifference)
 						if err != nil {
 							fmt.Println("Error calculating previous block approx time: ", err)
 							continue
 						}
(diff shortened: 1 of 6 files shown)


x/emissions/keeper/keeper.go
```diff
@@ -340,14 +342,20 @@ func (k *Keeper) IsReputerNonceUnfulfilled(ctx context.Context, topicId TopicId,
 	if err != nil {
 		return false, err
 	}
-
+	if nonce == nil {
+		return false, errors.New("nil reputer nonce provided")
+	}
 	// Check if the nonce is present in the unfulfilled nonces
 	for _, n := range unfulfilledNonces.Nonces {
+		if n == nil {
```
kpeluso, line 350: necessary?

app/topics_handler.go
```diff
@@ -132,19 +132,15 @@ func (th *TopicsHandler) PrepareProposalHandler() sdk.PrepareProposalHandler {
 						reputerValueBundle, inferencesBlockHeight, err := synth.GetNetworkInferencesAtBlock(ctx, th.emissionsKeeper, topic.Id, nonceCopy.ReputerNonce.BlockHeight)
 						if err != nil {
 							fmt.Println("Error getting latest inferences at block: ", nonceCopy.ReputerNonce.BlockHeight, ", error: ", err)
-							continue
-						}
-						if reputerValueBundle == nil {
```
kpeluso, line 137: you sure we should get rid of this?

kpeluso (approved):
Should we still keep all the print statements?

x/emissions/keeper/keeper.go
xmariachi: in theory not needed, but i wanted to make sure we see this if it happens

PR #179, opened 2024-04-22
Diego/fix cli breaking active topics

xmariachi:
Fix breaking cli when calling active-topics.

```
$ allorad query emissions active-topics '{"limit":"5"}'
Error: rpc error: code = Unknown desc = runtime error: invalid memory address or nil pointer dereference: panic
rpc error: code = Unknown desc = runtime error: invalid memory address or nil pointer dereference: panic

```

Creating the array with an estimated number based on topics to be inserted introduced an extra `nil` element which made the response break.


kpeluso (approved):
😆

PR #180, opened 2024-04-22
Test TestCalcNetworkLosses

br4e:
- Scale by 1e18
- Fix loss initialization logic
- Fix naive/combined variable switch
- test


br4e:
Should be good to go!

x/emissions/keeper/inference_synthesis/network_losses.go
```diff
@@ -112,6 +112,15 @@ func CalcNetworkLosses(
 			if err != nil {
 				return emissions.ValueBundle{}, err
 			}
+			oneE18, err := alloraMath.NewDecFromString("1000000000000000000")
```
kpeluso, line 115: should be a constant that gets imported from somewhere so you don't have to redefine each time

kpeluso (approved):
we may have to add more arithmetic with 1e18 other places e.g. in rewards

br4e:
yeah we can figure out a longer term solution

PR #183, opened 2024-04-23
remove activate-topic query

RedBird96:
- remove activate-topicId query
- activate topicId automatically on request-inference query


x/emissions/keeper/msgserver/msg_server_requests.go
```diff
@@ -123,5 +149,10 @@ func (ms msgServer) RequestInference(ctx context.Context, msg *types.MsgRequestI
 	if err != nil {
 		return nil, err
 	}
+	// 13. Activate topic
+	err = ms.ActivateTopic(ctx, request.TopicId)
+	if err != nil {
+		return nil, err
+	}
```
guilherme-brandao, line 156: If you want to activate the topic after the `RequestInference` is processed, you can do so by directly calling the keeper method. If `ActivateTopic` is not a method that is externally accessible, it doesn't make sense to keep it as a message in this specific case. Also, you will save an unnecessary query `GetTopicUnmetDemand` since you already have the BidAmount in the request struct.

guilherme-brandao (changes requested):

x/emissions/keeper/msgserver/msg_server_requests.go
RedBird96: The reason I used `GetParamsMinTopicUnmetDemand` is that in case of BidAmount can be not enough for meeting demand at first. So current request's BidAmount is less than `MinTopicUnmetDemand` but `topicUnmetDemand` meet for demand. In this case we can active topic?

x/emissions/keeper/msgserver/msg_server_requests.go
```diff
@@ -119,9 +119,22 @@ func (ms msgServer) RequestInference(ctx context.Context, msg *types.MsgRequestI
 	}
 	// 12. Write request state into the mempool state
 	request.BlockLastChecked = currentBlock
-	err = ms.k.AddToMempool(ctx, *request)
+	unmetDemand, err := ms.k.AddToMempool(ctx, *request)
```
kpeluso, line 122: I would've left the getter as a separate call but nbd

x/emissions/keeper/msgserver/msg_server_requests.go
```diff
@@ -119,9 +119,22 @@ func (ms msgServer) RequestInference(ctx context.Context, msg *types.MsgRequestI
...
+		return nil, err
+	}
+	minTopicUnmetDemandUint := cosmosMath.NewUintFromString(minTopicUnmentDemand.String())
+	isActivated, err := ms.k.IsTopicActive(ctx, request.TopicId)
+	if err != nil {
+		return nil, err
+	}
+	if unmetDemand.GTE(minTopicUnmetDemandUint) && !isActivated {
+		_ = ms.k.ActivateTopic(ctx, request.TopicId)
+	}
```
kpeluso, line 138: you should only bother to invoke all of this if topic is not active

kpeluso (changes requested):

kpeluso (approved):
looks good but idk why the commit is unverified...

x/emissions/keeper/msgserver/msg_server_requests.go
kpeluso: yeah a bid could have less than `MinTopicUnmetDemand`. That should not block the request from being submitted

x/emissions/keeper/keeper.go
```diff
@@ -1864,43 +1864,48 @@ func (k *Keeper) removeFromTopicMempool(ctx context.Context, topicId TopicId, re
...
+		return cosmosMath.Uint{}, err
+	}
+
+	err = k.AddUnmetDemand(ctx, request.TopicId, request.BidAmount)
+	if err != nil {
+		return cosmosMath.Uint{}, err
 	}
 
-	return k.AddUnmetDemand(ctx, request.TopicId, request.BidAmount)
+	return k.GetTopicUnmetDemand(ctx, request.TopicId)
```
guilherme-brandao, line 1908: `AddUnmetDemand` already get the topic unmet demand, you can just return it instead of querying again

x/emissions/keeper/msgserver/msg_server_requests.go
```diff
@@ -119,9 +119,25 @@ func (ms msgServer) RequestInference(ctx context.Context, msg *types.MsgRequestI
...
+	if !isActivated {
+		minTopicUnmentDemand, err := ms.k.GetParamsMinTopicUnmetDemand(ctx)
+		if err != nil {
+			return nil, err
+		}
+		minTopicUnmetDemandUint := cosmosMath.NewUintFromString(minTopicUnmentDemand.String())
+
+		if unmetDemand.GTE(minTopicUnmetDemandUint) {
+			_ = ms.k.ActivateTopic(ctx, request.TopicId)
```
guilherme-brandao, line 139: check error here

guilherme-brandao (changes requested):

x/emissions/keeper/keeper.go
RedBird96: Okay

x/emissions/keeper/msgserver/msg_server_requests.go
RedBird96: Do you think we should revert this request when activate topic failed?

guilherme-brandao: yeah, I mean, it should work properly. The user can be expecting that his request will activate the topic, it can be the main goal of the tx. The probability of getting an error at this point is near zero but it doesn't hurt to check.

guilherme-brandao: I mean If it generates an error at this point, it probably something wrong in protocol, so it's also good to have it for testing purposes.

guilherme-brandao (approved):

PR #189, opened 2024-04-24
fix issue on demand

RedBird96:
- fix error on get active topic (it returns wrong value if the number of topicId bigger than `maxTopicPages`)
- fix error on build best price (it was overwritten if the number of requests bigger than `maxRequestPages`)
- fix calculation logic in pagination cursor


kpeluso (approved):
awesome! thanks for the fixes to both pagination functions

PR #191, opened 2024-04-25
[ORA-1231] Test balance increase for rewards

br4e:
Sanity check that reputer and worker balances increase after inserting a worker and then a reputer payload and then crossing an epoch boundary


x/emissions/module/rewards/rewards_test.go
```diff
@@ -1,6 +1,7 @@
 package rewards_test
 
 import (
+	stdlog "log"
```
relyt29, line 4: can we use `s.T().Log()` to use the built in logging package in the testing package rather than importing something new?

relyt29 (approved):
approved pending one nitpick

kpeluso (approved):

PR #196, opened 2024-04-26
Prune old records after rewards distribution

guilherme-brandao:
Remove Network Losses, Reputer Losses, Inferences and Forecasts after they are being used for rewards. 

Introduces a new parameter called `MinEpochLengthRecordLimit` which is the minimum number of epoch lengths before the oldest unfulfilled reputation nonce that should be kept in storage. This parameter ensures that the network has the resources available to generate rewards for both old and current unfulfilled nonces.


x/emissions/keeper/keeper.go
```diff
@@ -2054,3 +2054,112 @@ func (k Keeper) CalcAppropriatePaginationForUint64Cursor(ctx context.Context, pa
...
+		return err
+	}
+	defer iter.Close()
+
+	for ; iter.Valid(); iter.Next() {
+		key, err := iter.KeyValue()
+		if err != nil {
+			return err
+		}
+		if err := k.networkLossBundles.Remove(ctx, key.Key); err != nil {
```
kpeluso, line 2159: im sure the tests pass but as far as I know, this pattern of deleting as you iterate is considered unsafe as it may interfere with the underlying logic of the iterator.

instead, I would add all keys to an array, then iterate over that array, removing each key from `networkLossBundles` as you go

this goes for this and the other few functions above

x/emissions/types/params.go
```diff
@@ -48,6 +48,7 @@ func DefaultParams() Params {
 		RegistrationFee:                 cosmosMath.NewInt(6),                      // how much workers and reputers must pay to register per topic
 		DefaultLimit:                    uint64(100),                               // default limit for pagination
 		MaxLimit:                        uint64(1000),                              // max limit for pagination
+		MinEpochLengthRecordLimit:       int64(3),                                  // minimum number of epochs to keep records for a topic
```
kpeluso, line 51: you need to add this new param to all the other places mentioned in [this document](https://www.notion.so/upshothq/Adding-New-Global-Parameters-Checklist-6216783e3b4f4280a2fda6f3487f0164)

x/emissions/module/rewards/rewards.go
```diff
@@ -259,7 +259,42 @@ func EmitRewards(ctx sdk.Context, k keeper.Keeper, block BlockHeight) error {
 		if err != nil {
 			return errors.Wrapf(err, "failed to delete topic reward nonce")
 		}
+
+		// Get oldest unfulfilled nonce - delete everything behind it
```
kpeluso, line 263: Consider abstracting much of this new code into a new function to make it easier to test.

kpeluso (changes requested):

kpeluso (approved):
Resolved all my own high-prio comments.

PR #194, opened 2024-04-26
add delegator reward logic

RedBird96:
add reward logic for delegator

PR #197, opened 2024-04-26
Increase stake for reputer rewards

br4e:
Delegator rewards should be handled separately by Michael's logic in https://github.com/allora-network/allora-chain/pull/194/ so we just need to handle reputer rewards correctly in this pr

### DIFF
--- a/x/emissions/module/rewards/rewards.go
+++ b/x/emissions/module/rewards/rewards.go
@@ -9,7 +9,6 @@ import (
 	chainParams "github.com/allora-network/allora-chain/app/params"
 	alloraMath "github.com/allora-network/allora-chain/math"
 	"github.com/allora-network/allora-chain/x/emissions/keeper"
-	"github.com/allora-network/allora-chain/x/emissions/keeper/inference_synthesis"
 	"github.com/allora-network/allora-chain/x/emissions/types"
 	mintTypes "github.com/allora-network/allora-chain/x/mint/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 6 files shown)


x/emissions/module/rewards/rewards.go
```diff
@@ -427,14 +428,32 @@ func payoutRewards(ctx sdk.Context, k keeper.Keeper, rewards []TaskRewards) erro
...
+		if reward.Reward.IsZero() {
+			continue
+		}
+
+		if reward.Type == ReputerRewardType {
+			oneE18, err := inference_synthesis.AlloraOneE18()
+			if err != nil {
+				return err
+			}
+			scaledReward, err := reward.Reward.Mul(oneE18)
```
kpeluso, line 440: Why do we multiply by `oneE18` here and not below?

Why don't we divide by `oneE18` at an earlier step, from which `reward` is derived?

kpeluso (changes requested):
The lack of parity in usage of `oneE18` with the other spots I highlighted is concerning.

Separately, we should explore doing this for delegators too, though could be a follow-up PR.

x/emissions/module/rewards/rewards.go
br4e: Because rewards are calculated as a decimal number in the rest of the system I believe. So if someones rewards are 1.2 allo, then reward.Reward is 1.2 allo, so we wouldn't need to divide by 1e18. However we do need to muliply by 1e18 (and the Math.floor the result) to get the number of tokens to send out or add to stake

br4e: and yes, should have multiplied in both places. I updated.

br4e:
looks like we might have to solve the 1e18 scaling question globally to solve failing test, since the module doesn't have enough funds to pay out that many tokens

x/emissions/module/rewards/rewards.go
relyt29: at no point in any of the reward code I wrote nor any of the code I think guilherme wrote do we ever operate with allo


all computations are done with the base unit of uAllo

I don't know why we need to do any multiplication or division by 1e18, ever. I don't think it's necessary in any point in the codebase

kpeluso: I think so too

kpeluso: removed

PR #198, opened 2024-04-26
[ORA-1244] All else equal. as total stake goes up, target emission per unit staked token should go down.

relyt29:
unit test to ensure relationships between various invariants are working as expected.


x/mint/module/abci.go
```diff
@@ -9,7 +9,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-func UpdateEmissionRate(
```
kpeluso, line 12: first time we've seen a misnomer like that!

kpeluso (approved):

PR #201, opened 2024-04-27
Ensure delegatee cant remove stake

br4e:
I checked that this will already throw on the line to fully remove the stake, but we should also have it throw when starting to remove stake. 

Add a unit test to be sure!


x/emissions/keeper/msgserver/msg_server_stake.go
```diff
@@ -80,6 +80,15 @@ func (ms msgServer) StartRemoveStake(ctx context.Context, msg *types.MsgStartRem
 		return nil, types.ErrInsufficientStakeToRemove
```
kpeluso, line 80: You should remove this conditional as it will always be false whenever there's any delegation (related to the purpose of this PR). But you should use the `ErrInsufficientStakeToRemove` error in the conditional you otherwise correctly added below.

kpeluso (changes requested):

x/emissions/keeper/msgserver/msg_server_stake.go
br4e: updated

kpeluso (approved):

PR #202, opened 2024-04-28
Test delegator reward

RedBird96:
Optimize delegator reward code and unit test reward query


x/emissions/module/rewards/reputer_rewards.go
```diff
@@ -193,14 +172,40 @@ func GetRewardForReputerFromTotalReward(
...
-			types.AlloraRewardsAccountName,
-			types.AlloraPendingRewardForDelegatorAccountName,
-			sdk.NewCoins(sdk.NewCoin(params.DefaultBondDenom, delegatorReward.SdkIntTrim())),
-		)
+		e18, err := alloraMath.NewDecFromSdkUint(keeper.GetAlloraExponent())
 		if err != nil {
-			return nil, errors.Wrapf(err, "failed to send coins to allora pend reward account")
+			return nil, err
+		}
+		delegatorReward, err = delegatorReward.Quo(e18)
```
kpeluso, line 179: you sure this needed?

x/emissions/keeper/keeper.go
```diff
@@ -652,6 +651,14 @@ func (k *Keeper) GetParamsMaxLimit(ctx context.Context) (uint64, error) {
 	return params.MaxLimit, nil
 }
 
+func (k *Keeper) GetAlloraExponent() Uint {
```
kpeluso, line 654: Should eventually move the contents of `x/emissions/keeper/inference_synthesis/common.go` here or replace their usage with this/similar functions.

kpeluso (approved):

x/emissions/keeper/keeper.go
RedBird96: I removed this function

PR #207, opened 2024-04-30
Adjust scripts to use seeds

vladupshot:


scripts/l1_node.sh
```diff
@@ -52,4 +52,5 @@ allorad \
     --moniker=${MONIKER} \
     --minimum-gas-prices=0${DENOM} \
     --rpc.laddr=tcp://0.0.0.0:26657 \
-    --p2p.persistent_peers=${PEERS}
+    --p2p.seeds=$SEEDS
```
kpeluso, line 55: should we not have both persistent peers and seeds?

kpeluso (approved):

kpeluso:
Should also merge this to `major-upgrade` branch

PR #208, opened 2024-04-30
Test Invariant: topics with met more demand get more reward

RedBird96:
topics with more met demand get more reward


relyt29 (approved):
this is a good test, I think we should keep it, but this test doesn't check that a topic with more demand gets more reward, it tests that a topic with more demand has a higher weight

yes perhaps that implies higher reward, but we should also test the reward directly, since that is ultimately what end-users care about rather than weight

this test is good though, and we should keep it in my opinion, or build on top of it to add checking the actual reward output

PR #210, opened 2024-04-30
[ORA1291] Fix Topic Registration Fee Bug, Test remaining lapses in msg server coverage

br4e:
Note that this fixes a bug


relyt29 (approved):
bug called topic creation fee rather than registration fees, they are potentially different fees.

lgtm

guilherme-brandao (approved):
lgtm!

PR #211, opened 2024-04-30
Fix single participant edge case

guilherme-brandao:
To resolve these bugs:
1. https://linear.app/upshot/issue/ORA-1307/chain-bug-div-by-zero-on-generatetasksrewards
2. https://linear.app/upshot/issue/ORA-1282/consensus-failure-failed-to-gen-task-rewards-when-inserting-losses


x/emissions/keeper/keeper.go
```diff
@@ -2,6 +2,7 @@ package keeper
 
 import (
 	"context"
+	errorsmod "cosmossdk.io/errors"
 	"encoding/binary"
 	"errors"
```
relyt29, line 7: ... Do we need to import this module if we're using cosmossdk.io/errors ?

kpeluso (approved):

PR #212, opened 2024-04-30
WIP Diego/ora 1298 chain no b7s requests are sent

xmariachi:
* New concept of Churn-ready vs Reward-Ready topics.
  * Changes flow of when/how the nonces, churnReady topics and rewards happen.
  * Pushing to churnReady happens for churnReady topics, removed from rewardReady topics (which made the chain unable to make requests).
* Untangle some other parts
  * A function `ApplyFuncOnAllChurnReadyTopics` done to untangle flow from reward-ready `SafeApplyFuncOnAllRewardReadyTopics`
  * GetWeights to be able to just get weights - not having to update state. Added flags to explicitly do so. 

* Lighten the recheck of epoch lengths on epochLength multiples only (avoiding too many operations in that case). 

* Added integration tests (from staking until workers and rewards incl nonces, etc)
* Upgrade wasm versions 

* **UPDATE** Contains stake, worker and reputer integration tests.


x/emissions/module/abci.go
```diff
@@ -33,10 +33,11 @@ func EndBlocker(ctx context.Context, am AppModule) error {
 		wg.Add(1)
 		go func(topic types.Topic) {
 			defer wg.Done()
-			// Check the cadence of inferences
+			// Check the cadence of inferences, and just in case also check multiples of epoch lengths
+			// to avoid potential situations where the block is missed
```
kpeluso, line 37: good call. I'm guessing this is the main reason why you moved `AddChurnReadyTopic()` here (out of `EmitRewards`, yeah?

x/emissions/module/rewards/topic_rewards.go
```diff
@@ -53,7 +53,7 @@ func SafeApplyFuncOnAllRewardReadyTopics(
 	fn func(ctx context.Context, topic *types.Topic) error,
 	topicPageLimit uint64,
 	maxTopicPages uint64,
-) error {
+	filterRewardNonces bool) error {
```
kpeluso, line 56: Instead of this and the other bool you added elsewhere probs have to "flip the relationship" between
`SafeApplyFuncOnAllRewardReadyTopics`
and
`ApplyFuncOnAllChurnReadyTopics`

the latter calls `GetTopicWeights` which calls the former. Meanwhile, reward ready topics are a subset of churnable topics

x/emissions/module/abci.go
xmariachi: AddChurnReady has been moved here, because EmitRewards should not be a dependency to making requests (which was causing the process to never start making requests).

Then, the use of multiple of EpochLength (instead of `> 2*EpochLength` ) is more efficient than the previous set up.

x/emissions/module/rewards/topic_rewards.go
xmariachi: However we first need to get all topic weights anyway, so we can get the topN (because that needs to be calculated on the fly).

The alternative would be to remove those filters, but because the filters are coupled with running a particular function on that data, so then to enable/disable filters.

Suggestion: We can decouple those `*ApplyFuncOnAll*ReadyTopics` functions into parts: return the data first, then do any particular filtering or action on each case (filtering: apply on rewardReady on nonces, no filter on churnReady; action: set previous weight on reWradReady and no action on churnReady).
This makes the separation explicit and easier to read.

x/emissions/module/abci.go
```diff
@@ -33,10 +33,11 @@ func EndBlocker(ctx context.Context, am AppModule) error {
...
 			if blockHeight == topic.EpochLastEnded+topic.EpochLength ||
-				blockHeight-topic.EpochLastEnded >= 2*topic.EpochLength {
-				fmt.Printf("Inference cadence met for topic: %v metadata: %s default arg: %s. \n",
+				(topic.EpochLastEnded-blockHeight)%topic.EpochLength == 0 {
```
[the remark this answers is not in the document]
xmariachi: good catch, it's should be `(blockHeight - topic.EpochLastEnded)%topic.EpochLength == 0`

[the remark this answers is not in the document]
xmariachi: I've also applied a simplification (since first condition is included in second condition now)

x/emissions/module/rewards/topic_rewards.go
```diff
@@ -104,13 +106,57 @@ func SafeApplyFuncOnAllRewardReadyTopics(
...
+	}
+	// Sort remaining active topics by weight desc and skim the top via SortTopicsByReturnDescWithRandomTiebreaker() and param MaxTopicsPerBlock
+	maxTopicsPerBlock, err := k.GetParamsMaxTopicsPerBlock(ctx)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get max topics per block")
+	}
+	weightsOfTopActiveTopics := SkimTopTopicsByWeightDesc(weights, maxTopicsPerBlock, block)
+
+	for topicId, weight := range weightsOfTopActiveTopics {
+		if weight.Equal(alloraMath.MustNewDecFromString("0")) {
```
guilherme-brandao, line 130: you can use `alloraMath.ZeroDec()`

x/emissions/module/rewards/topic_rewards.go
```diff
@@ -104,13 +106,57 @@ func SafeApplyFuncOnAllRewardReadyTopics(
...
+			continue
+		}
+		// Get the topic
+		topic, err := k.GetTopic(ctx, topicId)
+		if err != nil {
+			fmt.Println("Error getting topic: ", err)
+			continue
+		}
+		// Execute the function
+		err = fn(ctx, &topic)
```
guilherme-brandao, line 141: I'd name this function instead

x/emissions/module/rewards/topic_rewards.go
kpeluso: What you're suggesting ends up looking more complex than you think, because you need to paginate through all entities wherever you apply this logic now. Meanwhile, before, when it was all part of `*ApplyFuncOnAll*ReadyTopics`, you have all that pagination functionality abstracted away.

[Cosmos docs](https://docs.cosmos.network/v0.50/build/packages/collections#example-3) recommends it this way anyway, albeit with fewer safeguards than we have.

x/emissions/module/rewards/topic_rewards.go
xmariachi: Not sure what you mean... "fn" can be any function to execute, so it has to have a general name.

x/emissions/module/rewards/topic_rewards.go
xmariachi: ty! Changed it in all codebase.

x/emissions/keeper/inference_synthesis/network_inferences.go
```diff
@@ -619,3 +619,36 @@ func GetNetworkInferencesAtBlock(
...
+	if len(reputerRequestNonces.Nonces) <= N {
+		topN = reputerRequestNonces.Nonces
+	} else {
+		topN = reputerRequestNonces.Nonces[:N]
+	}
+	return topN
+}
+
+// Select the top N latest worker nonces
+func SelectTopNWorkerNonces(workerNonces emissions.Nonces, N int) []*emissions.Nonce {
```
kpeluso, line 646: if you're assuming any kind of order that's something that should be part of the "contract" you implicitly sign when you declare the function. For `SelectTopNReputerNonces()` obvi as well

x/emissions/module/rewards/rewards.go
```diff
@@ -184,12 +184,6 @@ func EmitRewards(ctx sdk.Context, k keeper.Keeper, blockHeight BlockHeight) erro
 	}
 	// for every topic
 	for topicId, topicReward := range topicRewards {
-		// To notify topic handler that the topic is ready for churn i.e. requests to be sent to workers and reputers
-		err = k.AddChurnReadyTopic(ctx, topicId)
```
kpeluso, line 188: def gonna conflict with @guilherme-brandao 's recent PR

x/emissions/module/rewards/topic_rewards.go
kpeluso: diego is right

kpeluso (commented):
it's probs good but i want you and me to confirm that's the case once more after resolving merge confs, as Guilherme's most recent PR is quite intimately related

kpeluso:
At risk of speaking in circles, I don't get why we don't move L30-70 of `abci.go` to near L102 of `rewards.go`. Line numbers given for new changes (not old version)

This way, we avoid an extra sort and pagination through all topics via `IdentifyChurnableAmongActiveTopicsAndApplyFn()`. Could possibly remove that function entirely thereafter.

xmariachi:
> At risk of speaking in circles, I don't get why we don't move L30-70 of `abci.go` to near L102 of `rewards.go`. Line numbers given for new changes (not old version)
> 
> This way, we avoid an extra sort and pagination through all topics via `IdentifyChurnableAmongActiveTopicsAndApplyFn()`. Could possibly remove that function entirely thereafter.

The set of weights is different when processing rewards (reward-ready) than when processing that nonces-update abci.go L30:70 code  (churn-ready). 
Currently weights are calculated twice. That is not necessary indeed. We could calculate weights only once, the churn-ready ones, to execute the churn-ready code; and then filter out those that are not reward-ready for the rewards code. 

This would be more efficient. However I may have to reorder the calls to make it cleaner. But this makes sense to me.

Also we need to be careful of adding nonces and updating EpochLastRun earlier (now we do it _after_ EmitRewards) doesn't have any undesired consequence.

x/emissions/module/rewards/rewards.go
xmariachi: will take a look, but yeah, this pr is conflicting every time I pull major-upgrade :P

xmariachi:
> > At risk of speaking in circles, I don't get why we don't move L30-70 of `abci.go` to near L102 of `rewards.go`. Line numbers given for new changes (not old version)
> > This way, we avoid an extra sort and pagination through all topics via `IdentifyChurnableAmongActiveTopicsAndApplyFn()`. Could possibly remove that function entirely thereafter.
> 
> The set of weights is different when processing rewards (reward-ready) than when processing that nonces-update abci.go L30:70 code (churn-ready). Currently weights are calculated twice. That is not necessary indeed. We could calculate weights only once, the churn-ready ones, to execute the churn-ready code; and then filter out those that are not reward-ready for the rewards code.
> 
> This would be more efficient. However I may have to reorder the calls to make it cleaner. But this makes sense to me.
> 
> Also we need to be careful of adding nonces and updating EpochLastRun earlier (now we do it _after_ EmitRewards) doesn't have any undesired consequence.

This is contained in last commit, please review that separately

kpeluso (approved):

PR #216, opened 2024-05-01
[ORA-1313] Fix Staking Supply Bonded Tokens Bug, Add tests for relationship between inflation rate and higher rate of staking

relyt29:
* Fixes bug where we were calling `StakingTokenSupply` instead of `TotalBondedTokens` thinking that the former was the supply of tokens staked when in reality it's just the total supply of uAllo and we need to get TotalBondedTokens to get the number of staked tokens
* Adds a test to show that the target emission per unit stake goes down when the number of tokens staked goes up, but that the total inflation rate continues to go up because the total amount of tokens staked went up.

### DIFF
--- a/x/mint/testutil/expected_keepers_mocks.go
+++ b/x/mint/testutil/expected_keepers_mocks.go
@@ -52,36 +52,20 @@ func (mr *MockStakingKeeperMockRecorder) BondedRatio(ctx interface{}) *gomock.Ca
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "BondedRatio", reflect.TypeOf((*MockStakingKeeper)(nil).BondedRatio), ctx)
 }
 
-// StakingTokenSupply mocks base method.
-func (m *MockStakingKeeper) StakingTokenSupply(ctx context.Context) (math.Int, error) {
-	m.ctrl.T.Helper()
-	ret := m.ctrl.Call(m, "StakingTokenSupply", ctx)
-	ret0, _ := ret[0].(math.Int)
-	ret1, _ := ret[1].(error)
-	return ret0, ret1
-}
-
-// StakingTokenSupply indicates an expected call of StakingTokenSupply.
-func (mr *MockStakingKeeperMockRecorder) StakingTokenSupply(ctx interface{}) *gomock.Call {
-	mr.mock.ctrl.T.Helper()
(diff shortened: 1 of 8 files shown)


x/mint/testutil/expected_keepers_mocks.go
```diff
@@ -67,6 +67,22 @@ func (mr *MockStakingKeeperMockRecorder) StakingTokenSupply(ctx interface{}) *go
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "StakingTokenSupply", reflect.TypeOf((*MockStakingKeeper)(nil).StakingTokenSupply), ctx)
 }
 
+// StakingTokenSupply mocks base method.
+func (m *MockStakingKeeper) TotalBondedTokens(ctx context.Context) (math.Int, error) {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "StakingTokenSupply", ctx)
```
RedBird96, line 73: I think we have to call "TotalBondedTokens" instead of "StakingTokenSupply"?

x/mint/testutil/expected_keepers_mocks.go
```diff
@@ -67,6 +67,22 @@ func (mr *MockStakingKeeperMockRecorder) StakingTokenSupply(ctx interface{}) *go
...
+	ret0, _ := ret[0].(math.Int)
+	ret1, _ := ret[1].(error)
+	return ret0, ret1
+}
+
+// StakingTokenSupply indicates an expected call of StakingTokenSupply.
+func (mr *MockStakingKeeperMockRecorder) TotalBondedTokens(ctx interface{}) *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "StakingTokenSupply", reflect.TypeOf((*MockStakingKeeper)(nil).StakingTokenSupply), ctx)
```
RedBird96, line 82: Same here about "StakingTokenSupply"?

x/mint/testutil/expected_keepers_mocks.go
relyt29: good catch ty

RedBird96 (approved):
lgtm

guilherme-brandao (approved):
lgtm!

PR #217, opened 2024-05-02
[ORA-1229] Added tests for reputer stake in rewards flow

guilherme-brandao:
Added tests to check if the reputer reward fraction increases when the stake increases and is zero when the stake is also zero.

Ticket: https://linear.app/upshot/issue/ORA-1229/test-invariant-all-else-the-same-reputers-with-higher-stake-should


relyt29 (approved):
sweet lgtm

PR #219, opened 2024-05-02
invariant test for only few top actors get reward

RedBird96:
invariant test that only few top actors get reward


x/emissions/module/rewards/rewards_test.go
```diff
@@ -270,6 +273,32 @@ func (s *RewardsTestSuite) TestStandardRewardEmission() {
...
+		s.ctx,
+		s.emissionsKeeper,
+		topicId,
+		block,
+		params.PRewardSpread,
+		networkLossBundles,
+	)
+	s.Require().Equal(len(inferers) <= int(params.GetMaxTopWorkersToReward()), true, "Only few Top workers can get reward")
+	s.Require().Equal(len(forecasts) <= int(params.GetMaxTopWorkersToReward()), true, "Only few Top workers can get reward")
+
```
guilherme-brandao, line 301: Not sure if it's a good idea to reuse this test for the following reasons:

1. It's not the main purpose of the test.
2. The default initial value for `MaxTopWorkers` and `MaxTopReputers` is 20. This test is only considering 5 actors for both, so it's not actually testing anything.

guilherme-brandao (changes requested):

x/emissions/module/rewards/rewards_test.go
RedBird96: Yes, `MaxTopWorkers` and `MaxTopReputers` is 20. So I did a test by changing the values as 5 and 4. And reverted it when commmit. Then I will make the script with more than 20 actors. Do you think it does make sense?

kpeluso (approved):
Note for later:
This PR also fixes a bug where one-out losses were being wrongly assigned as one-in losses.

PR #223, opened 2024-05-04
Errors in rewards calculations for topic A should not prevent rewards calculations for topic B

relyt29:
This PR is basically a mess of my work the last 3ish days. 

Contains the output of conversations between myself and kenny and diederik around standard devation being zero / more broadly error handling

also might have found a bug where reputers can cause denial of service on validators??? See `TestRewardsSkipsTopicsWithErrorsInCalculation`

**THIS CODE SHOULD NOT BE MERGED AS IS UNTIL THE DoS IS FIXED!!!**


kpeluso:
Converting to draft until DoS is fixed

kpeluso:
My PR https://github.com/allora-network/allora-chain/pull/224 that merges into this branch attempts to create a test to demonstrate if the DoS happens as a result of client or server logic. Currently unclear. Was having trouble on my hardware, so would be good if someone else could pickup for faster feedback loop.

kpeluso:
changes included here:
https://github.com/allora-network/allora-chain/pull/225

PR #226, opened 2024-05-07
Added stateless validations for main txs

guilherme-brandao:
- Refactor protos: create a file for each domain;
- Added stateless validation to the txs where it was necessary, the other txs already have sufficient validation => please let me know if you see another place that needs additional coverage.


RedBird96:
I think adding `Validate()` module is a good practices. Then I think we can remove validation checks in the main function. Because it's duplicated and I think we don't need to waste the time for checking again.

RedBird96 (approved):

PR #228, opened 2024-05-09
[ORA-1344] Add multi head node urls

RedBird96:
Added several BLOCKLESS_API_URLs to do retry when first head node is failed to receive

### DIFF
--- a/app/api.go
+++ b/app/api.go
@@ -144,22 +144,31 @@ func generateInferencesRequest(
 
 func makeApiCall(payload string) {
 	fmt.Println("Making Api Call, Payload: ", payload)
-	url := os.Getenv("BLOCKLESS_API_URL")
+	urls := readHeadUrls("BLOCKLESS_API_URL")
 	method := "POST"
 
-	client := &http.Client{}
-	req, err := http.NewRequest(method, url, strings.NewReader(payload))
-	if err != nil {
-		fmt.Println(err)
-		return
-	}
-	req.Header.Add("Accept", "application/json, text/plain, */*")
-	req.Header.Add("Content-Type", "application/json;charset=UTF-8")
+	for _, url := range urls {
+		client := &http.Client{}
+		req, err := http.NewRequest(method, url, strings.NewReader(payload))
+		if err != nil {
+			fmt.Println("Error sending request to head", err)
+			continue
+		}
+		req.Header.Add("Accept", "application/json, text/plain, */*")
+		req.Header.Add("Content-Type", "application/json;charset=UTF-8")
 
-	res, err := client.Do(req)
-	if err != nil {
-		fmt.Println(err)
+		res, err := client.Do(req)
+		if err != nil {
+			fmt.Println("Error sending request to head", err)
+			continue
+		}
+		defer res.Body.Close()
 		return
 	}
-	defer res.Body.Close()
+}
+
+func readHeadUrls(key string) []string {
+	urls := os.Getenv(key)
+	result := strings.Split(urls, ",")
+	return result
 }


kpeluso:
All retry and LB-related logic will be moved off of the chain to limit bloat, add retry and LB logic modularity, and ensure that there are no redundant requests (in this PR, each head making a request in succession)

app/api.go
```diff
@@ -144,22 +144,31 @@ func generateInferencesRequest(
...
+		}
+		req.Header.Add("Accept", "application/json, text/plain, */*")
+		req.Header.Add("Content-Type", "application/json;charset=UTF-8")
+
+		res, err := client.Do(req)
+		if err != nil {
+			fmt.Println("Error sending request to head", err)
+			continue
+		}
+		defer res.Body.Close()
```
srt0422, line 165: 🟡 **Medium — `defer` inside loop**

`defer res.Body.Close()` is inside a `for` loop. In Go, `defer` runs when the **function** returns, not when the loop iteration ends.

In this specific code, `return` immediately follows `defer`, so functionally it works — only one response body is ever opened. However, this is a well-known Go anti-pattern that will bite if the logic changes (e.g., if you want to collect responses from multiple URLs).

**Suggested fix:** Close the body explicitly:
```go
res, err := client.Do(req)
if err != nil {
    fmt.Println("Error sending request to head", err)
    continue
}
res.Body.Close() // not needed since we don't read it, but explicit close is clearer
return
```

app/api.go
```diff
@@ -144,22 +144,31 @@ func generateInferencesRequest(
...
+}
 
-	res, err := client.Do(req)
-	if err != nil {
-		fmt.Println(err)
-		return
-	}
-	defer res.Body.Close()
+func readHeadUrls(key string) []string {
+	urls := os.Getenv(key)
```
srt0422, line 171: 🟡 **Medium — Empty env var handling**

`strings.Split("", ",")` returns `[""]` (a slice with one empty string), not an empty slice. If `BLOCKLESS_API_URL` is unset or empty, this will attempt an HTTP request to an empty URL.

**Suggested fix:**
```go
func readHeadUrls(key string) []string {
    urls := os.Getenv(key)
    if urls == "" {
        return nil
    }
    result := strings.Split(urls, ",")
    // Optionally trim whitespace from each URL
    for i, u := range result {
        result[i] = strings.TrimSpace(u)
    }
    return result
}
```

app/api.go
```diff
@@ -144,22 +144,31 @@ func generateInferencesRequest(
...
+		req.Header.Add("Content-Type", "application/json;charset=UTF-8")
+
+		res, err := client.Do(req)
+		if err != nil {
+			fmt.Println("Error sending request to head", err)
+			continue
+		}
+		defer res.Body.Close()
 		return
 	}
```
srt0422, line 167: 🟢 **Low — Silent total failure**

If all URLs fail, the loop exits silently with no log. Consider adding a log after the loop:
```go
fmt.Println("Error: all head node URLs failed for request")
```

app/api.go
```diff
@@ -144,22 +144,31 @@ func generateInferencesRequest(
...
-	url := os.Getenv("BLOCKLESS_API_URL")
+	urls := readHeadUrls("BLOCKLESS_API_URL")
 	method := "POST"
 
-	client := &http.Client{}
-	req, err := http.NewRequest(method, url, strings.NewReader(payload))
-	if err != nil {
-		fmt.Println(err)
+	for _, url := range urls {
+		client := &http.Client{}
```
srt0422, line 151: 🟢 **Low — HTTP client created per iteration**

`&http.Client{}` is created inside the loop. Since the client is stateless and reusable, move it before the loop. Also consider adding a timeout:
```go
client := &http.Client{Timeout: 10 * time.Second}
for _, url := range urls {
    ...
}
```

srt0422 (commented):
## 🔍 Automated Code Review — PR #228

**PR:** [ORA-1344] Add multi head node urls
**Status:** CLOSED
**Scope:** 1 file, +22/-13 lines
**Triage Level:** Standard (small, isolated change)
**Reviewer:** OpenClaw Code Review Agent

---

### Summary

This PR adds retry/failover logic to `makeApiCall` by iterating over multiple head node URLs (comma-separated in `BLOCKLESS_API_URL`). On failure, it continues to the next URL. The approach is straightforward but has several issues worth noting.

### Findings

| # | Severity | Category | Description |
|---|----------|----------|-------------|
| 1 | 🟡 Medium | Correctness | `defer` inside loop — deferred `Body.Close()` only runs on function return, not loop iteration |
| 2 | 🟡 Medium | Correctness | Empty env var produces `[""]` → HTTP request to empty URL |
| 3 | 🟢 Low | Correctness | Silent total failure — no log when ALL URLs fail |
| 4 | 🟢 Low | Performance | `http.Client{}` created inside loop — should be reused |
| 5 | 🟢 Low | Security | No HTTP client timeout — requests can hang indefinitely |
| 6 | 🟢 Low | Style | `readHeadUrls` doesn't trim whitespace from URLs |

### Notes

Per existing review comment, retry/LB logic is being moved off-chain. These findings are relevant if the pattern is carried forward.

---
*Review by OpenClaw multi-agent code review system (models: claude-opus-4)*

PR #229, opened 2024-05-09
Prune worker and reputer nonces, do not request infinitely, tests

xmariachi:
* Add Pruning functions for nonces given a topic and blockheight (pruning needs to be done dependent on the epoch, and that is per-topic)
* Add deletion funtions for nonces (they clean the nonces array)
* tests for all new functions

* Call pruning of worker nonces before the current blockheight: if they are still unfulfilled, they're pruned - because we have reached a new epoch and so ground truth would be available to inferers.
* Leave several rounds (guided by param) of reputation allowance, to give some time for work done to be evaluated by reputers.


xmariachi:
Pending integ test fix here, will take a look.

kpeluso (approved):

PR #233, opened 2024-05-09
ORA1294 unit test misc other functions

br4e:
Update test TestGetAllReputersOutput to make sure that it checks that the values get closer to the correct value with more iterations of gradient descent. This way the test doesn't time out.


kpeluso (approved):

x/emissions/module/rewards/rewards_internal_test.go
```diff
@@ -671,28 +673,100 @@ func TestGetAllReputersOutput(t *testing.T) {
 	}
 	var numReputers int64 = 5
 	wantScores := []alloraMath.Dec{
-		alloraMath.MustNewDecFromString("17.536755245164326"),
-		alloraMath.MustNewDecFromString("20.302662649273707"),
-		alloraMath.MustNewDecFromString("24.278413872561256"),
-		alloraMath.MustNewDecFromString("11.365030585937692"),
-		alloraMath.MustNewDecFromString("15.211816727558011"),
+		alloraMath.MustNewDecFromString("17.46894"),
```
kpeluso, line 676: should probs comment here that these are the analytically-derived optimal scores that maximize the total consensus score, cuz unclear from variable name "wantScores" otherwise what this should be

PR #234, opened 2024-05-09
Guilherme/ora 1372 fix weight dec growing every block

guilherme-brandao:
- Fixing edge case when calculating the EMA 
- Apply new checks to avoid unnecessary ops

Reference: https://linear.app/upshot/issue/ORA-1372/fix-weight-dec-growing-every-block


kpeluso (approved):

kpeluso:
Bug where trailing 0 would grow after every Mul + extra instances of inefficient, extra comp

PR #238, opened 2024-05-13
Covering edge cases in Network Inference calc

guilherme-brandao:
- In network inference calc: 
If 0 inference: throw error
If 1 inference: network inference == single inference
if  >1 inference: calc network inference

- In network loss calc: 
If 0 forecasts: continue calc
If 1 inference: no need for to reputer to send one-out loss

- In scores calc :
If 1 forecast or inferer: score will be zero

In the case of scores, I needed to add code to handle when one-out losses are not empty (which will happen when we have single participant cases). Previously, we were generating random one-out losses for single participant cases in our tests and we were considering them; now we just do not consider one-out/one-in losses in these cases and use zero as score.

Reference: https://linear.app/upshot/issue/ORA-1374/consensus-failure-rewards-number-ratio-invalid-slice-length


relyt29:
Network inference calculation is called by the block proposer in the proposal preparation, so if there are no inferences by that time, it makes sense to skip the topic for rewards / not produce any inference

kpeluso (approved):

PR #240, opened 2024-05-13
[ORA-1406] Handle error cases for SendCoinsFromModuleToAccount

relyt29:
The various `SendCoins` functions throw errors deep within the [cosmos-sdk bank keeper](https://github.com/cosmos/cosmos-sdk/blob/eb7cc124a5035858e617e4d1d02c221a8d8a623b/x/bank/keeper/keeper.go#L273). We should always be catching those errors and doing something correct to handle them.

This code changes the behavior of the payment of rewards functions in the case of an error: before it would try to pay participants in a topic up until an error occured then every participant in that topic processed after the error would not get paid

now if there's an error trying to send someone funds, they get no rewards but the next participant in the chain gets their rewards.


relyt29:
should I write a unit test that tests that if there is an error the next participant still gets paid? probably

do I want to? no

kpeluso (approved):

PR #241, opened 2024-05-13
fix reputer + worker validation

kpeluso:
1. Remove duplicate and improper topicId check from worker bundle validation step
2. 1 greifing reputer or worker can’t fail the whole batch in the payload
3. Check and throw per reputer if any inputted inferences, forecasts, losses IsNaN
4. Actually call validation on worker bundles (was not being called)

Ticket: https://linear.app/upshot/issue/ORA-1410/fix-worker-and-reputer-payload-validations


x/emissions/types/reputer_value_bundle.go
```diff
@@ -0,0 +1,58 @@
+package types
+
+import (
+	"cosmossdk.io/errors"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+func (bundle *ReputerValueBundle) Validate() error {
```
xmariachi, line 9: This validation could also include the signature check validation that is now currently done when processing the bundle.

x/emissions/keeper/msgserver/msg_server_worker_payload.go
```diff
@@ -22,15 +22,18 @@ func (ms msgServer) VerifyAndInsertInferencesFromTopInferers(
...
 
-		// check signatures from the bundle throw if invalid!
+		if err := workerDataBundle.Validate(); err != nil {
+			return nil, err
+		}
+
+		// Check signatures from the bundle throw if invalid!
 		pk, err := hex.DecodeString(workerDataBundle.Pubkey)
 		if err != nil || len(pk) != secp256k1.PubKeySize {
 			return nil, types.ErrSignatureVerificationFailed
```
xmariachi, line 39: Here, if one bundle is invalid, it will throw the whole set as invalid (`return`,  not `continue` ). 
I think only that bundle should fail.
Other bundles should not have to suffer from one bundle being signed incorrectly.

Unless what we want here is to leave the verification also on the node leader who makes the insertion - in which case we'd need to make that explicit, and if it is not verified, punish the leader somehow for not verifying the data sent.

Applies to any `return`s in this function, there are other ones in the loop.

x/emissions/keeper/msgserver/msg_server_worker_payload.go
```diff
@@ -128,8 +131,7 @@ func (ms msgServer) VerifyAndInsertForecastsFromTopForecasters(
 		/// Do filters on the per payload first, then on each forecaster
 		/// All filters should be done in order of increasing computational complexity
 
-		// check signatures from the bundle throw if invalid!
-
+		// Check signatures from the bundle throw if invalid!
 		pk, err := hex.DecodeString(workerDataBundle.Pubkey)
 		if err != nil || len(pk) != secp256k1.PubKeySize {
 			return types.ErrSignatureVerificationFailed
```
xmariachi, line 137: Same comments here for in inferences above apply here as well

xmariachi (changes requested):
Some comments

x/emissions/types/reputer_value_bundle.go
guilherme-brandao: +1

guilherme-brandao: It can be added in the `WorkerDataBundle` validation as well.

guilherme-brandao: + I would also move the Msg Size validations into `Validate()` or `ValidateTopLevel()`, considering the goal of these validations is to perform stateless checks.

kpeluso: Think you're both right ab this and all other comments.

kpeluso: the size validations are dependent on state, namely a parameter in the keeper so have to keep that

xmariachi (approved):
lgtm

x/emissions/keeper/msgserver/msg_server_losses.go
```diff
@@ -77,9 +76,14 @@ func (ms msgServer) InsertBulkReputerPayload(
 	lossBundlesByReputer := make(map[string]*types.ReputerValueBundle)
 	latestReputerScores := make(map[string]types.Score)
 	for _, bundle := range msg.ReputerValueBundles {
+		if err := bundle.Validate(); err != nil {
+			fmt.Println("Error validating reputer value bundle: ", err)
```
relyt29, line 80: we should move away from using println we already print way too much 

should probably using `"cosmossdk.io/log"` instead

anyways the correct thing to do would be to flag some kind of error in the `MsgInsertBulkReputerPayload` and return an error from this function

relyt29: Note I am not saying to not continue here, I am saying to continue _and also return error_ this code is right now silently failing

relyt29: same with all the other continue cases below

x/emissions/keeper/msgserver/msg_server_worker_payload.go
```diff
@@ -22,25 +20,15 @@ func (ms msgServer) VerifyAndInsertInferencesFromTopInferers(
...
-			return nil, types.ErrSignatureVerificationFailed
-		}
-		pubkey := secp256k1.PubKey(pk)
-
-		src := make([]byte, 0)
-		src, _ = workerDataBundle.InferenceForecastsBundle.XXX_Marshal(src, true)
-		if !pubkey.VerifySignature(src, workerDataBundle.InferencesForecastsBundleSignature) {
-			return nil, types.ErrSignatureVerificationFailed
+		if err := workerDataBundle.Validate(); err != nil {
+			continue // Ignore only invalid worker data bundles
```
relyt29, line 31: but is it right to silently ignore, or should we be ignoring, continuing, and then raising error?

what do we do if we raise the error, what will the caller do with said error? log and move on?

relyt29:
other than my comments about not silently failing looks fine to me

x/emissions/keeper/msgserver/msg_server_losses.go
guilherme-brandao: +1  I think the best alternative is logging using the `ctx` 

e.g `ctx.Logger().Info(fmt.Sprintf("Error validating reputer value bundle: %s", err))`

x/emissions/keeper/msgserver/msg_server_worker_payload.go
kpeluso: I'm team "silent ignore" because the leader will be committing txs anyway, so already the logs are likely to be 1 anon away from the worker who produced the failing bundle. And don't think we have ability to emit events like in the EVM => relying on an explorer wouldn't help the worker gain insight either.

guilherme-brandao (approved):

integration/worker_inference_and_forecast_test.go
```diff
@@ -88,8 +89,19 @@ func InsertSingleWorkerBulk(m TestMetadata, topic *types.Topic, blockHeight int6
...
 	txResp, err := m.n.Client.BroadcastTx(m.ctx, m.n.BobAcc, workerMsg)
+	if err != nil && err.Error() == types.ErrNonceAlreadyFulfilled.Error() {
+		// May already be fulfilled due to other tests or multiple runs of these tests
+		log.Println("Nonce already fulfilled -- continuing")
+		return
+	}
 	require.NoError(m.t, err)
+
 	_, err = m.n.Client.WaitForTx(m.ctx, txResp.TxHash)
+	if err != nil && err.Error() == types.ErrNonceAlreadyFulfilled.Error() {
```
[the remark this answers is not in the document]
kpeluso: image

not so sure about that...

PR #242, opened 2024-05-14
Diego/ora 1398 use groundtruthlag

xmariachi:
Using GroundTruthLag :
* Wait for sending the request for reputation after the groundTruthLag has passed.
* Adapt pruning to take GroundTruthLag into account
* Tests


x/emissions/types/params.go
```diff
@@ -27,8 +27,8 @@ func DefaultParams() Params {
 		PInferenceSynthesis:             alloraMath.NewDecFromInt64(2),             // free parameter used in the gradient function phi' for inference synthesis
 		PRewardSpread:                   alloraMath.NewDecFromInt64(1),             // fiducial value = 1; Exponent for W_i total reward allocated to reputers per timestep
 		AlphaRegret:                     alloraMath.MustNewDecFromString("0.1"),    // how much to weight the most recent log-loss differences in regret EMA update
-		MaxUnfulfilledWorkerRequests:    uint64(5),                                 // maximum number of outstanding nonces for worker requests per topic from the chain
-		MaxUnfulfilledReputerRequests:   uint64(5),                                 // maximum number of outstanding nonces for reputer requests per topic from the chain
+		MaxUnfulfilledWorkerRequests:    uint64(100),                               // maximum number of outstanding nonces for worker requests per topic from the chain
+		MaxUnfulfilledReputerRequests:   uint64(100),                               // maximum number of outstanding nonces for reputer requests per topic from the chain
```
kpeluso, line 31: what's the rationale here? Would expect the old values...

x/emissions/module/abci.go
```diff
@@ -82,15 +82,19 @@ func EndBlocker(ctx context.Context, am AppModule) error {
 					MaxUnfulfilledReputerRequests = types.DefaultParams().MaxUnfulfilledReputerRequests
 					fmt.Println("Error getting max retries to fulfil nonces for worker requests (using default), err:", err)
 				}
-				reputerPruningBlock := blockHeight - int64(MaxUnfulfilledReputerRequests)*topic.EpochLength
-				fmt.Println("Pruning reputer nonces before block: ", reputerPruningBlock, " for topic: ", topic.Id, " on block: ", blockHeight)
-				am.keeper.PruneReputerNonces(sdkCtx, topic.Id, reputerPruningBlock)
+				reputerPruningBlock := blockHeight - (int64(MaxUnfulfilledReputerRequests)*topic.EpochLength + topic.GroundTruthLag)
```
kpeluso, line 85: related to this code but not exactly this code: I believe you also need to do a similar update to `pruneRecordsAfterRewards()` in `rewards.go` as well

app/topics_handler.go
```diff
@@ -31,14 +31,21 @@ func NewTopicsHandler(emissionsKeeper emissionskeeper.Keeper, mintKeeper mintkee
...
-func (th *TopicsHandler) calculatePreviousBlockApproxTime(ctx sdk.Context, blockDifference int64) (uint64, error) {
+func (th *TopicsHandler) calculatePreviousBlockApproxTime(ctx sdk.Context, inferenceBlockHeight int64, groundTruthLag int64) (uint64, error) {
 	mintParams, err := th.mintKeeper.GetParams(ctx)
 	if err != nil {
 		fmt.Println("Error getting mint params: ", err)
 		return 0, err
 	}
 	BlocksPerMonth := mintParams.GetBlocksPerMonth()
 	var approximateTimePerBlockSeconds float64 = float64(secondsInAMonth) / float64(BlocksPerMonth)
+	var blockDifference = ctx.BlockHeight() - inferenceBlockHeight
```
kpeluso, line 42: comment here would be helpful

kpeluso (changes requested):

x/emissions/module/abci.go
xmariachi: Checked the code, I think that can stay as is. 

The purpose of the pruneRecordsAfterRewards is to prune before any unfulfilledReputerNonce. 

However, the reputer nonces with gt lag remain unfulfilled for a longer time. So regardless that time, they will remain in there and oldest one will mark the limit in the same way = still works.

x/emissions/types/params.go
xmariachi: The rationale is that the unfulfilled worker nonces, with the advent of gt_lag, need to wait until they are requested so they can be fulfilled. 

So this number efffectively marks the max ratio that there could be `gt_lag / EpochLength`.
Bear in mind that this also means this needs to be set to keep all data from those blocks until they get fulfilled (which we need to).
So 100 is actually pretty conservative towards safety side imho, but the main point is that not only dev but also users  need to be aware as well.

Example: a 15 min freq for 24h prediction is already a ratio of 96.

app/topics_handler.go
xmariachi: added + renamed var name for clarity

x/emissions/types/params.go
kpeluso: ok think i get. 100% needs to go in the docs

kpeluso (approved):

PR #252, opened 2024-05-15
ORA1233 Varying alpha should change past/present performance

br4e:
Test Invariant: Varying Alpha (specifically the `TaskRewardAlpha` parameter) should change how performance important the past vs the present is for rewards


x/emissions/module/rewards/rewards_test.go
```diff
@@ -295,6 +296,429 @@ func (s *RewardsTestSuite) TestStandardRewardEmission() {
...
+
+	/// TEST 1 PART B
+
+	blockHeight3 := blockHeight2 + blockHeightDelta
+	s.ctx = s.ctx.WithBlockHeight(blockHeight3)
+
+	rewardsDistribution1_1 := s.getRewardsDistribution(topicId1, blockHeight3, workerValues1, reputerValues1)
+
+	require.True(areTaskRewardsEqualIgnoringTopicId(s, rewardsDistribution0_0, rewardsDistribution1_0))
+	require.True(areTaskRewardsEqualIgnoringTopicId(s, rewardsDistribution0_1, rewardsDistribution1_1))
```
kpeluso, line 597: this code is testing that the same code ran twice over, albeit at diff block times, yields the same output. did I understand that correctly? if so, this is a fine smoke test but tautological and disjoint from the intention of the ticket, which is to see if varying at least one of the alphas correctly affects the importance of current vs historical performance in determining rewards.

x/emissions/module/rewards/rewards_test.go
```diff
@@ -295,6 +296,429 @@ func (s *RewardsTestSuite) TestStandardRewardEmission() {
...
+
+	/// TEST 1 PART B
+
+	blockHeight3 := blockHeight2 + blockHeightDelta
+	s.ctx = s.ctx.WithBlockHeight(blockHeight3)
+
+	rewardsDistribution1_1 := s.getRewardsDistribution(topicId1, blockHeight3, workerValues1, reputerValues1)
+
+	require.True(areTaskRewardsEqualIgnoringTopicId(s, rewardsDistribution0_0, rewardsDistribution1_0))
+	require.False(areTaskRewardsEqualIgnoringTopicId(s, rewardsDistribution0_1, rewardsDistribution1_1))
```
kpeluso, line 719: the intention of this ticket is to see if varying at least one of the alphas _correctly_ affects the importance of current vs historical performance in determining rewards. If it's the case that increasing alpha has the opposite effect, then this test would still pass but it'd be misleading and false => we'd be lacking the coverage we intend to gain by this ticket.

Therefore, a proper test would correctly check if the direction of reward change implied by the changed alpha matches expectations. In this case, a higher alpha should imply that worse current performance in a worker impacts their rewards (they earn less) more severely than a lower alpha selection (assuming they were performing perceptibly better before => the impact of their comparatively better historical performance should keep their reward higher)

kpeluso (changes requested):

x/emissions/module/rewards/rewards_test.go
br4e: You understood that correctly. I was establishing that that was true before testing whether or not changing `TaskRewardAlpha` affects the resulting values to isolate the effect of changing `TaskRewardAlpha`

x/emissions/module/rewards/rewards_test.go
br4e: Sounds good! Ill update the task description and the pr

br4e: Updated

kpeluso (approved):
like i [said here](https://github.com/allora-network/allora-chain/pull/253#discussion_r1605920362), all of these new tests could benefit from a tin description describing the setup and what's actually being tested because the setup is so (necessarily) intricate. so plz add those in for all new tests here + in other PRs. but otherwise lgtm

PR #253, opened 2024-05-15
ORA1240 Higher AlphaRegret means higher impact of current worker performance on regret

br4e:


x/emissions/module/rewards/rewards_test.go
```diff
@@ -1061,6 +1071,165 @@ func (s *RewardsTestSuite) TestIncreasingTaskRewardAlphaIncreasesImportanceOfPre
...
+	require.NoError(err)
+	require.False(notFound)
+
+	worker2_1, notFound, err := k.GetInfererNetworkRegret(s.ctx, topicId1, sdk.AccAddress(workerAddrs[2].String()))
+	require.NoError(err)
+	require.False(notFound)
+
+	require.True(worker0_0.Value.Gt(worker0_1.Value))
+	require.True(alloraMath.InDelta(worker1_0.Value, worker1_1.Value, alloraMath.MustNewDecFromString("0.00001")))
+	require.True(alloraMath.InDelta(worker2_0.Value, worker2_1.Value, alloraMath.MustNewDecFromString("0.00001")))
```
kpeluso, line 1230: > We have 2 trials with 2 epochs each, and the first worker does worse in 2nd epoch in both trials, enacted by their increasing loss between epochs.
> We increase alpha between the trials to prove that their worsening performance exaggerates their regret.

that's the intention, right?

^that as a comment would improve this test's clarity tremendously.
your tests in general could benefit from quick snippets of clarity like that.

but also, is this test showing the opposite is true: cuz `worker0_0 > worker0_1`
=> in the second trial, we ultimately regret the worker less, though the weight of their worsening performance was increased
How do we explain that?

kpeluso: nvm I'm wrong. regret is more like “how grateful we are for the worker”/“how much we’d regret not having them”, less “how much we regret to include them”

PR #260, opened 2024-05-16
rm range over map keys

kpeluso:
Accompanies all maps that were formerly `range`d over with a sorted list of the map's keys.

This changes 33/35 hits for `:= range` that were over maps.
Remaining 2 I found to be non-harmful => left them as-is.

TODO
* Fix tests


x/emissions/module/rewards/rewards.go
```diff
@@ -35,7 +37,8 @@ func EmitRewards(ctx sdk.Context, k keeper.Keeper, blockHeight BlockHeight, weig
 
 	totalRewardToStakedReputers := alloraMath.ZeroDec()
 	// for every topic
-	for topicId, topicReward := range topicRewards {
+	for _, topicId := range sortedTopics {
```
xmariachi, line 40: shouldnt this iterate through topicRewards? if not, why not? 
(ie does `GenerateRewards...` guarantee same keyset?)

relyt29: `GenerateRewards` should only ever return a strict subset of the values in `sortedTopics` but I also agree, there's no reason to write it this way when we can simply sort topicRewards

relyt29: ah I see using sortedTopics reduces the need to sort twice

but it also looks like someone's merged PR overwrote my PR to remove the non-generic sort functions

xmariachi: a strict subset is not the same as the same subset, is that situation covered?

PR #261, opened 2024-05-17
ORA1426 Fix test after removing range over map keys

br4e:


kpeluso:
continued in https://github.com/allora-network/allora-chain/pull/265

PR #255, opened 2024-05-16
[ORA-1375] Implement Updated Tokenomics

relyt29:
The white paper was updated to implement an additional cap on the total rate of emission per unit stake (this then factors in to the total inflation rate of the token on the network). This PR is meant to implement those changes in the codebase.

PR #263, opened 2024-05-17
[ORA-1375] Update f_stakers value on same monthly cadence as ^e_i target emissions

relyt29:
This PR is optional in the sense that we can live without it, minor detail I forgot from #255 - lets not pointlessly write into storage every block


x/emissions/module/rewards/rewards.go
```diff
@@ -90,7 +90,11 @@ func EmitRewards(ctx sdk.Context, k keeper.Keeper, blockHeight BlockHeight, weig
 			continue
 		}
 	}
-	if !totalReward.IsZero() {
+	blocksPerMonth, err := k.GetParamsBlocksPerMonth(ctx)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get blocks per month")
+	}
+	if !totalReward.IsZero() && uint64(blockHeight)%blocksPerMonth == 0 {
```
kpeluso, line 97: Rewards should be continuously doled out though, right? so we do need to write this data somewhere each block, no?

relyt29: This if statement singularly controls the writing of (reputer component of) the f_stakers value in the mint module. The if statement has no effect on the rewards being doled out or not

f_stakers is used monthly so there's no need to update it every block when we're only using it once per month. We could do some complicated sum or average calculation over the course of the month that seems like too complicated for what is an indirect value

kpeluso (commented):
Just waiting for you to amend mint module BeginBlock to do `uint64(blockHeight)%blocksPerMonth == 1` in its analogous control flow.

relyt29:
> Just waiting for you to amend mint module BeginBlock to do `uint64(blockHeight)%blocksPerMonth == 1` in its analogous control flow.

apparently it already was ==1 

https://github.com/allora-network/allora-chain/blob/0cfc6a0050aa27c4f02206cb55483b0715ca0353/x/mint/module/abci.go#L130-L133

kpeluso (approved):

PR #268, opened 2024-05-18
rm needless AccAddress conversions

kpeluso:


x/emissions/keeper/msgserver/msg_server_params.go
```diff
@@ -4,15 +4,10 @@ import (
...
-	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
 func (ms msgServer) UpdateParams(ctx context.Context, msg *types.MsgUpdateParams) (*types.MsgUpdateParamsResponse, error) {
-	sender, err := sdk.AccAddressFromBech32(msg.Sender)
-	if err != nil {
-		return nil, err
-	}
-	isAdmin, err := ms.k.IsWhitelistAdmin(ctx, sender)
+	isAdmin, err := ms.k.IsWhitelistAdmin(ctx, msg.Sender)
```
guilherme-brandao, line 10: there's no address validation whatsoever now. We need to be careful to have at least one validation in the flow before considering strings for other operations.

guilherme-brandao (changes requested):

x/emissions/keeper/msgserver/msg_server_params.go
kpeluso: that's a great point.

kpeluso: what kinds of validation you think we need to be worried about if we remove the bech32 check…?
https://github.com/allora-network/allora-chain/pull/268#discussion_r1605906112

it’s not like they could e.g. sql injection…

I also don’t see any validation similar to what was suggested on osmosis e.g.
https://github.com/osmosis-labs/osmosis/blob/003dceb23c1773454f142e44acc7a911993e2a32/x/superfluid/keeper/msg_server.go#L47C23-L47C41

PR #269, opened 2024-05-18
ORA-1436 Disallow self delegation

br4e:
Halborn found an attack where someone could steal stake by reputers delegating to themselves. Prevent this attack by disallowing self delegation.


kpeluso:
Just waiting to hear back from Halborn their thoughts about this solution before merge. lgtm though

PR #272, opened 2024-05-19
[ORA-1445] Only reset topic fee revenue counter if topic is "top" topic being paid out rewards to

relyt29:
I checked the ordering of the flows, `GetAndOptionallyUpdateActiveTopicWeights` is called before `EmitRewards`, which is what does the resets (and is changed in this diff)

### DIFF
--- a/x/emissions/module/rewards/rewards.go
+++ b/x/emissions/module/rewards/rewards.go
@@ -169,8 +169,14 @@ func GenerateRewardsDistributionByTopic(
 					return nil, errors.Wrapf(err, "failed to add weight to sum")
 				}
 			} else {
-				// This way we won't double count from this earlier epoch revenue the next epoch
-				// This must come after GetTopicFeeRevenue() is last called per topic because otherwise the returned revenue will be zero
+				// For topics that are active and top, they will get their rewards paid out this block.
+				// Everybody else doesn't. Therefore, for topics that have topicFeeRevenue but havent received rewards,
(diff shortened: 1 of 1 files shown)


x/emissions/module/rewards/rewards.go
```diff
@@ -168,13 +168,13 @@ func GenerateRewardsDistributionByTopic(
...
 				}
-			}
-
-			// This way we won't double count from this earlier epoch revenue the next epoch
-			// This must come after GetTopicFeeRevenue() is last called per topic because otherwise the returned revenue will be zero
-			err = k.ResetTopicFeeRevenue(ctx, topicId, blockHeight)
-			if err != nil {
-				return nil, errors.Wrapf(err, "failed to reset topic fee revenue")
+			} else {
+				// This way we won't double count from this earlier epoch revenue the next epoch
```
kpeluso, line 172: well, we're clearly doing the opposite of these comments now

relyt29: right but the "double counting" will be for topics that haven't been paid out rewards yet!

relyt29: do you want me to just change the comment or do you want to do something else / handle this somehow else

kpeluso: just change the comment cuz we are "double counting" in the sense of the original comment. like, just explain why we're doing this control flow

kpeluso (approved):

PR #275, opened 2024-05-20
rm running avg calculation w cheaper simpler alt

kpeluso:
This way of computing averages was innappropriate for this use case leading to excess compute. Now, simplified.

### DIFF
--- a/x/emissions/keeper/inference_synthesis/network_losses.go
+++ b/x/emissions/keeper/inference_synthesis/network_losses.go
@@ -8,56 +8,40 @@ import (
 	emissions "github.com/allora-network/allora-chain/x/emissions/types"
 )
 
-type WorkerRunningWeightedLoss struct {
-	SumWeight Weight
-	Loss      Loss
+type RunningWeightedLoss struct {
+	UnnormalizedWeightedLoss Loss
+	SumWeight                Weight
 }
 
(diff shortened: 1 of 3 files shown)


x/emissions/keeper/inference_synthesis/network_losses.go
```diff
@@ -8,66 +8,59 @@ import (
...
+type RunningWeightedLoss struct {
+	UnnormalizedWeightedLoss Loss
+	SumWeight                Weight
 }
 
-// Update the running weighted loss for the worker
-// Source: "Weighted mean" section of: https://fanf2.user.srcf.net/hermes/doc/antiforgery/stats.pdf
-// nextValue format - raw value
-// weight format - logged value
+// Update the running information needed to calculate weighted loss per worker
```
relyt29, line 16: I think it would be helpful to have a comment description of what this function does mathematically (is this going to be defined in the white paper?)

kpeluso: it's just taking numerator and denominator needed to calculate a simple weighted average. I'll make the comment clearer.

relyt29:
[ORA-1456]

relyt29:
I'm good on this PR, but I think kenny should be the one to approve it

PR #276, opened 2024-05-20
rm allorarequestaccountname

kpeluso:
`allorarequestaccountname` is now excessive and we can just use ecosystem bucket as the only escrow for circulating funds


guilherme-brandao (approved):
Good idea! I'm in line with the change, but Tyler seems to be the best person to give the final approval.

relyt29:
Big supporter of the idea, but this PR seems to have compile errors, I don't think the job was finished right

relyt29:
I am now working on this PR

PR #273, opened 2024-05-20
ORA1395 Create flexible stress test for testing any number of topics and reputers

br4e:

PR #278, opened 2024-05-20
Diego/ora 1452 stress tests fix nonces error

xmariachi:
* Realignment with topic EpochLastRan in case of nonces failure. This makes the test to eventually get the right nonces again and this continue inserting without failure.
* Get approx block time from chain params, adding params utility fn.
* Add funding continuously so the topic doesn't run out of funds => no weights => no nonces
* Cleaned code


br4e:
This generally looks good to me. I'm a little bit wary of adding so much logging to the core protocol, but if others don't think that's an issue I'm cool with it. 

In terms of how this interacts with my pr https://github.com/allora-network/allora-chain/pull/273, I should be able to rebase on your changes and make it work, although it might be a little bit dicey to integrate. Let's discuss in our standup

br4e (approved):
Yup! Good to go.

PR #280, opened 2024-05-20
ORA1395 Create flexible stress test for any number of reputers, workers, or topics

br4e:
Allow any number of reputers, workers, and topics to be specified. The target for this pr is Diego's stress testing pr. Additional topics are run in a subroutine, and this pr attempts to handle all potential concurrency issues. 

Example command: 

`bash local_testnet_l1.sh ;  STRESS_TEST=true REPUTERS_PER_EPOCH=1 WORKERS_PER_EPOCH=1 TOPICS_PER_EPOCH=1 go test -v -timeout 0   -test.run TestStressTestSuite .`


stress/n_workers_reputers__stress_test.go
```diff
@@ -294,86 +414,114 @@ func WorkerReputerLoop(m TestMetadata, topicId uint64) {
...
 		}
-		err = StakeReputer(m, topicId, reputerAddress, reputerAccount, stakeToAdd)
-		if err != nil {
-			fmt.Println("Error staking reputer address: ", reputerAddress, " - ", err)
-			continue
+
+		if i == 0 {
+			createNewWorkerAccount()
+			createNewReputerAccount()
+		} else {
```
kpeluso, line 505: is this needed? don't we want to add `workersPerEpoch`-many workers, etc. in each epoch including the first one?

stress/n_workers_reputers__stress_test.go
```diff
@@ -272,16 +273,135 @@ func InsertReputerBulk(m TestMetadata,
...
+			return topicFunderAddress, topicFunderAccount
+		}
+		return topicFunderAddress, topicFunderAccount
+	}
+
+	if topicsPerEpoch == 0 {
+		topicFunderAddress, topicFunderAccount := getTopicFunder()
+		WorkerReputerLoop(m, topicFunderAddress, topicFunderAccount)
+		topicCount++
+	} else {
```
kpeluso, line 369: same thing w the control flow here

kpeluso (approved):
looks clean to me!

stress/n_workers_reputers__stress_test.go
br4e: This is because we want to start with one worker and one reputer even if `workersPerEpoch` and `reputersPerEpoch` are set to 0

stress/n_workers_reputers__stress_test.go
br4e: If we don't want to add any topics per epoch, then we just create a single topic by calling the `WorkerReputerLoop` 
 once. If we are adding topics every epoch, then we want to enter the loop to call `WorkerReputerLoop` repeatedly

stress/n_workers_reputers__stress_test.go
```diff
@@ -272,16 +273,183 @@ func InsertReputerBulk(m TestMetadata,
...
+	initialWorkerCount int,
+	initialReputerCount int,
+) {
+	topicId := SetupTopic(m, topicFunderAddress, topicFunderAccount)
+
+	report := func(a ...any) {
+		fmt.Println("[ TOPIC", topicId, "] ", a)
+	}
+
+	reputersPerEpoch := lookupEnvInt(m, "REPUTERS_PER_EPOCH", 0)
```
xmariachi, line 443: Env vars are read in two places. 
I think they should be read outside (in the stress_test.go for example) and passed in as function parameters. This helps declaring all function dependencies.

stress/n_workers_reputers__stress_test.go
```diff
@@ -272,16 +273,183 @@ func InsertReputerBulk(m TestMetadata,
...
+		topicFunderAddress, topicFunderAccount := getTopicFunder()
+		WorkerReputerLoop(m, topicFunderAddress, topicFunderAccount, workerCount, reputerCount)
+		topicCount++
+	} else {
+		for {
+			startIteration := time.Now()
+
+			for j := 0; j < topicsPerEpoch && topicCount < topicsMax; j++ {
+				topicFunderAddress, topicFunderAccount := getTopicFunder()
+				go WorkerReputerLoop(m, topicFunderAddress, topicFunderAccount, workerCount, reputerCount)
```
xmariachi, line 399: Not clear to me why initialWorkerCount, initialReputerCount (used here as workerCount, reputerCount) increase in every outside iteration (4 lines above)
Is it so the new topics "catch up" with the number of workers and reputers that the previous topics have?

stress/n_workers_reputers__stress_test.go
```diff
@@ -272,16 +273,183 @@ func InsertReputerBulk(m TestMetadata,
...
+				topicCount++
+			}
+
+			workerCount += workersPerEpoch
+			reputerCount += reputersPerEpoch
+
+			elapsedIteration := time.Since(startIteration)
+			sleepingTimeSeconds := iterationTimeSeconds - elapsedIteration
+			for sleepingTimeSeconds < 0 {
```
xmariachi, line 408: I think that this is not needed: if there is no need to sleep (ie neg number), just do not sleep.

stress/n_workers_reputers__stress_test.go
```diff
@@ -272,16 +273,183 @@ func InsertReputerBulk(m TestMetadata,
...
+				sleepingTimeSeconds += iterationTimeSeconds
+			}
+			fmt.Println(time.Now(), "Main loop sleeping", sleepingTimeSeconds)
+			time.Sleep(sleepingTimeSeconds)
```
xmariachi, line 412: We'd need a way to control the sync of the go routines. 
Use a :
`var wg sync.WaitGroup` 
similarly as it's done in other places.

xmariachi (changes requested):
Good work. 
I've left some comments, will be addressing some myself now. 

I've run it and the chain stands with 100 topics, which seems to suggest that the problem we saw in the depth-first tests (chain losing consensus) doesn't show up in this breadth-first one.

FYI It shows around 20 topics active per block, max 30, as a normal distribution would suggest (cadences hit at different blocks). Cadences met are consistent on same blocks - that's another test pass.

stress/n_workers_reputers__stress_test.go
xmariachi: added in merged pr

stress/n_workers_reputers__stress_test.go
xmariachi: removed in merged pr

stress/n_workers_reputers__stress_test.go
xmariachi: read separately in merged pr

stress/n_workers_reputers__stress_test.go
```diff
@@ -272,16 +273,183 @@ func InsertReputerBulk(m TestMetadata,
...
+	if err != nil {
+		m.t.Fatal(err)
+	}
+
+	err = RegisterWorkerForTopic(m, topicFunderAddress, topicFunderAccount, topicId)
+	if err != nil {
+		m.t.Fatal(err)
+	}
+
+	err = RegisterReputerForTopic(m, topicFunderAddress, topicFunderAccount, topicId)
```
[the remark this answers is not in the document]
br4e: It doesn't seem like there are any direct consequences of this, I see this just as a way to get the topic fully set up so we can perform all of the tests with additional workers, reputers, etc. 

If you want I can generate a set of `topicFunderReputerAddress` so we can have a separate reputer, but that might be overkill. Let me know if you feel strongly about this

stress/n_workers_reputers__stress_test.go
br4e: Yes, it's so the new topics catch up.

stress/create_topic_test.go
```diff
@@ -3,11 +3,12 @@ package stress_test
 import (
 	alloraMath "github.com/allora-network/allora-chain/math"
 	emissionstypes "github.com/allora-network/allora-chain/x/emissions/types"
+	"github.com/ignite/cli/v28/ignite/pkg/cosmosaccount"
 	"github.com/stretchr/testify/require"
 )
 
 // test that we can create topics and that the resultant topics are what we asked for
-func CreateTopic(m TestMetadata) (topicId uint64) {
+func CreateTopicTest(m TestMetadata, epochLength int64) (topicId uint64) {
```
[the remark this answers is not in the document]
br4e: done

xmariachi (approved):
lgtm

PR #281, opened 2024-05-21
Add WaitGroup, env vars and others

xmariachi:
* Use WaitGroup for the subroutines and give an end to the process
* Change waiting time (no need to wait if elapsed time is bigger than an epoch)
* env vars defined as function parameters


br4e (approved):
Awesome, Lgtm!

PR #282, opened 2024-05-21
Covering edge cases for zero values in all reported losses

guilherme-brandao:
-> Covering edge cases where:
1) All reported inferer values are 0
2) All reported forecaster values are 0
...
Considering all properties of the Reputer Value Bundle. 

-> Critical fix where we start using the fTolerance from the topic instead of a mocked one. 

Ticket: https://linear.app/upshot/issue/ORA-1453/cover-edge-case-zero-values-in-all-reported-losses-values

### DIFF
--- a/x/emissions/module/rewards/worker_rewards.go
+++ b/x/emissions/module/rewards/worker_rewards.go
@@ -99,7 +99,11 @@ func GetWorkersRewardFractions(
 		scores = append(scores, workerLastScoresDec)
 	}
 
-	rewardFractions, err := GetScoreFractions(latestWorkerScores, flatten(scores), pRewardSpread)
+	epsilon, err := k.GetParamsEpsilon(ctx)
+	if err != nil {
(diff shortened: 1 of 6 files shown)


relyt29 (approved):
approved pending tests pass (github actions are being wonky today)

x/emissions/module/rewards/worker_rewards.go
```diff
@@ -99,7 +99,11 @@ func GetWorkersRewardFractions(
 		scores = append(scores, workerLastScoresDec)
 	}
 
-	rewardFractions, err := GetScoreFractions(latestWorkerScores, flatten(scores), pRewardSpread)
+	epsilon, err := k.GetParamsEpsilon(ctx)
```
kpeluso, line 102: you could probs pass this into the function scope instead of this getter

kpeluso (approved):

PR #285, opened 2024-05-22
[ORA-1407] Turn back on Checking that Validators and Reputers get paid inflationary tokens

relyt29:
Someone had just casually turned off the tests, no big deal

This PR relies on https://github.com/allora-network/networks/pull/3 - the initial token distribution at genesis MUST follow the distribution laid out in the whitepaper, otherwise N_Circ in the e_i calculation becomes infinitesimally small relative to our 1e27 token supply and so the network just refuses to make any inflation

### DIFF
--- a/integration/local_testnet_l1.sh
+++ b/integration/local_testnet_l1.sh
@@ -43,8 +43,7 @@ popd
 
 echo "Download generate_genesis.sh from testnet"
 mkdir -p ${LOCALNET_DATADIR}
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.stale
kpeluso, line 1: so delete the file...?

integration/postinit.sh
kpeluso, line 1: @xmariachi you're still not using this, right?

kpeluso (approved):

.github/workflows/integration.stale
relyt29: It's in the git history if we ever need it, no reason to keep it

PR #287, opened 2024-05-22
Rm worker payload entropy

kpeluso:


kpeluso:
handled by recent PR from @xmariachi

PR #289, opened 2024-05-22
Diego/ora-1452 ORA-1455 stress tests fix nonces error

xmariachi:
Sorting of maps is nondeterministic. Hence it was placing the sequence of pseudorandom numbers in wrong order. 
Sorting the keys in advance ensures a deterministic map iteration.


kpeluso (approved):
Great sleuthing!

You sure we wanna keep all these print statements?

relyt29 (approved):
so much work for so little lines of code changed :sob:

x/emissions/module/rewards/topic_skimming.go
```diff
@@ -48,7 +48,10 @@ func SkimTopTopicsByWeightDesc(weights map[TopicId]*alloraMath.Dec, N uint64, bl
 	for topicId := range weights {
 		topicIds = append(topicIds, topicId)
 	}
-
```
kpeluso, line 51: we might not actually need this because the operations of `SortTopicsByWeightDescWithRandomTiebreaker()` don't involve a map-like data structure => could save compute to remove

PR #290, opened 2024-05-22
Make CI run faster and cheaper by merging github action workflows into one job

relyt29:
This PR changes the CI workflow so that instead of having 4 separate VMs that all need to install the same golang tools and compile the same allora-chain codebase, we install all the tools once and compile once, and then run the various format and test actions one after another.

This makes things faster because we don't have to wait for github to allocate us 4 vms instead of 1 and we don't have to build the same code 4 times, and the various tasks can share the same build cache.

it makes things cheaper because github charges us for runtime minutes used, and this uses less minutes used by reusing the build process per thing we have to do


okedeji (approved):
This looks good to me, maybe we can also make them separate by jobs so they can run in parralel?  say Job for setup which they all depend on but separate job for fmt, unit test and integration text.

that will make the actions faster

vladupshot (approved):
lgtm

relyt29:
> This looks good to me, maybe we can also make them separate by jobs so they can run in parralel? say Job for setup which they all depend on but separate job for fmt, unit test and integration text.
> 
> that will make the actions faster

so my understanding is that separate jobs run on separate virtual machines. If you find evidence that's not true I'm happy to separate the tasks out by job, but if they can't share the build cache then there's no point

[](https://docs.github.com/assets/cb-25535/mw-1440/images/help/actions/overview-actions-simple.webp)

PR #291, opened 2024-05-22
ORA1471 Check that worker balance and reputer stake increases during stress testing

br4e:
This is not a strict test because some of the newly created accounts haven't had a chance to acquire rewards in some cases, either due to no contributing enough or not having a chance to be present during a round where there was not a nonce error

Screenshot 2024-05-22 at 2 59 00 PM


relyt29 (approved):
i like this, it would be good if we could do some test like

"if by block N, x% of workers haven't had a balance increase, throw an error" - what do you think

br4e:
Good idea, I added a simple version of this.

kpeluso (approved):

PR #296, opened 2024-05-23
Print Final Summary Report of Percent of Failures

relyt29:
Hacked this together a bit quickly so it's not perfect but thought it would be good to at least be able to see how many errors we got after each stress test run, even to just give a vague sense of success

@xmariachi could use your eyes to make sure i'm capturing all the errors we care about in the right places


test/stress/concurrency.go
```diff
@@ -0,0 +1,116 @@
+package stress_test
+
+import (
+	"fmt"
+	"sync"
+)
+
+type TOPIC_ID = uint64
+type IS_REPUTER = bool // false = WORKER
```
xmariachi, line 9: doesnt seem applicable here, if it does please explain usage

xmariachi (approved):
LGTM, just I think the IS_REPUTER could be removed

PR #298, opened 2024-05-24
[ORA-1413] Remove fmt.Print statements

guilherme-brandao:
• Stop using `fmt.Print` for general logs and use `ctx.Logger` with `Debug` or `Info` depending on the purpose.
• Remove redundant logs and some logs that apparently is not useful anymore.

Adapt the error log pattern:
• For msg-related errors: use `Wrapf ` to return a meaningful message to the user.
• For validator processing errors: use `ctx.Logger` - use `Warn` for non-throwing errors, otherwise use `Error`.

Ticket: https://linear.app/upshot/issue/ORA-1413/println-log

### DIFF
--- a/app/api.go
+++ b/app/api.go
@@ -2,14 +2,13 @@ package app
 
 import (
 	"encoding/json"
-	"fmt"
 	"net/http"
 	"strconv"
 
+	emissionstypes "github.com/allora-network/allora-chain/x/emissions/types"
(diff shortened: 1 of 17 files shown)


relyt29 (approved):
lgtm

br4e (approved):
Awesome work!

kpeluso (approved):

PR #299, opened 2024-05-24
[ORA-1468] implement round robin, random rpc node target of queries for stress testing

relyt29:
Right now the stress test only ever submits transactions via validator0s RPC. This code changes it so that you can choose to submit queries round robin between the validator RPC URLs, or randomly choose the next RPC target based on a random seed in the bash environment

PR #300, opened 2024-05-25
[ORA-1468] Refactor Stress-Test

relyt29:
Refactors the stress-test code.

While debugging #299 I found that the `initializeNewWorkerAccount` code was written in such a way that the outer for loop iterates through all workers that need initialization and the for loop returns when getting the account from the registry fails. That would be fine, except the stress testing code that prints errors can't distinguish between a failure case that is genuine and a failure case that is being used to end the loop, which causes the report at the end to report false positives. This led me to try to refactor that code, which snowballed into a PR that refactors the entire stress-test.

Things this code does:

1. Puts line breaks in on long lines, especially function declarations
2. Moves functions into their own files so not all of the code is in one humongous n_stress_blah.go file
3. Attempts to add meaningful comments to every function declaration
4. For the two important functions, `workerReputerLoop` and `workerReputerCoordinationLoop`, go through the function line by line and add comments at every concrete step taken with what that step is.
5. For the two important functions, take large swaths of related code and factor them into their own functions
6. Remove all anonymous function usage. The use of anonymous functions that were mutating the state of the surrounding code made understanding where variable values change VERY difficult. 
7. Related to anonymous functions, refactor mappings of the workers / reputers inside a topic so that it is not necessary to re-get/query for `cosmosaccount`s more than once - just store them in a map at the beginning of the workerReputerLoop and use it later
8. Renames environment variables that were describing untrue things such as "perEpoch" when in fact "iterations" and "epochs" are separate concepts

Future work for the stress test could include getting rid of all the timing and duration code and simply using `Client.WaitForNBlocks` though if we want to accurately simulate erratic network behavior perhaps using duration code is better.


test/stress/n_workers_reputers_stress_test.go
```diff
@@ -0,0 +1,314 @@
...
+		m,
+		funder,
+		reputers,
+		initialWorkerReputerFundAmount,
+	)
+	if err != nil {
+		m.T.Log(topicLog(topicId, "Error funding reputer accounts: ", err))
+	} else {
+		m.T.Log(topicLog(topicId, "Funded", len(reputers), "reputer accounts."))
+	}
```
br4e, line 185: Just combine these into a single call with both lists, each call to the chain takes a long time espeically when compounded across an entire topic stress test

test/stress/n_workers_reputers_stress_test.go
```diff
@@ -0,0 +1,314 @@
...
+		makeReport,
+	)
+	if err != nil {
+		m.T.Log(topicLog(topicId, "Error checking worker rewards: ", err))
+		if makeReport {
+			saveTopicError(topicId, err)
+		}
+	}
+
+	// 11. Check that the reputer have been paid rewards via a stake greater than the initial amount
```
br4e, line 294: you have two 11's. Maybe just drop the numbers so we don't need to worry about changing the numbers every time we update the logic

br4e (approved):
I didn't read every line, but this generally looks good with a couple of comments. 

I'm assuming that you tested this thoroughly, and if so, go ahead and address the comments and then merge if you are ready

relyt29:
added a commit that renames the names of the accounts so accounts aren't reused across stress test runs, you can now run the stress test multiple times by changing the `SEED` environment variable or try to reuse the same accounts and see if it breaks by leaving SEED alone. SEED is initialized to 0

relyt29:
> Ill run another round of the topics test

@br4e I'd like to wait to get another set of eyes on this since it's such a big PR, @xmariachi or @guilherme-brandao would be helpful, but you should run another round of tests with this branch in the meantime, just not to lose time

relyt29:
I'm concerned that I'm seeing `no valid bundles found` on this branch - is that normal to see, or did I mess up the bundle generation code somehow? @xmariachi

br4e:
Cool yeah sounds good, I was doing a faster approve because I assumed that this was blocking me being able to test the https://github.com/allora-network/allora-chain/pull/299 on major-upgrade, but since that isn't the case I'm all for deeper consideration!

br4e:
Ah I didn't realize that https://github.com/allora-network/allora-chain/pull/299 is now a part of this pr. Yeah there is a lot going on here and we should take time to review. I can do a thorough high topic test run later.

kpeluso:
Yeah @relyt29 next time you do this please make a cascading PR so others can more easily ascertain and critique the unique changes your proposing.

relyt29:
I did make it a cascading PR, #299 is the cascade. I could have merged 299 in, as i had two approvals but I was hoping for more eyes before merging both 299 and this

xmariachi:
> I'm concerned that I'm seeing `no valid bundles found` on this branch - is that normal to see, or did I mess up the bundle generation code somehow? @xmariachi
> 
> I think this code likely has a bug but my brain is too fried right now to figure it out

You should not see that... so maybe something has been messed up there. Do you see it on every iteration?

relyt29:
> You should not see that... so maybe something has been messed up there. Do you see it on every iteration? 

I think yes, I will have to dig into it

guilherme-brandao (approved):
looks good! nice work

PR #301, opened 2024-05-25
Kenny/ora 1390 base define validation policy on reputernonces

kpeluso:
Add multi-reputer lookup. Useful for forthcoming release of `allora-inference-base` that will look this up to do a stake-weighted vote with maximal efficiency (1 request) (reasoning for this [given here](https://linear.app/upshot/issue/ORA-1390/base-define-validation-policy-on-reputernonces#comment-bca021fc))

To probably/possible do:
* Should I use a diff global var? Seems to carry same risk in both places it's used.
* Is there a better way to define these constants such that they're more easily tunable? Does their "tunability" matter?
   * DEFAULT_MAX_REPUTERS_FOR_STAKE_QUERY
   * MAX_NUMBER_STAKE_QUERIES_PER_REQUEST


x/emissions/keeper/queryserver/query_server_stake.go
```diff
@@ -34,6 +35,34 @@ func (qs queryServer) GetReputerStakeInTopic(ctx context.Context, req *types.Que
...
+// including their stake in themselves and stake delegated to them.
+// Also includes stake that is queued for removal.
+func (qs queryServer) GetMultiReputerStakeInTopic(ctx context.Context, req *types.QueryMultiReputerStakeInTopicRequest) (*types.QueryMultiReputerStakeInTopicResponse, error) {
+	maxLimit, err := qs.k.GetParamsMaxLimit(ctx)
+	if err != nil {
+		maxLimit = types.DefaultParamsMaxLimit()
+	}
+
+	if uint64(len(req.Addresses)) > maxLimit {
+		return nil, status.Error(codes.InvalidArgument, "cannot query more than 100 addresses at once")
```
relyt29, line 48: maxLimit is dynamic in the sense that it's set by params yeah? then this error should be 
```golang
status.Error(codes.InvalidArgument , fmt.sprintf("cannot query more than %d addresses at once", maxLimit))
```

kpeluso: yessir

PR #302, opened 2024-05-27
ORA1289 Test delegators cant easily game the system

br4e:
Test some of the ways that delegators might be able to game the system, and ensure that they are not possible.


x/emissions/keeper/msgserver/msg_server_stake_test.go
```diff
@@ -817,3 +817,223 @@ func (s *KeeperTestSuite) TestRewardDelegateStake() {
...
+	score := alloraMath.MustNewDecFromString("17.53436")
+
+	delegatorAddr := sdk.AccAddress(PKS[0].Address())
+	reputerAddr := sdk.AccAddress(PKS[1].Address())
+	workerAddr := sdk.AccAddress(PKS[2].Address()) // target
+
+	registrationInitialBalance := cosmosMath.NewInt(10000)
+	reputerStakeAmount := cosmosMath.NewInt(1e2)
+	delegatorRatio := cosmosMath.NewInt(1e3)
+	delegatorStakeAmount := reputerStakeAmount.Mul(delegatorRatio)
```
kpeluso, line 947: I think this test is fine but I'd add in another that does something more like the following:
1. Have a reputer and delegator
2. Have them earn some rewards for 1-2 rounds.
   a. Record what each is entitled to receive as of that moment.
3. Have a 1000x delegator enter
4. Do another reward round
5. See if that 1000x delegator can withdraw more than they should, "stealing rewards" because they rebalance the current proportion of delegator stake to be in their favor
   a. So, the 1000x delegator should receive just rewards from the round they were in, NOT that PLUS any rewards that were otherwise payable to existing reputer or delegator

I'd add in 1 additional delegator to this check that delegates less / comparable to the reputer, as there is some accounting that is unique to how their rewards are counted that may admit this attack.

ik just a draft, just letting you know my thoughts in case helpful

br4e: definitely helpful, thanks! Im at the stage of figuring out which more complex tests to run, Ill add this one.

x/emissions/keeper/msgserver/msg_server_test.go
```diff
@@ -257,3 +257,44 @@ func (s *KeeperTestSuite) TestCreateSeveralTopics() {
 	s.Require().NotNil(result)
 	s.Require().Equal(initialTopicId+2, result)
 }
+
+func (s *KeeperTestSuite) inEpsilon(value alloraMath.Dec, target string, epsilon string) {
```
relyt29, line 261: can we make a test util folder and move these inEpsilon functions into a separate package if we need them in multiple places in the tests, seems like bad practice to have duplicate code

br4e: updated! Check it out. There is a little weirdness around not being able to call s.Require() when initializing the alloratestutil package INSIDE initializing an outer package, so I switched to this function contructor pattern

test/util/inEpsilon.go
```diff
@@ -0,0 +1,58 @@
+package alloratestutil
+
+import (
+	alloraMath "github.com/allora-network/allora-chain/math"
+	"github.com/stretchr/testify/require"
+)
+
+func (s *AlloraTestUtilSuite) inEpsilon(value alloraMath.Dec, target string, epsilon string, require *require.Assertions) {
```
relyt29, line 8: you could also pass `s.T` as a `t *testing.Testing` parameter then import the require package and use it as require.NoError(t, err)

I'm ok with something going in the setup that "sets up" the epsilon function, which I think is what this is

relyt29: frankly think the inEpsilon stuff can be its own pr, easier to parse that way,

let me see if I can push that out quickly

x/emissions/keeper/msgserver/msg_server_test.go
relyt29: https://github.com/allora-network/allora-chain/pull/324

br4e: merge

br4e:
Thanks @relyt29 for taking a look. I'll wait for @kpeluso to sign off also because he indicated he wanted to review in the task

kpeluso (approved):

PR #304, opened 2024-05-27
Fix TestTotalInferersRewardFractionGrowsWithMoreInferers

kpeluso:
`TestTotalInferersRewardFractionGrowsWithMoreInferers` was not testing to see if the correct ratios (inferer and forecaster reward fractions) were increasing with increasing inferer and forecasters, resp. It was testing a completely different ratio not of concern.

Additionally, this test was flaky in that it would usually pass but only sometimes fail. This was due to a randomness seeded with a sensitive value (nanosecond time).

That volatility meant that extra inferers could have high losses => they actually may not earn a greater fraction, because the protocol will pay classes of actors out more based on their numbers but also their performance. so the test failed because ~20% of the times their losses were so high that despite their greater number => they earned less, which is actually expected. I solved this in an expedient way, the better way would be for me to care about how high the losses were before i decide what to test, the best way would be for me to systematically adjust that loss, but not of immediate concern


x/emissions/module/rewards/scores_test.go
```diff
@@ -320,7 +319,7 @@ func PrepareMockLosses(reputersCount int, workersCount int) (
 	reputersOneInNaiveLosses [][]alloraMath.Dec,
 ) {
 	for i := 0; i < reputersCount; i++ {
-		rand.Seed(time.Now().UnixNano())
```
relyt29, line 323: oh lord almighty

relyt29: why is this inside the for loop anyways, can't it just be seeded once

relyt29:
Title of this PR could be better, doesn't really fully explain the problem or the solution

kpeluso:
Updated title + description

relyt29 (approved):
thanks for the explanation

x/emissions/module/rewards/scores_test.go
kpeluso: yeah true

PR #305, opened 2024-05-27
[ORA-1397] Add more debug logging

relyt29:
I realize @xmariachi has some ideas for alternative loggers, and I support that

this pr adds additional logging (also fixes some random go-staticcheck warnings I saw while poking through the codebase)

Does not change functionality at all

I am not going to die on this hill if people don't like where I chose to put logging or if you think I didn't log enough feel free to suggest additional things to log. We can also close this and not log anything. This PR just represents what I thought might be nice to log by default (at level debug)


kpeluso (approved):
Keep an eye out for if the passing of the larger "sdkcontext" and the added calls to "unwrap context" noticeably impact performance. This would be quickly revealed by noting if unit tests take perceptibly longer to run. I could see that being one of those things that sneaks up on us.

Otherwise, lgtm

relyt29:
> Keep an eye out for if the passing of the larger "sdkcontext" and the added calls to "unwrap context" noticeably impact performance. This would be quickly revealed by noting if unit tests take perceptibly longer to run. I could see that being one of those things that sneaks up on us.
> 
> Otherwise, lgtm

that's a fair point I wonder if we can get around it by passing pointers to context structs if we're worried about it

kpeluso:
> that's a fair point I wonder if we can get around it by passing pointers to context structs if we're worried about it

I don't see that being done elsewhere, but don't see why we can't do that. You just have to be cautious about constantly dereferencing in that case, so it'd work best if we're simply relaying the context, which might not actually be that much.

PR #306, opened 2024-05-28
[ORA-1487] Make sending only valid inferer values

RedBird96:


app/topics_handler.go
```diff
@@ -107,6 +107,10 @@ func (th *TopicsHandler) requestTopicReputers(ctx sdk.Context, topic emissionsty
 			ctx.Logger().Error("Error getting latest inferences at block: ", nonceCopy.ReputerNonce.BlockHeight, ", error: ", err)
 			continue
 		}
+		if len(reputerValueBundle.InfererValues) == 0 {
```
xmariachi, line 110: if reputerValueBundle is actually nil, this will panic :P 
that needs to be checked too

if reputerValueBundle == nil || len(reputerValueBundle.InfererValues) == 0 {

xmariachi (changes requested):
Add a panic coverage. 

And, I'd also add that coverage inside the `api.go:generateLossesRequest` function.
The function should be made sure to not fire the request, independently from where it's called, if stdin is nil.
For example, this could also happen if the marshalling fails for some reason.

app/topics_handler.go
RedBird96: okay, I didn't add it because it doesn't return nil in any case. Anyway, it's good to add this one.

PR #307, opened 2024-05-28
Fix logging

xmariachi:
Fix logging - previous statements were not printing the values


relyt29 (approved):
more logs more better!!!!

PR #309, opened 2024-05-28
ORA1232 Test invariant: Reputer closer to consensus should get more rewards

br4e:


br4e:
https://github.com/allora-network/allora-chain/pull/312 instead

PR #311, opened 2024-05-29
ORA1311: Reward per participant goes up when participant drops out

br4e:


relyt29 (approved):
looks good to my limited eyesight

PR #312, opened 2024-05-29
ORA1232 Test invariant: Reputer closer to consensus should get more rewards

br4e:
I think basically we lost the concept of absolute value from EQ32 in the way we are calculating log values without actually taking logs

`abs(log(1/2)) == log(2/1)`

so I think that we need to make this change to push the difference in the right direction


x/emissions/module/rewards/rewards_internal.go
```diff
@@ -333,9 +333,17 @@ func GetConsensusScore(
...
-		rLossOverConsensusLoss, err := rLoss.Quo(consensusLosses[i])
-		if err != nil {
-			return alloraMath.ZeroDec(), err
+		var rLossOverConsensusLoss alloraMath.Dec
+		if rLoss.Gt(consensusLosses[i]) {
+			rLossOverConsensusLoss, err = rLoss.Quo(consensusLosses[i])
+			if err != nil {
+				return alloraMath.ZeroDec(), err
+			}
+		} else {
```
relyt29, line 342: abs(log(1/2)) == log(2/1) is the principle here

br4e: Like tyler said, this change fits with EQ32

kpeluso: long rabbit hole later... plz see my latest changes

PR #315, opened 2024-05-29
Diego/ora 1507 reputer calls depend on being churn ready renamed

xmariachi:
* Add param TopicFeeRevenueDecayRate , set to 0.025
* Apply to ResetTopicRevenue 
* The decay will be applied now to all topics. Because the feeRevenue of the non-topN is added to the total revenue (via  `sumRevenueOfBottomTopics` ) , this can be considered added to the mix and so it's ok to apply decay. This reduces the variability of topics going in/out of the topN, leading to more stable topics.

WIP passes unit tests, but untested on integration/stress.


math/utils.go
```diff
@@ -41,6 +41,21 @@ func CalcEma(
 	return ret, nil
 }
 
+func CalcExpDecay(
```
relyt29, line 44: lacking comment function header describing the function

PR #319, opened 2024-05-30
Use clear to remove data

xmariachi:
Clear instead of iterate the collection to remove, use a ranger where applicable, nil ranger where all data is removed.


x/emissions/keeper/keeper.go
```diff
@@ -1904,111 +1888,19 @@ func (k *Keeper) PruneRecordsAfterRewards(ctx context.Context, topicId TopicId,
...
-
-	// Remove data at all keys
-	for _, key := range keysToDelete {
-		if err := k.allInferences.Remove(ctx, key); err != nil {
-			return err
-		}
-	}
-
-	return nil
+	return k.allInferences.Clear(ctx, blockRange)
```
relyt29, line 1891: the type of `Inferences` is a `[]*Inference` - I hope there's no memory leaks in this, because I could see it being likely

but golang is garbage collected so maybe it's ok? i'm not sure

relyt29 (approved):
broadly speaking I approve this PR and I like it but I do think we should make a note to look back when we have time and make sure that these nested map clearing functions are correctly freeing the memory to pointers to arrays

or as I said in the comment above maybe golang is garbage collected and it's fine - At some point further down the line when we have people running full nodes and we have a sense for the memory consumption that would be a more appropriate time to worry about memory leaks from this segment of the code

relyt29:
https://vivasoftltd.com/golang-mistakes-1-maps-and-memory-leaks/
https://stackoverflow.com/questions/39392467/in-go-deleting-an-entry-of-a-map-of-pointers-causes-memory-leak

we might be ok but I just don't know, needs more in depth research and probably the best way to confirm would be to make a mocked example and test the actual memory leak doesn't occur 

either way, future work - I approve this pr

PR #320, opened 2024-05-30
Use golines tool to enforce line length of 100 in repo, enforce via Github Action CI

relyt29:
Opinionated change obviously, but some of the long lines in this repo have become excessive to the point of making the code harder to read.

This PR adds the [golines](https://github.com/segmentio/golines) tool to the repo to make sure that lines are not too long. It modifies our CI action to also check line length is less than 100 characters. The reason why we need a separate tool is because the golang devs have [deliberately chosen not to enforce line length in their gofmt tool](https://github.com/golang/go/issues/11915) even though they make all sorts of other arbitrary decisions for users, like using tabs instead of spaces.

to run and fix issues yourself you can do something like this from the root of the repo:

```
find . -name "*.go" -exec golines -w {} \;
```

### DIFF
--- a/.github/workflows/format_and_test.yml
+++ b/.github/workflows/format_and_test.yml
@@ -32,6 +32,12 @@ jobs:
           gofmt-path: '.'
           gofmt-flags: '-l -d'
 
+    - name: Install golines tool
+      run: go install github.com/segmentio/golines@latest
+
+    -name: Check code long line formatting using golines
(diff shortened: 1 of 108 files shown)


br4e:
I'm a fan, I thought our lines were getting too long

relyt29:
putting this on hold for a while until we get a more stable repo

PR #326, opened 2024-06-01
update params

kpeluso:


xmariachi (approved):
lgtm, esp the epochLength. 
epochLength < 12 could be helpful for testing only (that's the only caveat we can think ab here - we may want to change this for integ/stress testing)

PR #329, opened 2024-06-03
revamp synth to handle variation in regrets

kpeluso:
## Questions
* rm f_tolerance ??
	* Seems to be doing diff things from epsilon?
	* pros/cons. Having both gives us control over diff things. we do use epsilon quite widely as it is
	* => would like everyone's opinion here
* reuse palette functions in in forecast_implied setting ??
	* Should we reuse the following functions in `CalcForecastImpliedInferences()` of `synth_palette_forecast_implied.go` ??
	* `weightAccumulator()`
	* `CalcWeightedInference()`
* ~~**double-check that the palettes don't mutate with each builder step**~~
	* ~~Guilherme largely fixed this in his commit but still something to lookout for~~
* ~~**allPeersAreNew — which to use ??**~~
	* ~~this is related to the edge case tickets~~
	* ~~see: `weightAccumulator()` in `synth_palette_weight.go`~~
	* ~~should we use `allPeersAreNew` or dynamically use `allInferersAreNew` vs `allForecastersAreNew`~~

## Also todo
* Double-check comments and error logs on functions
* Double-check error handling strategy esp within the builder and where it is used
	* thus far just used in `GetNetworkInferencesAtBlock()`
* Test
	* unit test coverage galore
	* stress test again


relyt29:
> rm f_tolerance

If it's not in the paper, it shouldn't be in our codebase. I don't recall seeing that parameter in the paper and I just searched for it and don't see it ... so

br4e:
Im wondering where fTolerance came from then, but yeah happy to simplify and get rid of fTolerance

br4e:
I think for question 4 you have to check `allInferersAreNew || allForecastersAreNew` because otherwise inside `CalcWeightedInference` you are taking an average across all inferer and forecaster weights including some that are weighted properly and some that are weighted as "1". That's going to give unexpected results.

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
```diff
@@ -72,18 +89,19 @@ func (p *SynthPalette) CalcWeightsGivenWorkers() (RegretInformedWeights, error)
...
+// 	inferers []Worker,
+// 	forecastedRegrets map[string]StatefulRegret,
+// ) (RegretInformedWeights, error) {
+// 	if inferers != nil && len(inferers) > 0 && forecastedRegrets != nil {
+// 		p.InfererRegrets = forecastedRegrets
+// 		p.ForecasterRegrets = forecastedRegrets
+// 	}
+
+// 	return p.CalcWeightsGivenWorkers()
+// }
```
br4e, line 104: lets just delete this right?

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
```diff
@@ -36,14 +51,16 @@ func (p *SynthPalette) CalcWeightsGivenWorkers() (RegretInformedWeights, error)
...
-		if regretFrac.Gt(maxRegret) {
-			maxRegret = regretFrac
+	if len(p.ForecasterRegrets) > 0 {
+		for i, worker := range p.Forecasters {
+			regretFrac, err := p.ForecasterRegrets[worker].regret.Quo(stdDevRegretsPlusFTolerance)
+			if err != nil {
+				return RegretInformedWeights{}, errorsmod.Wrapf(err, "Error calculating regret fraction")
+			}
+			normalizedForecasterRegrets[worker] = regretFrac
+			if i == 0 || regretFrac.Gt(maxRegret) {
```
br4e, line 61: I don't think we want this `i == 0` else we are re-initializing the maxRegret

br4e (approved):

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
kpeluso: yes you are correct, idk why this got added back in 

cc @guilherme-brandao

PR #332, opened 2024-06-03
[ORA-1552] Change Mint Module to use same whitelist as Emissions Module

relyt29:
rather than using the standard cosmos-sdk `x/gov` module which is more tedious to use, for our testnet implementation just use the same whitelist we already use for the emissions module.


x/emissions/keeper/keeper.go
```diff
@@ -1821,7 +1821,7 @@ func (k Keeper) GetPreviousPercentageRewardToStakedReputers(ctx context.Context)
 
 /// WHITELISTS
 
-func (k *Keeper) IsWhitelistAdmin(ctx context.Context, admin ActorId) (bool, error) {
```
kpeluso, line 1824: surprised this is needed...

kpeluso (approved):

x/emissions/keeper/keeper.go
relyt29: yeah I don't really understand it fully but it's something about using golang `interface{}` and the reflect package which the `expected_keepers` use. Won't compile otherwise

PR #334, opened 2024-06-03
ORA1554 Thoroughly test revamp synth pr

br4e:
This is _so close_. All of the test values are doing the thing where they are basically right, except they are about 5% off of their correct values. 

I have noticed that the Forecast implied inferences are a little bit off, and that is throwing off all of the other values, EXCEPT for the naive network inference which is exactly correct, _which does not use forecast implied inferences_

If we can just figure out why forecast implied inferences are not returning the correct value, I think we've got it. 

I've looked at it a few times. Anyone else want to give it a look? You can run `TestForecastImpliedInferences` to test.


x/emissions/keeper/inference_synthesis/network_inference_builder_test.go
```diff
@@ -1,20 +1,24 @@
...
+	cosmosMath "cosmossdk.io/math"
 
 	// cosmosMath "cosmossdk.io/math"
 	storetypes "cosmossdk.io/store/types"
 	"github.com/allora-network/allora-chain/app/params"
 	alloraMath "github.com/allora-network/allora-chain/math"
+	alloratestutil "github.com/allora-network/allora-chain/test/testutil"
 	"github.com/allora-network/allora-chain/x/emissions/keeper"
-	inferencesynthesis "github.com/allora-network/allora-chain/x/emissions/keeper/inference_synthesis"
+	inference_synthesis "github.com/allora-network/allora-chain/x/emissions/keeper/inference_synthesis"
```
guilherme-brandao, line 21: using it without `_`   or camel case seems to be the right pattern, but nbd

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -24,16 +24,19 @@ type NetworkInferenceBuilder struct {
...
 func NewNetworkInferenceBuilderFromSynthRequest(
 	req SynthRequest,
-) *NetworkInferenceBuilder {
+) (*NetworkInferenceBuilder, error) {
 	paletteFactory := SynthPaletteFactory{}
-	palette := paletteFactory.BuildPaletteFromRequest(req)
+	palette, err := paletteFactory.BuildPaletteFromRequest(req)
+	if err != nil {
+		return nil, errorsmod.Wrapf(err, "Error building palette from request")
+	}
```
br4e, line 32: We should be returning errors

x/emissions/keeper/inference_synthesis/network_inferences.go
```diff
@@ -102,7 +102,11 @@ func GetNetworkInferencesAtBlock(
 				CNorm:               moduleParams.CNorm,
 			},
 		)
-		networkInferences = networkInferecesBuilder.CalcAndSetNetworkInferences().Build()
+		if err != nil {
+			ctx.Logger().Warn(fmt.Sprintf("Error constructing network inferences builder topic: %s", err.Error()))
+			return nil, err
+		}
```
br4e, line 108: We should be returning errors

x/emissions/keeper/inference_synthesis/synth_palette_factory.go
```diff
@@ -27,17 +28,20 @@ func (f *SynthPaletteFactory) BuildPaletteFromRequest(req SynthRequest) SynthPal
...
 	palette.AllWorkersAreNew = palette.AllInferersAreNew && palette.AllForecastersAreNew
 
-	return palette
+	// Populates: forecastImpliedInferenceByWorker,
+	err := palette.UpdateForecastImpliedInferences()
+	if err != nil {
+		return SynthPalette{}, err
+	}
+
+	return palette, nil
```
br4e, line 46: There were a lot of bugs caused by these operations being out of order, also we should return errors

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
```diff
@@ -15,8 +15,8 @@ func (p *SynthPalette) CalcWeightsGivenWorkers() (RegretInformedWeights, error)
 	if err != nil {
 		return RegretInformedWeights{}, errorsmod.Wrapf(err, "Error calculating standard deviation of regrets")
 	}
-	// Add epsilon to standard deviation
-	stdDevRegretsPlusEpsilon, err := stdDevRegrets.Abs().Add(p.Epsilon)
+	// Add f Tolerance to standard deviation
+	stdDevRegretsPlusEpsilon, err := stdDevRegrets.Abs().Add(p.FTolerance)
```
br4e, line 19: Use fTolerance in this equation as Guilherme suggested

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
```diff
@@ -111,7 +111,7 @@ func (p *SynthPalette) CalcWeightedInference(weights RegretInformedWeights) (Inf
 
 	for _, forecaster := range p.Forecasters {
 		runningUnnormalizedI_i, sumWeights, err = AccumulateWeights(
-			p.InferenceByWorker[forecaster],
+			p.ForecastImpliedInferenceByWorker[forecaster],
```
br4e, line 114: This is a big one!

br4e (commented):
noting important changes on this pr

x/emissions/keeper/inference_synthesis/synth_palette_factory.go
br4e, line 46: There were a lot of bugs caused by these operations being out of order, also we should return errors

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
```diff
@@ -111,7 +111,7 @@ func (p *SynthPalette) CalcWeightedInference(weights RegretInformedWeights) (Inf
...
 			weights.forecasters[forecaster],
 			p.ForecasterRegrets[forecaster].noPriorRegret,
 			p.AllForecastersAreNew,
```
br4e, line 117: This is a big one!

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
```diff
@@ -36,13 +36,13 @@ func (p *SynthPalette) CalcWeightsGivenWorkers() (RegretInformedWeights, error)
...
 		regretFrac, err := p.ForecasterRegrets[worker].regret.Quo(stdDevRegretsPlusEpsilon)
 		if err != nil {
 			return RegretInformedWeights{}, errorsmod.Wrapf(err, "Error calculating regret fraction")
 		}
 		normalizedForecasterRegrets[worker] = regretFrac
-		if i == 0 || regretFrac.Gt(maxRegret) {
+		if regretFrac.Gt(maxRegret) {
 			maxRegret = regretFrac
 		}
 	}
```
br4e, line 48: We don't want to re-initialize the max value

kpeluso (approved):

PR #336, opened 2024-06-03
Fix forecast implied inf tests

guilherme-brandao:


x/emissions/keeper/inference_synthesis/synth_palette_weight.go
```diff
@@ -1,31 +1,51 @@
...
 		return RegretInformedWeights{}, errorsmod.Wrapf(err, "Error calculating standard deviation of regrets")
 	}
-	// Add epsilon to standard deviation
-	stdDevRegretsPlusEpsilon, err := stdDevRegrets.Abs().Add(p.Epsilon)
+	fmt.Println("stdDevRegrets :::", stdDevRegrets)
+	// Add f_tolerance to standard deviation
+	stdDevRegretsPlusFTolerance, err := stdDevRegrets.Abs().Add(p.FTolerance)
 	if err != nil {
-		return RegretInformedWeights{}, errorsmod.Wrapf(err, "Error adding epsilon to standard deviation of regrets")
+		return RegretInformedWeights{}, errorsmod.Wrapf(err, "Error adding f_tolerance to standard deviation of regrets")
```
kpeluso, line 40: what's the rule of thumb or f_tolerance vs epsilon?

x/emissions/keeper/inference_synthesis/synth_palette_forecast_implied.go
```diff
@@ -75,10 +77,15 @@ func (p SynthPalette) CalcForecastImpliedInferences() (map[Worker]*emissionstype
 					if err != nil {
 						return nil, errorsmod.Wrapf(err, "error calculating network loss per value")
 					}
-					R_ik[j] = StatefulRegret{regret: R_ijk, noPriorRegret: false}
+					R_ik[j] = &StatefulRegret{regret: R_ijk, noPriorRegret: false}
+				}
+
+				if sortedInferersInForecast != nil && len(sortedInferersInForecast) > 0 {
```
kpeluso, line 83: Maybe this function should simply never be called if there are no inferers... that's what we try to enforce everywhere else

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
guilherme-brandao: I think we use `fTolerance` for some calculations where it's explicit in the paper that we need it, represented by `ε`. Epsilon we use for some additional situations/potential edge cases where we need to avoid `0`.

`fTolerance = 0.01`
`Epsilon = 0.0001`

x/emissions/keeper/inference_synthesis/synth_palette_forecast_implied.go
guilherme-brandao: which function are you talking about `CalcForecastImpliedInferences` or `CalcWeightsGivenWorkers` ?

kpeluso: both actually. if there are no inferers, we shouldn't have even ever accepted the worker payload

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
kpeluso: that's a p good rule of thumb. if we're instituting that now, please ensure it's universally enforced by the time we merge my PR

kpeluso (approved):

PR #337, opened 2024-06-04
Cover regrets/weights new participants edge cases

guilherme-brandao:
Consider the edge cases:
- If all inferers are new, ignore the forecasters, and set the weights of all inferers to 1, irrespectively of the number of not-new forecasters.
- If one inferer is not new and all other inferers are new, set the weight of the not-new inferer to 1, and all others (including the forecasters, because we don’t need them anyway in this case) get a zero weight.
- If >1 inferer is not new, calculate weight as in the paper for all not-new workers (inferer and forecaster), and don’t include any new workers in the regret calculation.

### DIFF
--- a/x/emissions/keeper/inference_synthesis/synth_palette_weight.go
+++ b/x/emissions/keeper/inference_synthesis/synth_palette_weight.go
@@ -65,33 +65,33 @@ func (p *SynthPalette) CalcWeightsGivenWorkers() (RegretInformedWeights, error) 
 		}
 	}
 
-	moreThanOneNotNewInferer := p.SingleInfererNotNew == ""
-
-	// Calculate the weights from the normalized regrets
 	infererWeights := make(map[Worker]Weight)
-	for _, worker := range p.Inferers {
-		// If there is more than one not-new inferer, calculate the weight for the ones that are not new
-		var infererWeight = alloraMath.ZeroDec()
-		if moreThanOneNotNewInferer && !p.InfererRegrets[worker].noPriorRegret {
-			infererWeight, err = CalcWeightFromNormalizedRegret(normalizedInfererRegrets[worker], maxRegret, p.PNorm, p.CNorm)
-			if err != nil {
-				return RegretInformedWeights{}, errorsmod.Wrapf(err, "Error calculating inferer weight")
-			}
-		}
-		infererWeights[worker] = infererWeight
-	}
-
 	forecasterWeights := make(map[Worker]Weight)
-	if len(p.ForecasterRegrets) > 0 {
-		for _, worker := range p.Forecasters {
-			var forecasterWeight = alloraMath.ZeroDec()
-			if moreThanOneNotNewInferer && !p.ForecasterRegrets[worker].noPriorRegret {
-				forecasterWeight, err = CalcWeightFromNormalizedRegret(normalizedForecasterRegrets[worker], maxRegret, p.PNorm, p.CNorm)
+	if p.SingleInfererNotNew {
+		// Calculate the weights from the normalized regrets
+		for _, worker := range p.Inferers {
+			// If there is more than one not-new inferer, calculate the weight for the ones that are not new
+			var infererWeight = alloraMath.ZeroDec()
+			if !p.InfererRegrets[worker].noPriorRegret {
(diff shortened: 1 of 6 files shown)


x/emissions/keeper/inference_synthesis/synth_palette_weight.go
```diff
@@ -99,31 +111,48 @@ func (p *SynthPalette) CalcWeightedInference(weights RegretInformedWeights) (Inf
...
-		runningUnnormalizedI_i, sumWeights, err = AccumulateWeights(
-			p.InferenceByWorker[inferer],
-			weights.inferers[inferer],
-			p.InfererRegrets[inferer].noPriorRegret,
-			p.AllInferersAreNew,
-			runningUnnormalizedI_i,
-			sumWeights,
-		)
+	// If there is only one not-new inferer, then just consider that inferer
+	if p.SingleInfererNotNew != "" {
```
kpeluso, line 115: it's not wrong this way but I would prefer to use the booleans as booleans as opposed to derivative properties such as this (this is only populated if those booleans are of a certain value)

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
```diff
@@ -64,23 +65,34 @@ func (p *SynthPalette) CalcWeightsGivenWorkers() (RegretInformedWeights, error)
...
 
 	forecasterWeights := make(map[Worker]Weight)
-	for _, worker := range p.Forecasters {
-		forecasterWeight, err := CalcWeightFromNormalizedRegret(normalizedForecasterRegrets[worker], maxRegret, p.PNorm, p.CNorm)
-		if err != nil {
-			return RegretInformedWeights{}, errorsmod.Wrapf(err, "Error calculating forecaster weight")
+	if len(p.ForecasterRegrets) > 0 {
+		for _, worker := range p.Forecasters {
+			var forecasterWeight = alloraMath.ZeroDec()
+			if moreThanOneNotNewInferer && !p.ForecasterRegrets[worker].noPriorRegret {
```
kpeluso, line 88: for the check of `moreThanOneNotNewInferer` here and above, seems like you can simply skip both loops entirely if it's false, then whenever you check the weight you just check for inclusion before you do calculations with the weight. this fits with the pattern we established of using `palette.[inferers | forecasters]` to guide computation in all steps

x/emissions/keeper/inference_synthesis/types.go
```diff
@@ -68,13 +68,12 @@ type SynthPalette struct {
...
-	AllForecastersAreNew bool
-	AllWorkersAreNew     bool // Simple conjunction of the two above
-	NetworkCombinedLoss  Loss
-	Epsilon              alloraMath.Dec
-	FTolerance           alloraMath.Dec
-	PNorm                alloraMath.Dec
-	CNorm                alloraMath.Dec
+	ForecasterRegrets   map[Worker]*StatefulRegret
+	AllInferersAreNew   bool
+	SingleInfererNotNew Worker
```
kpeluso, line 73: i get that it's more efficient this way to cache this value. it does feel cleaner to check this in the accumulator using the existing "noPriorRegret" prop. I can go either way

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
```diff
@@ -64,23 +65,34 @@ func (p *SynthPalette) CalcWeightsGivenWorkers() (RegretInformedWeights, error)
 		}
 	}
 
+	moreThanOneNotNewInferer := p.SingleInfererNotNew == ""
```
kpeluso, line 68: you use this condition check in multiple places. You should probs have a palette method for it

kpeluso (approved):
approving because it accurately implements the intentions. my previous comments can just possibly make the code better

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
guilherme-brandao: Yeah, also thought about this but was opposed to adding a new property for the palette just for that... maybe it's a better idea to add...

br4e:
Proposed changes to address kenny's comments and also my own concerns: 

see here https://github.com/allora-network/allora-chain/pull/338/files

PR #341, opened 2024-06-05
WIP [ORA-1561] Perform Emissions module software upgrade in integration test

relyt29:
Given that 329 is already going to be merged pre edgenet, we instead create a software upgrade to do a mock migration of state in the emissions module as an integration test. This code can then be used as a reference the next time we need to do a _real_ software upgrade live on chain.

Contributes to ORA-1566 ORA-1567 ORA-1568 ORA-1569 ORA-1570 ORA-1580 ORA-1585


relyt29:
the 251 commits are because this branch was originally built off of `major-upgrade` rather than `main`. The first commit that represents actual work in this PR is 6d1b267 - there are not actually 250+ commits, only like 10.

In any case non of this is a real issue as when the PR is squashed and merged, it will appear as only one commit on the `main` history.

kpeluso (approved):

RedBird96 (approved):
lgtm

PR #344, opened 2024-06-06
[ORA-1510] Compute actor properties from workers and reputers

RedBird96:
Simulation test is added in this PR

Reference: https://linear.app/upshot/issue/ORA-1474/recreate-most-important-graphs-from-research-to-prove-implementation


test/simulation/simulation_test.go
```diff
@@ -0,0 +1,40 @@
...
+	)
+
+	t.Log(">>> Setup Topic <<<")
+	topicId := SetupTopic(testConfig)
+	t.Log(">>> Generate Actors <<<")
+	GenerateActors(testConfig, inferersCount, forecastersCount, reputersCount)
+	t.Log(">>> Register and Stake Topic <<<")
+	RegisterAndStakeTopic(testConfig, inferersCount, forecastersCount, reputersCount, topicId)
+	t.Log(">>> Repute Simulation <<<")
+	ReputeSimulation(testConfig, seed, iterationCount, inferersCount, forecastersCount, reputersCount, topicId)
```
kpeluso, line 39: to recreate the most important graphs I would have figured you would setup the topic and register the actors once, then let it iterate from there, adding funding and checking things as it goes

xmariachi:
Please add a simulation/README.md file on how to run and what's expected to get out, ty

test/simulation/simulation_reputer.go
```diff
@@ -0,0 +1,41 @@
+package simulation
+
+import (
+	testCommon "github.com/allora-network/allora-chain/test/common"
+	"time"
+)
+
+const EpochLength = 5
```
xmariachi, line 8: Now minimum EpochLength = 60 on chain, so this fails to create a topic.
Ofc this can be reduced only if the parameter that sets the minimum epochLength also has to be explicitly reduced - so this can be added (also to stress and integration tests btw) , a check where if the EpochLength received is under the minimum, reset the minimum to that value, just to accomodate tests.

Also, this and other settings should be parameterised like in the stress tests. Here topic.EpochLength is defined in the simulation_reputer file which is not directly related.

RedBird96: Thanks for your review. And the reason I didn't parameterize it is because the goal of this simulation is to generate a graph to verify that the loss calculation works as we researched. Ofc, if all settings are parameterized, it can be very helpful to generate several graphs to check the logic.

xmariachi:
Another comment, having run the tests successfully, the files generated are not obvious to read for me. They could have some context of what they mean, columns, etc. 

image
image

xmariachi:
Destination for this PR should be  `main` now, not `major-upgrade` :)

RedBird96:
> Another comment, having run the tests successfully, the files generated are not obvious to read for me. They could have some context of what they mean, columns, etc.
> 
> image image

RedBird96:
> Another comment, having run the tests successfully, the files generated are not obvious to read for me. They could have some context of what they mean, columns, etc.
> 
> image image

Yes, I was going to add headers....:)

relyt29:
@RedBird96 Is this PR still relevant? can it be closed?

PR #345, opened 2024-06-06
replace legacy event with typed events

kpeluso:
I made this PR because I thought the issue in our lack of event emission was that I was using a "legacy event" version (even though that seems to be the more widely used version).

However, it turns out we just weren't listening for all possible events in our indexer. We need to do this:
https://docs.cosmos.network/main/learn/advanced/events#subscribing-to-events
for `NewBlock` events.

So, now we can decide what we like working with more, then either:
* close this PR
* or merge this PR, move the tag to this commit, then upgrade the chain

Both are easy so depends what format and experience we like more.
If there's no discernible difference then I lean this "typed event" version because "legacy bad"


xmariachi (approved):
lgtm
Legacy events (via `EmitEvent`) are explicitly deprecated, so I lean towards typed events.

RedBird96 (approved):

kpeluso:
[This is](https://github.com/allora-network/allora-chain/pull/345/commits/d76eef218fae67b619d906c05be80b44b6ceca99) the first commit. The other commits exist because this PR was made before the original base branch `major-upgrade` was merged to `main`.

PR #348, opened 2024-06-10
Add nil dereference coverage and error control

xmariachi:
Add coverage against unexpected nil pointer dereferences.
It will throw error in that case.

Note: This is a very conservative take here, sending error if there is any - thus sending no weights; another option could be to ignore those that are nil but still return the completed ones, if any. But I'm unsure of the consequences, so rather return error and handle as such.


relyt29 (approved):
looks appropriate to me

relyt29:
running stress tests against this branch now

PR #350, opened 2024-06-12
Remove unnecessary struct and fixes

guilherme-brandao:
- Remove unnecessary `TopicFeeRevenue` struct which we weren't using the `Epoch` property;
- Improve error handling in some keeper functions;
- Fixes

Ticket: https://linear.app/upshot/issue/ORA-1603/devnet-nil-pointer-dereference-getcurrenttopicweight


x/emissions/keeper/keeper.go
```diff
@@ -1419,23 +1419,22 @@ func (k *Keeper) DripTopicFeeRevenue(ctx context.Context, topicId TopicId, block
...
+
 	moduleParams, err := k.GetParams(ctx)
 	if err != nil {
 		return err
 	}
 	epsilon := moduleParams.Epsilon
 	topicFeeRevenueDecayRate := moduleParams.TopicFeeRevenueDecayRate
-	topicFeeRevenueDec, err := alloraMath.NewDecFromSdkInt(topicFeeRevenue.Revenue)
+
+	topicFeeRevenueDec, err := alloraMath.NewDecFromSdkInt(topicFeeRevenue)
```
xmariachi, line 1430: By the way, I think I this "pattern" could be the source of other `nil dereference` errors: the `err` is not checked to be nil (!). 

So in case of error, the returned object may be nil, hence triggering a nil dereference in our face. 
Can you please check this in other parts of the changes of the new synth especially?

xmariachi (changes requested):
* Requested changes on err handling 
* re. the overall intention of the PR, lgtm - my only concern is that if sometimes those `struct`s could be nil, now they may come as default value , ie. `0`, thus causing differences. 
Please check if that would have any impact (maybe not, but pls check)

relyt29 (approved):
this looks good to me

x/emissions/keeper/keeper.go
```diff
@@ -1251,8 +1251,11 @@ func (k *Keeper) GetReputerAddressByP2PKey(ctx context.Context, p2pKey string) (
 // Returns ((0,0), true) if there was no prior topic weight set, else ((x,y), false) where x,y!=0
 func (k *Keeper) GetPreviousTopicWeight(ctx context.Context, topicId TopicId) (alloraMath.Dec, bool, error) {
 	topicWeight, err := k.previousTopicWeight.Get(ctx, topicId)
-	if errors.Is(err, collections.ErrNotFound) {
-		return alloraMath.ZeroDec(), true, nil
+	if err != nil {
+		if errors.Is(err, collections.ErrNotFound) {
+			return alloraMath.ZeroDec(), true, nil
```
xmariachi, line 1256: Is this used down the road in a division? 
If so then it should not be 0 but epsilon. 
But if that's not the case, 0 is ok I guess.

xmariachi (approved):
lgtm, though I left a comment on default values to be reviewed.

x/emissions/keeper/keeper.go
guilherme-brandao: It is not. We are handling potential division by zero cases in the function scope, not in keepers atm

PR #352, opened 2024-06-13
tobi/ora-1626-add-restore-snapshot-to-l1_node-for-nops

okedeji:


scripts/l1_node.sh
```diff
@@ -1,18 +1,21 @@
 #!/bin/bash
-set -exu
+# set -exu
```
relyt29, line 2:
```
       -x    The shell shall write to standard error a trace for each command after it expands the command and before it executes it. It is unspecified whether the command that turns tracing off is traced.
```

i'm fine with deleting, but lets delete this commented out line

scripts/l1_node.sh
```diff
@@ -1,18 +1,21 @@
...
-APP_HOME="${APP_HOME:-/data}"
+export APP_HOME="${APP_HOME:-./data}"
 INIT_FLAG="${APP_HOME}/.initialized"
 MONIKER="${MONIKER:-$(hostname)}"
 KEYRING_BACKEND=test                              #! Use test for simplicity, you should decide which backend to use !!!
 GENESIS_FILE="${APP_HOME}/config/genesis.json"
 DENOM="uallo"
 
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+"${SCRIPT_DIR}/restore_snapshot.sh"
```
relyt29, line 17: I am a little nervous about this line

l1_node is set as the entrypoint for `docker-compose.yaml` 

the local_testnet_l1.sh script uses docker compose, my only concern is I don't want the script that devs use locally to test their code for e.g. integration, stress tests, to be pushing out or pulling down s3 snapshots every time the local_testnet_l1.sh runs - especially if they don't have the keys/token to access s3

Are you sure this will restore_snapshot will not execute if a dev is running a local testnet?

scripts/l1_node.sh
okedeji: not sure where this comment is gotten from? not in this code

relyt29: it's present in the diff, you commented out line 2?

okedeji: yeah I mean the "The shell shall write to standard error a trace for each command after it expands ..."

okedeji: I commented it out just for convience for whoever want to debug and see what variable is what

scripts/l1_node.sh
okedeji: So we may just separete concerns maybe? initialy the docker-compose was meant to be a way for nops to run node easily right? if we now use that for integration test, then we probably have to provide another means by which nops run the node.

or maybe I can make it optional. i.e commented off and added if needed?

scripts/l1_node.sh
okedeji: removed!

scripts/l1_node.sh
okedeji: Did the latter!

relyt29 (approved):

PR #354, opened 2024-06-13
[ORA-1620] Clear data from stake removal queue after a ConfirmStakeRemoval call is completed

relyt29:
This prevents an attacker from being able to withdraw stake multiple times after waiting the delay period once.


br4e (approved):
looks right to me! This pr is huge for what is ultimately a simple change

RedBird96 (approved):

PR #355, opened 2024-06-13
[ORA-1621] Implement stake removal active window - requests to withdraw stake after window expires fail

relyt29:
Mitigates issues where an attacker waits out a full stake withdrawal delay, but then permanently after that time they are able to chose any time of their liking to withdraw stake.

Now stake withdrawals have an "active window" by default 48 hours worth of blocks, where they are able to withdraw stake. If the withdrawal window expires, a staker has to restart the stake withdrawal process and wait out the delay once more.


RedBird96 (approved):
looks good to me!

x/emissions/keeper/msgserver/msg_server_stake.go
```diff
@@ -111,9 +111,14 @@ func (ms msgServer) ConfirmRemoveStake(ctx context.Context, msg *types.MsgConfir
 	if err != nil {
 		return nil, err
 	}
-	if stakeRemoval.BlockRemovalStarted+moduleParams.RemoveStakeDelayWindow >= currentBlock {
+	// if the delay window has not yet passed, its too early
+	if currentBlock < stakeRemoval.BlockRemovalStarted+moduleParams.RemoveStakeDelayWindow {
```
br4e, line 115: if you wanted you could calculate this sum once (and re-use in the next statement), it would also look cleaner

x/emissions/keeper/msgserver/msg_server_stake.go
```diff
@@ -243,9 +248,14 @@ func (ms msgServer) ConfirmRemoveDelegateStake(ctx context.Context, msg *types.M
...
```
br4e, line 252: same here

br4e (approved):
looks good to me!

xmariachi (approved):
lgtm

br4e:
SO MUCH APPROVAL

relyt29:
After conversation with Alejandro, I think I will close this PR and instead propose one that brings emissions-based reputer staking in line with /x/staking based validator staking - where the endBlocker unstakes people rather than having to have a second transaction. This is better UX for the users, brings us in line with cosmos-sdk staking, and is also a solution to this "attacker controls when the unstake happens" attack

PR #356, opened 2024-06-13
ORA1635 Investigate testing out normalized regret calculation for a single regret

br4e:


br4e:
Not able to test, see task for more info

PR #361, opened 2024-06-18
update docker-compose to point to edgenet

okedeji:


relyt29 (approved):
i hope we have a run book somewhere where we're tracking all the little details required for every new release

PR #362, opened 2024-06-19
[ORA-1650] Use endblock to expire stake removals instead of a second transaction

relyt29:
This brings us more in line with cosmos validator staking which is how that works

Summary of Changes:
* EndBlock removes pending stake removal requests (`x/emissions/module/abci.go` and `x/emissions/module/stake_removals.go`)
* Keeper now stores pending stake removals in two ways, "forwards" (indexed by block, for the end block removal code) and "backwards" (indexed by reputer/delegator, for user queries, and for replacement /canceling of stake removal requests)
    * New Key datatype `Quadruple.go` that mimics collections.Triple to support this indexing
* Change msg server function `RemoveStake` to now request a stake removal. If a stake removal is pending, sending another stake removal request replaces your existing request and resets the delay
    * Add `CancelRemoveStake` function to cancel removing your stake during the delay window
* Add query functions to see what stake removals are pending, indexed by block (`GetStakeRemovalsForBlock`) or by reputer/delegator (`GetStakeRemovalInfo`)
* Bunch of Protobuf messages renamed or tweaked as appropriate

Note: this code _requires_ a state migration and so will have to be implemented via governance proposal and upgrade module

The state migration is not yet done, will flip the PR to ready for review when it is


guilherme-brandao (commented):
Looking good! Automatic removal is way better UX as well. 

I was trying to find a argument/alternative for not using quadruple keys, but it seems necessary indeed for a more efficient automatic removal.

relyt29:
Ok migrations written
* New upgrade in the app/upgrades folder, and corresponding upgrade handler. The name of the upgrade is "v0.2.8" for the proposed governance proposal
* Bump ConsensusVersion in /x/emissions
* New migration in /x/emissions/migrations. All it does is `Clear` delete all the data in the old stakeRemovals map. New removals can go through the new functions which don't touch the old data structure.
    * In the _next_ upgrade after this one, the data structures can be completely deleted from the repo because this migration has to clear them, so they have to be present in the code to be cleared. After clearing, they can be deleted. Added comments saying the data structures are deprecated and scheduled to be deleted in the next upgrade.

x/emissions/keeper/keeper.go
```diff
@@ -811,6 +831,7 @@ func (k *Keeper) AddDelegateStake(ctx context.Context, topicId TopicId, delegato
 // Removes stake to the system for a given topic and reputer
```
br4e, line 831: from the system?

x/emissions/keeper/keeper.go
```diff
@@ -1139,44 +1161,205 @@ func (k *Keeper) SetDelegateStakeUponReputer(ctx context.Context, topicId TopicI
...
-func (k *Keeper) SetStakeRemoval(ctx context.Context, address ActorId, removalInfo types.StakePlacement) error {
-	key := collections.Join(removalInfo.TopicId, address)
-	return k.stakeRemoval.Set(ctx, key, removalInfo)
+func (k *Keeper) SetStakeRemoval(ctx context.Context, removalInfo types.StakeRemovalInfo) error {
+	byBlockKey := collections.Join3(removalInfo.BlockRemovalCompleted, removalInfo.TopicId, removalInfo.Reputer)
+	err := k.stakeRemovalsByBlock.Set(ctx, byBlockKey, removalInfo)
+	if err != nil {
+		return err
+	}
+	byActorKey := collections.Join3(removalInfo.Reputer, removalInfo.TopicId, removalInfo.BlockRemovalCompleted)
```
br4e, line 1173: This key is exactly the same as the above, if that was intentional, can you create one key and use it twice?

Or can you just set this data in one place instead of two?

x/emissions/keeper/keeper.go
```diff
@@ -1139,44 +1161,205 @@ func (k *Keeper) SetDelegateStakeUponReputer(ctx context.Context, topicId TopicI
...
+		return err
+	}
+	if !has {
+		return types.ErrStakeRemovalNotFound
+	}
+	err = k.stakeRemovalsByBlock.Remove(ctx, byBlockKey)
+	if err != nil {
+		return err
+	}
+	byActorKey := collections.Join3(address, topicId, blockHeight)
```
br4e, line 1196: same here

x/emissions/keeper/keeper.go
br4e: its also a little spooky to do it in a different order each time, what if there is a mistake?

x/emissions/keeper/keeper.go
```diff
@@ -1139,44 +1161,205 @@ func (k *Keeper) SetDelegateStakeUponReputer(ctx context.Context, topicId TopicI
...
+	keys, err := iter.Keys()
+	if err != nil {
+		return types.StakeRemovalInfo{}, false, err
+	}
+	keysLen := len(keys)
+	if keysLen == 0 {
+		return types.StakeRemovalInfo{}, false, nil
+	} else {
+		if keysLen > 1 {
+			ctx.Logger().Warn("Invariant failure! More than one stake removal found for reputer and topicId")
```
br4e, line 1251: ok cool

br4e: if we do get into this state though, will this prevent this reputer from ever removing their stake or something similar?

x/emissions/keeper/keeper.go
```diff
@@ -1139,44 +1161,205 @@ func (k *Keeper) SetDelegateStakeUponReputer(ctx context.Context, topicId TopicI
...
-	key := collections.Join3(removalInfo.TopicId, removalInfo.Reputer, removalInfo.Delegator)
-	return k.delegateStakeRemoval.Set(ctx, key, removalInfo)
+func (k *Keeper) SetDelegateStakeRemoval(ctx context.Context, removalInfo types.DelegateStakeRemovalInfo) error {
+	byBlockKey := Join4(removalInfo.BlockRemovalCompleted, removalInfo.TopicId, removalInfo.Delegator, removalInfo.Reputer)
+	err := k.delegateStakeRemovalsByBlock.Set(ctx, byBlockKey, removalInfo)
+	if err != nil {
+		return err
+	}
+	byActorKey := Join4(removalInfo.Delegator, removalInfo.Reputer, removalInfo.TopicId, removalInfo.BlockRemovalCompleted)
+	return k.delegateStakeRemovalsByActor.Set(ctx, byActorKey)
```
br4e, line 1273: same

x/emissions/keeper/keeper.go
```diff
@@ -1139,44 +1161,205 @@ func (k *Keeper) SetDelegateStakeUponReputer(ctx context.Context, topicId TopicI
...
+	}
+	if !has {
+		return types.ErrStakeRemovalNotFound
+	}
+	err = k.delegateStakeRemovalsByBlock.Remove(ctx, byBlockKey)
+	if err != nil {
+		return err
+	}
+	byActorKey := Join4(delegator, reputer, topicId, blockHeight)
+	return k.delegateStakeRemovalsByActor.Remove(ctx, byActorKey)
```
br4e, line 1297: same

x/emissions/keeper/keeper.go
```diff
@@ -1139,44 +1161,205 @@ func (k *Keeper) SetDelegateStakeUponReputer(ctx context.Context, topicId TopicI
...
+	keys, err := iter.Keys()
+	if err != nil {
+		return types.DelegateStakeRemovalInfo{}, false, err
+	}
+	keysLen := len(keys)
+	if keysLen == 0 {
+		return types.DelegateStakeRemovalInfo{}, false, nil
+	} else {
+		if keysLen > 1 {
+			ctx.Logger().Warn("Invariant failure! More than one delegate stake removal found for delegator, reputer and topicId")
```
br4e, line 1353: same

br4e (commented):
lots of comments

x/emissions/keeper/msgserver/msg_server_stake.go
```diff
@@ -78,62 +76,54 @@ func (ms msgServer) StartRemoveStake(ctx context.Context, msg *types.MsgStartRem
...
-	return &types.MsgStartRemoveStakeResponse{}, nil
-}
 
-// Function for reputers or workers to call to remove stake from an existing stake position.
-func (ms msgServer) ConfirmRemoveStake(ctx context.Context, msg *types.MsgConfirmRemoveStake) (*types.MsgConfirmRemoveStakeResponse, error) {
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
-	// Pull the stake removal from the delayed queue
-	stakeRemoval, err := ms.k.GetStakeRemovalByTopicAndAddress(ctx, msg.TopicId, msg.Sender)
+	// find out if we have a stake removal in progress, if so overwrite it
+	removal, found, err := ms.k.GetFirstStakeRemovalForReputerAndTopicId(sdkCtx, msg.Sender, msg.TopicId)
```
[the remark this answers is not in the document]
relyt29: as long as there are no bugs they'll only have 1 :joy:

x/emissions/keeper/quadruple.go
```diff
@@ -0,0 +1,389 @@
...
+	k4 *K4
+}
+
+// Join4 instantiates a new Quadruple instance composed of the three provided keys, in order.
+func Join4[K1, K2, K3, K4 any](k1 K1, k2 K2, k3 K3, k4 K4) Quadruple[K1, K2, K3, K4] {
+	return Quadruple[K1, K2, K3, K4]{&k1, &k2, &k3, &k4}
+}
+
+// K1 returns the first part of the key. If nil, the zero value is returned.
+func (t Quadruple[K1, K2, K3, K4]) K1() (x K1) {
```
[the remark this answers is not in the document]
relyt29: the code it's based off is 1 indexed, I'm inclined to leave it this way to minimize the diff from https://github.com/cosmos/cosmos-sdk/blob/main/collections/triple.go

x/emissions/keeper/quadruple.go
```diff
@@ -0,0 +1,389 @@
...
+}
+
+// NewTriplePrefixedQuadrupleRange provides a Range for all keys prefixed with the given
+// first, second, and third parts of the Quadruple key.
+func NewTriplePrefixedQuadrupleRange[K1, K2, K3, K4 any](k1 K1, k2 K2, k3 K3) collections.Ranger[Quadruple[K1, K2, K3, K4]] {
+	key := QuadrupleTriplePrefix[K1, K2, K3, K4](k1, k2, k3)
+	r := &collections.Range[Quadruple[K1, K2, K3, K4]]{}
+	r = r.Prefix(key)
+	return r
+}
```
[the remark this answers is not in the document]
relyt29: see comment at top of file but yea https://github.com/cosmos/cosmos-sdk/blob/main/collections/triple.go

x/emissions/migrations/v2/migrations.go
```diff
@@ -3,16 +3,27 @@ package v2
...
 			"### MIGRATING STORE FROM VERSION 1 TO VERSION 2 ###" +
 			"###################################################",
 	))
-	return nil
+	err := stakeRemoval.Clear(ctx, nil)
+	if err != nil {
+		return err
+	}
+	return delegateStakeRemoval.Clear(ctx, nil)
 }
```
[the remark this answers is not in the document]
relyt29: that's correct, we're changing from one data structure to another for the stake removals

given that we're performing this on a testnet with little to no activity, it's simpler to just delete the removals and tell people to restart them rather than actually migrate them. It's 99% likely that there will be no removals in the state machine anyways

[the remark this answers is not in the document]
relyt29: (if we were doing this on mainnet, I wouldn't be so cavalier, and instead would copy the removals)

x/emissions/keeper/keeper.go
relyt29: The order is what's important here, they're two different types of keys in the [IAVL store](https://github.com/cosmos/iavl/blob/master/docs/overview.md). In the first case you're looking up ("Alice", "Topic 1", "Block 356") and in the second case you're looking up ("Block 356", "Topic 1", "Alice").

My understanding of the system is basically that keys are concatenated together and then put into a tree data structure. So changing the order of the components that make up the key completely changes the resulting location in the tree - ("Alice", "Topic 1", "Block 356") may end up in the left most position in one tree and ("Block 356", "Topic 1", "Alice") will end up in the middle or something like that

We care about the ability to iterate over members of the data store so we need to be able to traverse the tree both by block and also by delegator/reputer. Hence the "Actor" and "Block" suffixes. I'd like to have just one data structure but I know of no efficient way to traverse the tree to iterate over the values in the data store without having to either go through every available block which has a removal in the case of looking up an actor, or iterate over every actor looking for a specific block.

This is akin to solidity having two mappings forwards and backwards to store the same data in an O(1) lookup indexed by different keys

x/emissions/keeper/keeper.go
relyt29: I believe based on the way the code is written, if the invariant is broken, this will allow someone to remove stake more quickly than the delay would have otherwise permitted.

The alternative would be to prevent someone from ever removing stake which seems maybe worse?

x/emissions/keeper/queryserver/query_server_stake.go
```diff
@@ -114,3 +115,80 @@ func (qs queryServer) GetTopicStake(ctx context.Context, req *types.QueryTopicSt
...
+		maxLimit = moduleParams.MaxPageLimit
+	}
+	if uint64(len(removals)) > maxLimit {
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("cannot query more than %d removals at once", maxLimit))
+	}
+	removalPointers := make([]*types.StakeRemovalInfo, len(removals))
+	for i, removal := range removals {
+		removalPointers[i] = &removal
+	}
+	return &types.QueryStakeRemovalsForBlockResponse{Removals: removalPointers}, nil
```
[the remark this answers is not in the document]
relyt29: so I'm with you that it's weird but this is how the autogenerated types come out of protobuf, since it's a complex type (struct StakeRemovalInfo) protobuf returns that into a pointer. When you actually query via the CLI it returns the data, though:

```yaml
$ allorad --home=test/devnet/validator0 query emissions stake-removals-for-block 362936

removals:
- amount: "7048718058408485432"
  block_removal_completed: "362936"
  block_removal_started: "56"
  reputer: allo1vja3ec4a707znzlzhs4pyqw8m5y98euwka5uf5
  topic_id: "1"
```

so I think not an issue

relyt29:
Going to add ORA-1691 tomorrow morning to this, the reputer tests are done, but we never had delegation staking tests from the integration standpoint written apparently. Will do in morning

x/emissions/keeper/keeper.go
br4e: as long as its not possible to lock them out of taking an action they should be able to take

x/emissions/types/params.go
```diff
@@ -14,7 +14,7 @@ func DefaultParams() Params {
 		MinTopicWeight:                  alloraMath.MustNewDecFromString("100"),    // total weight for a topic < this => don't run inference solicatation or loss update
 		MaxTopicsPerBlock:               uint64(128),                               // max number of topics to run cadence for per block
 		RequiredMinimumStake:            cosmosMath.NewInt(100),                    // minimum stake required to be a worker or reputer
-		RemoveStakeDelayWindow:          int64(60 * 60 * 24 * 7 * 3),               // number of seconds to wait before finalizing a stake withdrawal
+		RemoveStakeDelayWindow:          int64(60 * 60 * 24 * 7 * 3 / 5),           // ~approx 3 weeks assuming 5 second block time, number of blocks to wait before finalizing a stake withdrawal
```
br4e, line 17: math like this where the order of operations is potentially ambiguous always creeps me out, but I suppose this is fine. If it were me I would put parentheses around everything except `/ 5`

x/emissions/keeper/genesis.go
```diff
@@ -17,6 +17,9 @@ func (k *Keeper) InitGenesis(ctx context.Context, data *types.GenesisState) erro
 	k.authKeeper.SetModuleAccount(ctx, alloraRewardsModuleAccount)
 	alloraPendingRewardsModuleAccount := k.authKeeper.GetModuleAccount(ctx, types.AlloraPendingRewardForDelegatorAccountName)
 	k.authKeeper.SetModuleAccount(ctx, alloraPendingRewardsModuleAccount)
+	if err := k.SetParams(ctx, data.Params); err != nil {
+		return err
+	}
```
br4e, line 22: so the genesis params weren't actually being set before right?

x/emissions/keeper/keeper.go
br4e: oh oops I ddin't realize that this doesn't return an error. Should it though, since it's an "invariant" ?

x/emissions/keeper/keeper.go
```diff
@@ -1139,44 +1161,205 @@ func (k *Keeper) SetDelegateStakeUponReputer(ctx context.Context, topicId TopicI
...
-func (k *Keeper) SetDelegateStakeRemoval(ctx context.Context, removalInfo types.DelegateStakePlacement) error {
-	key := collections.Join3(removalInfo.TopicId, removalInfo.Reputer, removalInfo.Delegator)
-	return k.delegateStakeRemoval.Set(ctx, key, removalInfo)
+func (k *Keeper) SetDelegateStakeRemoval(ctx context.Context, removalInfo types.DelegateStakeRemovalInfo) error {
+	byBlockKey := Join4(removalInfo.BlockRemovalCompleted, removalInfo.TopicId, removalInfo.Delegator, removalInfo.Reputer)
+	err := k.delegateStakeRemovalsByBlock.Set(ctx, byBlockKey, removalInfo)
+	if err != nil {
+		return err
+	}
+	byActorKey := Join4(removalInfo.Delegator, removalInfo.Reputer, removalInfo.TopicId, removalInfo.BlockRemovalCompleted)
```
br4e, line 1272: in the other places you had the keys in exact reverse order for block and actor key, that is not the case here. Do you want it to be?

x/emissions/keeper/msgserver/msg_server_stake.go
```diff
@@ -210,59 +199,69 @@ func (ms msgServer) StartRemoveDelegateStake(ctx context.Context, msg *types.Msg
...
+			return nil, errors.Wrap(err, "failed to delete previous delegate stake removal")
+		}
+	}
+	stakeToRemove := types.DelegateStakeRemovalInfo{
+		BlockRemovalStarted:   sdkCtx.BlockHeight(),
+		TopicId:               msg.TopicId,
+		Reputer:               msg.Reputer,
+		Delegator:             msg.Sender,
+		Amount:                msg.Amount,
+		BlockRemovalCompleted: sdkCtx.BlockHeight() + moduleParams.RemoveStakeDelayWindow,
```
br4e, line 231: maybe put this second to be consistent with the previous call

x/emissions/keeper/queryserver/query_server_stake.go
```diff
@@ -114,3 +115,80 @@ func (qs queryServer) GetTopicStake(ctx context.Context, req *types.QueryTopicSt
...
+	if err != nil {
+		return nil, err
+	}
+	maxLimit := types.DefaultParams().MaxPageLimit
+	moduleParams, err := qs.k.GetParams(ctx)
+	if err == nil {
+		maxLimit = moduleParams.MaxPageLimit
+	}
+	if uint64(len(removals)) > maxLimit {
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("cannot query more than %d removals at once", maxLimit))
```
br4e, line 156: can't you just return the ones that are less than the page limit? Or if there are too many that means you can't get them at all?

x/emissions/keeper/queryserver/query_server_stake.go
```diff
@@ -114,3 +115,80 @@ func (qs queryServer) GetTopicStake(ctx context.Context, req *types.QueryTopicSt
...
```
br4e, line 133: just return the ones less than the max limit?

br4e (approved):
lgtm! I trust you with the answers to the couple of open questions I posed

x/emissions/keeper/genesis.go
relyt29: shockingly it appears indeed they were not

it would just always use the defaults and literally ignore what was in the genesis.json

x/emissions/keeper/keeper.go
relyt29: actually we should be able to produce an error here you are right, because the CancelStake operation is exposed

so if the invariant is broken they should be able to cancel their stake until the invariant is no longer broken LOL

BUT BETTER TO NOT HAVE THE INVARIANT EVER BREAK

x/emissions/keeper/queryserver/query_server_stake.go
relyt29: this is just copied and pasted from how our other "pagination" query server functions work

but yeah the right thing to do would not be to deny the user at least some portion of their data

relyt29: i'll fix it here but keep in mind this is existing elsewhere in our query server functions

x/emissions/keeper/keeper.go
relyt29: I put up a PR that now makes the operation return an error

in the case of an error the cancel stake function now continues to let you cancel even if this specific error is observed

guilherme-brandao (approved):

relyt29:
I will merge this as soon as I simulate the upgrade happening on a forked edgenet

x/emissions/migrations/v2/migrations.go
```diff
@@ -3,16 +3,27 @@ package v2
...
+// the stakeRemoval and delegateStakeRemoval
+// key value stores
+func MigrateStore(
+	ctx sdk.Context,
+	stakeRemoval *collections.Map[collections.Pair[uint64, string], emissionsTypes.StakePlacement],
+	delegateStakeRemoval *collections.Map[collections.Triple[uint64, string, string], emissionsTypes.DelegateStakePlacement],
+) error {
+	ctx.Logger().Warn(fmt.Sprintf(
 		"###################################################" +
 			"### MIGRATING STORE FROM VERSION 1 TO VERSION 2 ###" +
```
kpeluso, line 21: no semver here?

x/emissions/migrations/v3/migrations.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// An example MigrateStore function that might be run in the migration
+// handler.
+func MigrateStore(ctx sdk.Context) error {
```
kpeluso, line 11: v3 needed?

kpeluso (approved):
Just know that we'll run the upgrade to include these as well:
* https://github.com/allora-network/allora-chain/pull/367
* https://github.com/allora-network/allora-chain/pull/366

Not that I think they effect state.
jsyk in case you have to update the upgrade logic at all (i don't think so)

x/emissions/migrations/v2/migrations.go
relyt29: doesn't appear that's how the modules themselves track versioning, i agree it's weird but it's in line with how other chains do it

https://github.com/evmos/evmos/blob/49cab07c39f64be0b260d67423c12380a475ce4f/x/evm/module.go#L28-L29

https://github.com/osmosis-labs/osmosis/blob/5b92bcb75ca8f9caae83a2a928d2426afa822f82/x/gamm/module.go#L155-L156

x/emissions/migrations/v3/migrations.go
relyt29: needed for integration testing upgrade flow

relyt29:
Integration tests are failing because we never actually wrote any functioning query function to let a reputer figure out how much stake they put on themselves in the past.


Will need to add that, since rewriting the entire staking model is not reasonable at this stage this late in the game with the time pressures we are up against

but we should in the future blow away this needless separation between "delegated staking" and "normal staking"

PR #363, opened 2024-06-19
Updating GetLastNetworkInference query and update old query

guilherme-brandao:
- `GetLastNetworkInference` was using `EpochLastEnded` as a reference to query the network inference, which is not optimal since this property is updated before the worker sends the inferences.

- Also removed a potentially unnecessary calculation in the `GetNetworkInferenceAtBlock` function. We're already calculating and setting the previous network combined loss and saving it during the losses insert from the reputer in the previous time step, so we just need to query it; we were calculating it again.

Ticket: https://linear.app/upshot/issue/ORA-1666/debug-getlatestnetworkinference-returning-empty-values-on-edgenet


x/emissions/keeper/inference_synthesis/network_losses.go
```diff
@@ -210,43 +210,6 @@ func CalcNetworkLosses(
 	return output, nil
 }
 
-// Same as CalcNetworkLosses() but just returns the combined loss
-func CalcCombinedNetworkLoss(
```
relyt29, line 214: is there a reason we're deleting this test, don't love deleting tests without a reason

x/emissions/keeper/inference_synthesis/network_losses_test.go
```diff
@@ -61,144 +61,6 @@ func (s *InferenceSynthesisTestSuite) TestRunningWeightedAvgUpdate() {
 	}
 }
 
-func (s *InferenceSynthesisTestSuite) TestCalcCombinedNetworkLossSimpleCaseWithOneReputer() {
```
relyt29, line 64: is there a reason we're deleting this test, don't love deleting tests without a reason

relyt29:
I'm not super familiar with this part of the codebase, but to my untrained eye it looks fine

x/emissions/keeper/inference_synthesis/network_losses_test.go
br4e: Same, does this test break with the changes?

x/emissions/keeper/inference_synthesis/network_losses.go
br4e: this isn't a test though right?

x/emissions/keeper/inference_synthesis/network_inferences.go
```diff
@@ -71,55 +69,131 @@ func GetNetworkInferencesAtBlock(
...
+				Epsilon:             moduleParams.Epsilon,
+				FTolerance:          moduleParams.FTolerance,
+				PNorm:               topic.PNorm,
+				CNorm:               moduleParams.CNorm,
+			},
 		)
 		if err != nil {
-			Logger(ctx).Warn(fmt.Sprintf("Error calculating network combined loss: %s", err.Error()))
-			return networkInferences, nil, infererWeights, forecasterWeights, nil
+			Logger(ctx).Warn(fmt.Sprintf("Error constructing network inferences builder topic: %s", err.Error()))
```
br4e, line 100: logger.warn here, but not in other places

x/emissions/keeper/inference_synthesis/network_inferences.go
```diff
@@ -71,55 +69,131 @@ func GetNetworkInferencesAtBlock(
...
+	}
+	if len(inferences.Inferences) > 1 {
+		moduleParams, err := k.GetParams(ctx)
+		if err != nil {
+			return nil, nil, infererWeights, forecasterWeights, err
 		}
 		topic, err := k.GetTopic(ctx, topicId)
 		if err != nil {
-			Logger(ctx).Warn(fmt.Sprintf("Error getting topic: %s", err.Error()))
+			fmt.Printf("Error getting topic: %s", err.Error())
```
br4e, line 170: it looks like we are doing loger.warn in some places and fmt.printf in others? Let's just pick on and stick to it (I don't have a strong opinion, but I think logger.warn is what we want right?)

x/emissions/keeper/inference_synthesis/network_inferences.go
```diff
@@ -71,55 +69,131 @@ func GetNetworkInferencesAtBlock(
...
+				FTolerance:          moduleParams.FTolerance,
+				PNorm:               topic.PNorm,
+				CNorm:               moduleParams.CNorm,
+			},
 		)
 		if err != nil {
-			Logger(ctx).Warn(fmt.Sprintf("Error calculating network combined loss: %s", err.Error()))
-			return networkInferences, nil, infererWeights, forecasterWeights, nil
+			Logger(ctx).Warn(fmt.Sprintf("Error constructing network inferences builder topic: %s", err.Error()))
+			return nil, nil, infererWeights, forecasterWeights, err
```
br4e, line 101: why can't we return network inferences here? we did above

x/emissions/keeper/inference_synthesis/network_inferences.go
```diff
@@ -71,55 +69,131 @@ func GetNetworkInferencesAtBlock(
...
-			stakeAmount, err := k.GetStakeOnReputerInTopic(ctx, topicId, bundle.ValueBundle.Reputer)
-			if err != nil {
-				Logger(ctx).Warn(fmt.Sprintf("Error getting stake on reputer: %s", err.Error()))
-				return networkInferences, nil, infererWeights, forecasterWeights, nil
-			}
-			stakesByReputer[bundle.ValueBundle.Reputer] = stakeAmount
+		topic, err := k.GetTopic(ctx, topicId)
+		if err != nil {
+			Logger(ctx).Warn(fmt.Sprintf("Error getting topic: %s", err.Error()))
+			return networkInferences, nil, infererWeights, forecasterWeights, nil
```
br4e, line 81: we returned networkInferences here, so can we also return them below?

br4e (commented):
For testing, can you just test using the existing test `TestGetLatestNetworkInferences`?

x/emissions/keeper/inference_synthesis/network_losses_test.go
guilherme-brandao: this is a test for the `CalcCombinedNetworkLoss` which is a function that I deleted

x/emissions/keeper/inference_synthesis/network_losses.go
relyt29: oh for some reason i thought it was in network_losses_test.go

PR #364, opened 2024-06-21
[ORA-1666] Fix get last network inference query

guilherme-brandao:
- GetLastNetworkInference was using EpochLastEnded as a reference to query the network inference, which is not optimal since this property is updated before the worker sends the inferences.
- Getting the last inferences reported and its `blockHeight` -> then getting the topic `EpochLength` -> `blockHeight - EpochLength` should get the losses from the previous time step. If it exists, calculate and return the network inference.

Ticket: https://linear.app/upshot/issue/ORA-1666/debug-getlatestnetworkinference-returning-empty-values-on-edgenet

### DIFF
--- a/x/emissions/keeper/inference_synthesis/network_inferences.go
+++ b/x/emissions/keeper/inference_synthesis/network_inferences.go
@@ -151,3 +151,119 @@ func GetNetworkInferencesAtBlock(
 
 	return networkInferences, forecastImpliedInferencesByWorker, infererWeights, forecasterWeights, nil
 }
+
+func GetLatestNetworkInference(
+	ctx sdk.Context,
+	k keeper.Keeper,
+	topicId TopicId,
+) (
+	*emissions.ValueBundle,
+	map[string]*emissions.Inference,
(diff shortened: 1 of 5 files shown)


x/emissions/keeper/inference_synthesis/network_inferences.go
```diff
@@ -151,3 +151,119 @@ func GetNetworkInferencesAtBlock(
...
+		moduleParams, err := k.GetParams(ctx)
+		if err != nil {
+			return networkInferences, nil, infererWeights, forecasterWeights, err
+		}
+		topic, err := k.GetTopic(ctx, topicId)
+		if err != nil {
+			Logger(ctx).Warn("Error getting topic: %s", err.Error())
+			return networkInferences, nil, infererWeights, forecasterWeights, nil
+		}
+		previousLossNonce := infBlockHeight - topic.EpochLength
```
br4e, line 196: we might want to check to make sure this isn't a negative number

br4e (approved):
this all looks good to me, one comment.

PR #367, opened 2024-06-23
[ORA-1689] Remove duplicated calc in network inference calcs

guilherme-brandao:
Ticket: https://linear.app/upshot/issue/ORA-1689/improvement-remove-duplicated-calculation-during-network-inference


br4e (approved):
Smart! Makes sense to me. However it doesn't look like the tests pass so let's fix that

kpeluso (approved):

PR #368, opened 2024-06-23
ORA1701 Add check for zero weight sum

br4e:
Check for zero weight sum for forecast implied inferences and skip if there is no weight


br4e:
covered in https://github.com/allora-network/allora-chain/pull/369

PR #376, opened 2024-06-25
epoch length must be > 59 seconds

kush-alloralabs:


relyt29:
epoch length is measured in blocks though, and 10 blocks is 50 seconds, also what's the motivation for having to be greater than a minute?

relyt29 (approved):
thanks for the catch!

kpeluso (approved):

PR #378, opened 2024-06-26
Minimize exposed function surface in the Message Server

relyt29:
make functions internal to the package

just code cleanup


br4e (approved):

guilherme-brandao (approved):
necessary!

PR #379, opened 2024-06-26
Exposing api from val

kpeluso:


scripts/l1_node.sh
```diff
@@ -56,6 +56,12 @@ allorad \
     --minimum-gas-prices=0${DENOM} \
     --rpc.laddr=tcp://0.0.0.0:26657 \
     --p2p.seeds=$SEEDS \
-    --log_level "*:error,state:info,server:info,rewards:debug,inference_synthesis:debug,topic_handler:debug" 
-
-
+    --log_level "*:error,state:info,server:info,rewards:debug,inference_synthesis:debug,topic_handler:debug" \
```
okedeji, line 59: There are duplicates here. need cleanup and just pass one flag each. FYI this is mostly what nops use to standup external val

okedeji (changes requested):

scripts/l1_node.sh
```diff
@@ -56,6 +56,9 @@ allorad \
...
+    --grpc.address=0.0.0.0:9090 \
+    --api.address=tcp://0.0.0.0:1317 \
```
relyt29, line 61: I'm fine with this so long as we're doing the right things in terms of exposing these ports via firewall to the outside world


IIRC some of these ports (1317, 9090) aren't meant to be exposed to untrusted environments

kpeluso: @okedeji @vladupshot @RedBird96  for comment

okedeji: this is for external nops though, they will decide to comment it off or not and then they surely will have to handle their own security... our docker-compose is just the basic for them to start

our nodes are running from EKS and we are already finetuning on our WAF to handle that. We had it but it was blocking requests so it's back to in progress

kpeluso: @okedeji ok true. but our ports are still protected because we have our own scripts and infra (excluding WAF per above) protecting these ports already?

okedeji: We have to be open, but they are behind firewall.

We are not exposing these ports in the sense of it, they are being accessed through nodeport and through the loadbalancer. The LB and nodeport are still open though because without them being opened, they can't be accessible.

WAF (Web Access Firewall) is what we use to manage external access through ratelimit, DDOS attacks checks, bots and unwanted automated requests checks etc.. 

Without WAF, we are bare-chest open and can be DDOSed hence why we have WAF.. only issue is that our WAF is currently malfunctioning for now, but that's temporal, as soon as we have more bandwidth, we def will fix it up.

kpeluso:
cherry picking what we actually want to do here:
https://github.com/allora-network/allora-chain/pull/384

PR #380, opened 2024-06-26
rename ftol to tol

kpeluso:
We were inconsistently using `Tolerance` in some places and `FTolerance` in others.


relyt29 (approved):
Largely ok with this, though i do wonder if tolerance is a good name it's kind of vague - tolerance of _what_

kpeluso:
> Largely ok with this, though i do wonder if tolerance is a good name it's kind of vague - tolerance of what

tolerance to number of digits / precision

less vague than "fTolerance"

guilherme-brandao (approved):

PR #388, opened 2024-06-28
Update flag during one-in network inference calc

guilherme-brandao:
- We use different regrets for the inferers in the One-In inference. We should update the flag `InferersNewStatus` to check if all inferers are new, only one is new, or none are new.
- We need to confirm if this is the correct way to handle this scenario. Since we use a key with 2 indexes for the One-In regrets, it depends on the inferer-forecaster combination to determine if there's a prior regret or not.
- The code from this draft is not optimized and repetitive, the main initial goal is checking if it makes sense.

### DIFF
--- a/x/emissions/keeper/inference_synthesis/network_inference_builder.go
+++ b/x/emissions/keeper/inference_synthesis/network_inference_builder.go
@@ -249,7 +249,7 @@ func (b *NetworkInferenceBuilder) calcOneInValue(oneInForecaster Worker) (allora
 	forecastImpliedInferencesWithForecaster[oneInForecaster] = palette.ForecastImpliedInferenceByWorker[oneInForecaster]
 	palette.ForecastImpliedInferenceByWorker = forecastImpliedInferencesWithForecaster
 	palette.Forecasters = []Worker{oneInForecaster}
-	
+
 	// Get one-in regrets for the forecaster and the inferers they provided forecasts for
 	palette.InferersNewStatus = InferersAllNew
 	for _, inferer := range palette.Inferers {
--- a/x/emissions/keeper/inference_synthesis/network_inference_builder.go
+++ b/x/emissions/keeper/inference_synthesis/network_inference_builder.go
@@ -295,7 +295,6 @@ func (b *NetworkInferenceBuilder) SetOneInValues() *NetworkInferenceBuilder {
 		oneInValue, err := b.calcOneInValue(oneInForecaster)
 		if err != nil {
 			b.logger.Warn(fmt.Sprintf("Error calculating one-in inferences: %s", err.Error()))
-			oneInInferences = make([]*emissions.WorkerAttributedValue, 0)
 			return b
 		}
 		oneInInferences = append(oneInInferences, &emissions.WorkerAttributedValue{
--- a/x/emissions/keeper/inference_synthesis/network_inference_builder.go
+++ b/x/emissions/keeper/inference_synthesis/network_inference_builder.go
@@ -249,13 +249,26 @@ func (b *NetworkInferenceBuilder) calcOneInValue(oneInForecaster Worker) (allora
 	forecastImpliedInferencesWithForecaster[oneInForecaster] = palette.ForecastImpliedInferenceByWorker[oneInForecaster]
 	palette.ForecastImpliedInferenceByWorker = forecastImpliedInferencesWithForecaster
 	palette.Forecasters = []Worker{oneInForecaster}
-
+	
 	// Get one-in regrets for the forecaster and the inferers they provided forecasts for
+	palette.InferersNewStatus = InferersAllNew
 	for _, inferer := range palette.Inferers {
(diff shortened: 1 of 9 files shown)


br4e:
oh I see, interesting. Nice catch! Wouldn't we also want to do the same with one out also? 

How did you find this? Did something break? Or were you just thinking through the logic?

guilherme-brandao:
> oh I see, interesting. Nice catch! Wouldn't we also want to do the same with one out also?
> 
> How did you find this? Did something break? Or were you just thinking through the logic?

Since we are considering different regrets for the one-in, I think it's necessary for this. It's an issue where we get zero one-in network inferences for an inferer-forecaster (key) that doesn't have a prior regret.

br4e:
yeah I agree that we should do it, I was just wondering what made you think of it. 

What about one out regrets? the same logic applies right?

kpeluso:
We're still internally debating if this is actually a valid change or not. Will keep this thread updated.

guilherme-brandao:
> yeah I agree that we should do it, I was just wondering what made you think of it.
> 
> What about one out regrets? the same logic applies right?

@br4e you're right, we should also update flags (single not new, all new, etc) during one-out calcs

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -250,31 +251,37 @@ func (b *NetworkInferenceBuilder) calcOneInValue(oneInForecaster Worker) (allora
...
 	palette.Forecasters = []Worker{oneInForecaster}
 
+	regret, noPriorRegret, err := palette.K.GetOneInForecasterSelfNetworkRegret(palette.Ctx, palette.TopicId, oneInForecaster)
+	if err != nil {
+		return alloraMath.Dec{}, errorsmod.Wrapf(err, "Error getting one-in forecaster regret")
+	}
+
+	if noPriorRegret {
+		return alloraMath.NewNaN(), nil
+	}
```
br4e, line 261: remind me why we want this?

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -250,31 +251,37 @@ func (b *NetworkInferenceBuilder) calcOneInValue(oneInForecaster Worker) (allora
...
+
+	palette.ForecasterRegrets[oneInForecaster] = &StatefulRegret{
+		regret:        regret.Value,
+		noPriorRegret: noPriorRegret,
```
br4e, line 265: so this will always be false? Should we mark it as such by setting it to false?

x/emissions/keeper/inference_synthesis/network_inference_builder_test.go
```diff
@@ -435,9 +435,10 @@ func (s *InferenceSynthesisTestSuite) testCorrectOneInForecasterValuesForEpoch(e
 	}
 }
 
-func (s *InferenceSynthesisTestSuite) TestCorrectOneInForecasterValuesEpoch0() {
-	s.testCorrectOneInForecasterValuesForEpoch(0)
-}
+// TODO: Check if we can consider this test as valid
+// func (s *InferenceSynthesisTestSuite) TestCorrectOneInForecasterValuesEpoch0() {
+// 	s.testCorrectOneInForecasterValuesForEpoch(0)
+// }
```
br4e, line 441: this is fine to delete

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
```diff
@@ -201,14 +201,24 @@ func (p *SynthPalette) CalcWeightedInference(weights RegretInformedWeights) (Inf
...
+		if !regretInfo.noPriorRegret {
+			if p.InferersNewStatus == InferersAllNew {
+				p.InferersNewStatus = InferersAllNewExceptOne
+				p.SingleNotNewInferer = address
+			} else {
+				p.InferersNewStatus = InferersNotNew
+				p.SingleNotNewInferer = ""
+			}
+		}
+		regrets = append(regrets, regretInfo.regret)
```
br4e, line 221: could maybe refactor this into a helper function somewhere since I think it is used in multiple places in the codebase, but I'm flexible

x/emissions/keeper/inference_synthesis/network_inference_builder.go
guilherme-brandao: yeah, why you think it's better to set as false. I think it is understood that it always has to be false in this case.

br4e: just more clarity, doesn't really matter though

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
```diff
@@ -225,8 +225,68 @@ func (p *SynthPalette) GetForecasterRegretsSlice() []alloraMath.Dec {
 	return regrets
 }
 
+func (p *SynthPalette) UpdateInferersInfo(newInferers []Worker) error {
```
br4e, line 228: wooo

PR #393, opened 2024-07-02
 Check Invariants Hold During Integration Testing

relyt29:
Check the following invariants during the course of an integration test run:

* The balance of the totalStake in the emissions keeper should always equal the cosmos-sdk /x/bank balance of the "allorastaking" bank account
* The sum of all topicStake in every topic should always equal the totalStake
* The sum of all stakeReputerAuthority for a given topic should always equal the topic stake for that topic
* The sum of all stakeReputerAuthority should equal the total stake
* The stakeSumFromDelegator should always equal the sum of `delegatedStakes[topicid, delegator, all reputers for that delegator])`
* The stakeFromDelegatorsUponReputer should always equal the sum of `delegatedStakes[topicid, all delegators for that reputer, reputer]`
* The sum of all (delegateRewardsPerShare * delegated stake - reward debt) = the balance of the /x/bank AlloraPendingRewardForDelegatorAccountName module account
* The state of the stakeRemovalsByBlock and stakeRemovalsByActor should always be identical in terms of the number of keys in the map, and the content of those keys
* The sum of all stakeRemovals should always be less than or equal to the totalStake

The invariant checking lives in a file called invariants.go. The cosmos /x/invariants module apparently is not functional at this time so rather than rely on their module, we just edit the ABCI endBlock to check the invariants at the end of every block. This is not perfect, but better than nothing. The code to do that is stored in a patch that is applied in the github workflow via `git apply ./test/invariant/invariants.patch`

### DIFF
--- a/.github/workflows/format_and_test.yml
+++ b/.github/workflows/format_and_test.yml
@@ -51,6 +51,9 @@ jobs:
       with:
         go-version: 1.22.x
 
+    - name: Apply Invariants Patch
+      run: git apply ./test/invariant/invariants.patch
+
     - name: Run allora l1 chain
       run: bash ./test/local_testnet_l1.sh
 
--- a/test/invariant/invariants.patch
+++ b/test/invariant/invariants.patch
@@ -1,0 +1,43 @@
+diff --git a/x/emissions/module/abci.go b/x/emissions/module/abci.go
+index 1ca10a02..7a755bac 100644
+--- a/x/emissions/module/abci.go
++++ b/x/emissions/module/abci.go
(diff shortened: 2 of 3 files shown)


relyt29:
lmao this breaks the unit tests because they abuse the invariants to do testing I'm pretty sure

I'm going to split up the github action unit test again to make it parallelized again

relyt29:
gonna rebase this thing and force push in a minute here

guilherme-brandao (approved):
Checks look good! Wondering if you think we should add more staking tests to take advantage of this, or if we are okay for now?

PR #394, opened 2024-07-02
Parallelize Unit and Integration Testing Jobs

relyt29:
In #290 I made all the integration and unit tests run on the same github VM in some weird goal of either making our github actions cheaper and or faster

I am now of the conclusion that it is definitely not faster

therefore, we use job dependencies to allow parallelizing the github action on multiple VMs. I used [this](https://medium.com/tradeling/how-to-achieve-parallel-execution-using-github-actions-d534404702fb) as a reference.

### DIFF
--- a/.github/workflows/format_and_test.yml
+++ b/.github/workflows/format_and_test.yml
@@ -3,29 +3,24 @@ name: Run Format, Unit Tests, and Integration Tests
 on: [pull_request]
 
 jobs:
-  build-and-format:
+  gofmt: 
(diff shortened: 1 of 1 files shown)


relyt29:
2024-07-02-111156_1083x513_scrot
we first build and format, and if that passes then we will run the tests and integration tests

relyt29:
2024-07-02-112643_1114x586_scrot
this is what it looks like visualized in terms of running in different containers

okedeji (approved):

PR #395, opened 2024-07-02
ORA1746 Make changes to fTolerance

br4e:
Goals: 

- remove the median from eqs. 8, 11, 28
- use new global epsilon_reputer = 0.01 for eq. 32
- remove bound requirement (I think this was <=0.01) on epsilon used in eqs. 8, 11, 28.
- remove the new machine-precision global f_tolerance variable that was introduced last week


x/emissions/keeper/inference_synthesis/synth_palette_forecast_implied_test.go
```diff
@@ -405,7 +405,7 @@ func (s *InferenceSynthesisTestSuite) TestCalcForcastImpliedInferencesEpoch3() {
 			alloraMath.InDelta(
 				expectedValue.Value,
 				actualValue.Value,
-				alloraMath.MustNewDecFromString("0.001"),
+				alloraMath.MustNewDecFromString("0.01"),
```
guilherme-brandao, line 408: prob we should add the updated version of the simulator outputs instead of lowering precision here, if it's available ofc

guilherme-brandao (approved):
Lgtm!

x/emissions/keeper/inference_synthesis/synth_palette_forecast_implied_test.go
br4e: the updated simulator outputs are included

PR #398, opened 2024-07-03
Added missing validations to query server

guilherme-brandao:
Ticket: https://linear.app/upshot/issue/ORA-1719/[swarm-tests]-consumer-side


br4e (approved):
seems straightforward enough!

PR #399, opened 2024-07-04
ORA1761 New “strict” query that returns latest inferences and losses when both are fully available

br4e:
This includes both inferences and losses


x/emissions/proto/emissions/v1/query.proto
```diff
@@ -170,8 +170,13 @@ service Query {
...
   }
 
-  rpc GetLatestNetworkInference(QueryLatestNetworkInferencesAtBlockRequest)
-      returns (QueryLatestNetworkInferencesAtBlockResponse) {
+  rpc GetLatestNetworkInference(QueryLatestNetworkInferencesRequest) returns (QueryLatestNetworkInferencesResponse) {
+    option (cosmos.query.v1.module_query_safe) = true;
+    option (google.api.http).get = "/emissions/v1/network_inferences/{topic_id}";
+  }
+
+  rpc GetLatestNetworkInferenceWithLosses(QueryLatestNetworkInferencesRequest)
```
guilherme-brandao, line 178: I'd name this something like`GetLatestAvailableNetworkInference`

x/emissions/keeper/queryserver/query_server_inferences.go
```diff
@@ -80,10 +83,76 @@ func (qs queryServer) GetLatestNetworkInference(
...
+	if err != nil {
+		return nil, err
+	}
+
+	previousLossBlockHeight := inferenceBlockHeight - topic.EpochLength
+
+	for true {
+		if previousLossBlockHeight < 0 {
+			return nil, status.Errorf(codes.NotFound, "losses not yet available for topic %v", req.TopicId)
+		}
```
guilherme-brandao, line 118: We can have this case and also a case where losses are available but there isn't a subsequent inference. Since this is a 'strict' query, I think we should also cover this case.

guilherme-brandao (changes requested):

x/emissions/keeper/queryserver/query_server_inferences.go
br4e: Hmm this should be handled with the line `if len(inferences.Inferences) > 0 {` 

right?

br4e:
Also are we worried about a possible infinite loop? Should we limit the iterations to 10 epochs back or something?

x/emissions/keeper/queryserver/query_server_inferences.go
br4e: ah with some more tests I see the edge case. Working on a fix

br4e: fixed!

x/emissions/keeper/queryserver/query_server_inferences.go
```diff
@@ -97,10 +100,81 @@ func (qs queryServer) GetLatestNetworkInference(
...
+) (
+	*types.QueryLatestNetworkInferencesResponse,
+	error,
+) {
+
+	// Find the latest inference block height
+	_, inferenceBlockHeight, err := qs.k.GetLatestTopicInferences(ctx, req.TopicId)
+	if err != nil {
+		return nil, err
+	}
```
guilherme-brandao, line 125: You can also add a check here if the blockHeight is equal to zero like in the other queries

br4e: updated

br4e:
I think default page limit makes sense, it's kind of like looking through that many "pages" of inferences before giving up so we don't have something like an infinite loop.

x/emissions/keeper/queryserver/query_server_inferences.go
```diff
@@ -97,10 +100,84 @@ func (qs queryServer) GetLatestNetworkInference(
...
+			// check for losses. If they exist, then break so we can query
+			_, err = qs.k.GetNetworkLossBundleAtBlock(ctx, req.TopicId, previousLossBlockHeight)
+			if err == nil {
+				break
+			}
+		}
+
+		inferenceBlockHeight -= topic.EpochLength
+		previousLossBlockHeight -= topic.EpochLength
+	}
```
kpeluso, line 160: 1. previous block height won't strictly be at that time e.g. if the topic was inactive for a period
2. this is a very inefficient way to discover the right block. we should find an appropriate block and then use functions (GetInferencesAtBlock, GetNetworkLossBundleAtBlock) for their intended purpose given the appropriate blocks

kpeluso (changes requested):

x/emissions/keeper/queryserver/query_server_inferences.go
br4e: Do you have an alternative method in mind for the second point?

br4e:
Ok, this is updated and ready for re-review. I recorded the relevant information so it can be used to determine the latest blocks for reputer and works payloads

kpeluso (approved):

PR #402, opened 2024-07-04
ORA1762 Add Error bar/confidence interval calculation to query returns

br4e:


x/emissions/keeper/queryserver/query_server_inferences.go
```diff
@@ -164,17 +185,131 @@ func (qs queryServer) GetLatestAvailableNetworkInference(
...
+	}
+
+	forecastersToWeights := make(map[string]float64)
+	for worker, weight := range forecasterWeights {
+		floatValue, err := stringToFloat64(weight.String())
+		if err != nil {
+			return nil, nil, err
+		}
+		forecastersToWeights[worker] = floatValue
+	}
```
guilherme-brandao, line 261: Why we are using `float64` here instead of `Dec`?

x/emissions/keeper/inference_synthesis/network_inferences.go
```diff
@@ -247,3 +248,88 @@ func GetLatestNetworkInference(
...
+	for i, p := range percentiles {
+		quantiles[i] = p / 100
+	}
+	result, err := linearInterpolation(quantiles, normalizedWeights, sortedData)
+	if err != nil {
+		return nil, err
+	}
+
+	return result, nil
+}
```
guilherme-brandao, line 335: I think this can be moved to `math/utils`

guilherme-brandao (changes requested):

x/emissions/keeper/inference_synthesis/network_inferences.go
relyt29: agree

x/emissions/keeper/queryserver/query_server_inferences.go
relyt29: agreed we should not have floats anywhere in the codebase, they break consensus because CPU architectures implement finer details of float IEEE standard differently

x/emissions/keeper/inference_synthesis/network_inferences.go
relyt29: and none of these functions should be using float as discussed elsewhere

kpeluso: ^^all the above

x/emissions/keeper/queryserver/query_server_inferences.go
```diff
@@ -167,17 +187,103 @@ func (qs queryServer) GetLatestAvailableNetworkInference(
...
+
+	inferersToInference := make(map[string]alloraMath.Dec)
+	for _, infererValue := range networkInferences.InfererValues {
+		inferersToInference[infererValue.Worker] = infererValue.Value
+	}
+
+	forecastersToInference := make(map[string]alloraMath.Dec)
+	for _, forecasterValue := range networkInferences.ForecasterValues {
+		forecastersToInference[forecasterValue.Worker] = forecasterValue.Value
+	}
```
kpeluso, line 237: I don't understand why these maps exist at all... If anything they make the order of elements in the query nondeterministic, which possibly has effects on downstream applications

x/emissions/keeper/queryserver/query_server_inferences.go
```diff
@@ -167,17 +187,103 @@ func (qs queryServer) GetLatestAvailableNetworkInference(
...
+) ([]alloraMath.Dec, []alloraMath.Dec, error) {
+
+	inferersToInference := make(map[string]alloraMath.Dec)
+	for _, infererValue := range networkInferences.InfererValues {
+		inferersToInference[infererValue.Worker] = infererValue.Value
+	}
+
+	forecastersToInference := make(map[string]alloraMath.Dec)
+	for _, forecasterValue := range networkInferences.ForecasterValues {
+		forecastersToInference[forecasterValue.Worker] = forecasterValue.Value
```
kpeluso, line 236: should `forecasterValue.Value` be `forecastImpliedInferenceByWorker` instead??

kpeluso (changes requested):

x/emissions/keeper/queryserver/query_server_inferences.go
kpeluso: proposed in my latest commit

x/emissions/keeper/queryserver/query_server_inferences.go
kpeluso: proposed in my latest commit

br4e:
This makes sense!

kpeluso (approved):

PR #404, opened 2024-07-05
Functions create their own context.Ctx to call cosmosclient.Client

relyt29:
Remove global shared context from TestConfig struct

Every place in the test suite where `m.Ctx` was called, instead just call context.Background and create a new separate context object.

Not actually sure this works, just a test PR to test the idea

### DIFF
--- a/test/common/node_config.go
+++ b/test/common/node_config.go
@@ -1,7 +1,6 @@
 package testcommon
 
 import (
-	"context"
 	"testing"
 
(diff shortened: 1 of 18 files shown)


relyt29:
I'm going to flip this to ready for review, it seems to work as far as I can tell. See https://linear.app/upshot/issue/ORA-1765/improve-stress-tests-properly-use-goroutines for motivation

kpeluso (approved):

PR #406, opened 2024-07-05
Fix pagination

kpeluso:
Fixes for pagination so they pass the newly added test cases.

The preexisting pagination method would fail to produce the desired page of topic ids when the set of active topics includes a non-consecutive set of topic ids.
e.g. topic ids={1,2,3} but 2 is not active => `GetIdsOfActiveTopics({limit:2})` would return `{1}` when it should return `{1,3}`, for any passed `limit≥2`.

Test cases for this and other behaviors added as well.


x/emissions/keeper/keeper.go
```diff
@@ -1607,28 +1607,38 @@ func (k Keeper) GetIdsOfActiveTopics(ctx context.Context, pagination *types.Simp
...
-
-	rng, err := k.activeTopics.IterateRaw(ctx, startKey, nextKey, collections.OrderAscending)
+	iter, err := k.activeTopics.Iterate(ctx, rng)
 	if err != nil {
 		return nil, nil, err
 	}
-	activeTopics, err := rng.Keys()
-	if err != nil {
-		return nil, nil, err
+	defer iter.Close()
```
br4e, line 1616: :)

x/emissions/keeper/keeper.go
```diff
@@ -1607,28 +1607,38 @@ func (k Keeper) GetIdsOfActiveTopics(ctx context.Context, pagination *types.Simp
 	if err != nil {
 		return nil, nil, err
 	}
+	rng := new(collections.Range[uint64]).StartExclusive(start)
```
br4e, line 1610: would we maybe want inclusive instead of exclusive?

x/emissions/keeper/keeper.go
```diff
@@ -1607,28 +1607,38 @@ func (k Keeper) GetIdsOfActiveTopics(ctx context.Context, pagination *types.Simp
...
+	activeTopicIds := make([]TopicId, 0)
+	nextKey := make([]byte, binary.MaxVarintLen64)
+	nextTopicId := uint64(0)
+	for ; iter.Valid(); iter.Next() {
+		topicId, err := iter.Key()
+		if err != nil {
+			return nil, nil, err
+		}
+
+		if uint64(len(activeTopicIds)) >= limit {
```
br4e, line 1627: Hmm, is this how we want pagination to work? Shouldn't it iterate a limited number of times, regardless of what data is found? 

Otherwise someone could iterate much longer than the pagination limit if there are no active topics

This is an unlikely scenario, but it seems to possibly defeat the purpose of pagination

br4e (approved):

x/emissions/keeper/keeper.go
kpeluso: In theory this has the same effect as keeping a counter that we increment at each iteration, and breaking once that reaches a threshold. Any way that logic is no good?

x/emissions/keeper/keeper.go
kpeluso: `nextTopicId` is set after the value is added to the array, so if we make "inclusive" then we'd return the same value twice across 2 consecutive calls

x/emissions/keeper/keeper.go
br4e: well as you've written it the counter increments only if an activeTopicId is found, whereas in this comment it increments whether an active topic or inactive topic is found

br4e: I don't have a strong opinion about this given that an attack is probably not viable

RedBird96 (approved):

x/emissions/keeper/keeper.go
br4e: ooopsss I misunderstood the code. We are only incrementing over active topics. Cool, this is correct.

PR #408, opened 2024-07-08
Hotfix/fix range and catch map lookups

kpeluso:
Fixes 2 bugs:
1. Inference synthesis was using non-deterministic ranges => would eventually cause inconsistent state across nodes for any data saved from inference synthesis
2. Catching map lookups for keys that don't exist.


br4e (approved):
This looks right to me! I didn't review the test cases but the code looks good

x/emissions/keeper/inference_synthesis/synth_palette_bootstrap.go
```diff
@@ -55,29 +55,49 @@ func (p *SynthPalette) BootstrapRegretData() error {
...
-	for k, v := range p.ForecasterRegrets {
-		regretCopy := *v
-		forecasterRegrets[k] = &regretCopy
+	for _, worker := range p.Forecasters {
+		data, ok := p.ForecasterRegrets[worker]
+		if !ok {
+			continue
+		}
+		regretCopy := *data
+		forecasterRegrets[worker] = &regretCopy
```
guilherme-brandao, line 100: I think we will keep having the same results here, it will not make any difference with this change...

x/emissions/keeper/inference_synthesis/common.go
```diff
@@ -34,20 +34,24 @@ func MakeMapFromForecasterToTheirForecast(forecasts []*emissions.Forecast) map[W
...
 	forecastImpliedInferenceByWorker map[string]*types.Inference,
 ) []*types.WorkerAttributedValue {
 	forecastImpliedInferences := make([]*types.WorkerAttributedValue, 0)
-	for worker, inference := range forecastImpliedInferenceByWorker {
-		forecastImpliedInferences = append(forecastImpliedInferences, &types.WorkerAttributedValue{Worker: worker, Value: inference.Value})
+	for _, worker := range workers {
+		forecastImpliedInferences = append(
+			forecastImpliedInferences,
+			&types.WorkerAttributedValue{Worker: worker, Value: forecastImpliedInferenceByWorker[worker].Value},
+		)
```
guilherme-brandao, line 54: need to catch map lookup here?

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
```diff
@@ -144,6 +160,15 @@ func (p *SynthPalette) CalcWeightedInference(weights RegretInformedWeights) (Inf
...
 		for _, inferer := range p.Inferers {
+			if _, ok := p.InferenceByWorker[inferer]; !ok {
+				continue
+			}
+			if _, ok := weights.inferers[inferer]; !ok {
+				continue
+			}
+			if _, ok := p.InfererRegrets[inferer]; !ok {
+				continue
+			}
```
guilherme-brandao, line 171: If we want to move forward with these checks, it might be a good idea to add some alerts here as well. Considering the actual implementation, If we ever fall into these check, we may have some critical issue in these calculations.

guilherme-brandao (commented):
I think this PR is adding some redundancy. The change in Regrets iteration maybe makes sense.

x/emissions/keeper/inference_synthesis/common.go
kpeluso: just adding it to part of the "function's contract/declaration" in comments

x/emissions/keeper/inference_synthesis/synth_palette_bootstrap.go
kpeluso: adding debug logging just in case

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
kpeluso: adding debug logging just in case as suggested

PR #416, opened 2024-07-10
Diego/ora 1785 worker nonces branch

xmariachi:
Closes: #ORA-1785

Reworks the nonces system as per ticket ORA-1875.
This contains modification of nonce types, which means breaking changes for workers and reputers. 

## Testing and Verifying

This change modifies affected tests, and removes others that no longer apply.
Manual test via docker of the changes to ensure changes preserve functionality and do not add noise/unexpected issues.

  - [ X ] Allora documentation site `docs.allora.network` source code at: `https://docs.allora.network/learn/architecture#nonces`
  - [ ] Code comments?
  - [ ] N/A


x/emissions/module/abci.go
```diff
@@ -65,7 +65,7 @@ func EndBlocker(ctx context.Context, am AppModule) error {
 					sdkCtx.Logger().Warn(fmt.Sprintf("Error updating last inference ran: %s", err.Error()))
 				}
 				// Add Worker Nonces
-				nextNonce := types.Nonce{BlockHeight: blockHeight + topic.EpochLength}
+				nextNonce := types.Nonce{BlockHeight: blockHeight}
```
RedBird96, line 68: Shouldn't we keep this as `blockHeight + topic.EpochLength`? Because this is called at end of block so means this nonce(block) was already passed without fullfilled. Wdyt?

kpeluso: per [the ticket](https://uploads.linear.app/9b3a8ec3-68c2-4405-b1f3-51f697acd644/95a520f8-f890-417a-a448-ea3ec4bc04e5/ec7f533c-be30-48da-bea2-f492bf2ad739) the nonce refers to when the request was generated, and is the id for the entire sequence of events that cumulates in a network inference

x/emissions/keeper/inference_synthesis/nonce_mgmt.go
```diff
@@ -37,7 +37,8 @@ func SelectTopNReputerNonces(reputerRequestNonces *emissions.ReputerRequestNonce
 	// loop reputerRequestNonces
 	for _, nonce := range sortedSlice {
 		nonceCopy := nonce
-		if currentBlockHeight >= nonceCopy.ReputerNonce.BlockHeight+groundTruthLag {
+		// Only select when the ground truth is available.
+		if currentBlockHeight >= nonceCopy.ReputerNonce.BlockHeight+groundTruthLag+epochLength {
```
kpeluso, line 41: what's the reasoning for the extra  epochLength? the ground truth lag should be enough, no?

app/topics_handler.go
```diff
@@ -95,18 +95,30 @@ func (th *TopicsHandler) requestTopicReputers(ctx sdk.Context, topic emissionsty
...
+	lastCommit, err := th.emissionsKeeper.GetTopicLastCommit(ctx, topic.Id, emissionstypes.ActorType_REPUTER)
+	if err != nil {
+		Logger(ctx).Warn("Error getting reputer last commit: "+err.Error(), ", first reputer commit?")
+	}
 	// iterate over all the reputer nonces to find if this is unfulfilled
 	for _, nonce := range topNReputerNonces {
 		nonceCopy := nonce
-		Logger(ctx).Warn(fmt.Sprintf("Reputer block height found unfulfilled, requesting reputers for block: %v, worker: %v", nonceCopy.ReputerNonce.BlockHeight, nonceCopy.WorkerNonce.BlockHeight))
+		// Get previous losses from keeper, default: reputer - epochLength
+		lastReputerCommitBlockHeight := nonceCopy.ReputerNonce.BlockHeight - topic.EpochLength
```
kpeluso, line 109: I don't get why, if `lastCommit.Nonce.BlockHeight` does not exist (as checked below), why anything should exist at this value (nonce minus epoch length)...

kpeluso (commented):
just 2 things I don't get yet...

x/emissions/keeper/inference_synthesis/nonce_mgmt.go
xmariachi: gt_lag is added extra to epochLength. Reasons are historical (before we had gt_lag, epochLength was the measure), but this can now be simplified, so only gt_lag is considered for reputation (gt_lag since nonce, regardless of epochLength); also adding a requirement that gt_lag is >= epochLength. 

This changes the current meaning so affects topic definitions, but looks cleaner.
Will apply.,

x/emissions/module/abci.go
xmariachi: What @kpeluso said. 
Also, logs show that the request for nonce X is made at block X.

app/topics_handler.go
xmariachi: A tentative value is given, which is the most common value - but not having them should not break the flow, request still needs to be made even if there are no losses at that block at the time of check.
An alternative would be sending a value that marks this explicitly, like 0. Although sending the most common value allows for a last-minute regular reputation being inserted in the in-between time, especially for very short epochLengths.

xmariachi: Actually it should be 
`nonceCopy.ReputerNonce.BlockHeight + topic.GroundTruthLag - topic.EpochLength`

PR #418, opened 2024-07-11
rm worker and reputer request retries

kpeluso:
Closes: #XXX

Removes retries for worker and reputer nonces. This is to simplify and lessen operations.

## Testing and Verifying

This change is already covered by existing unit and integration tests.

  - [x] Does this pull request introduce a new feature or user-facing behavior changes? NO

  - [] Specification (`x/{module}/README.md`)
  - [x] Allora documentation site `docs.allora.network` source code at: `https://github.com/allora-network/docs`
  - [ ] Code comments?
  - [ ] N/A


app/topics_handler.go
```diff
@@ -61,87 +61,84 @@ func (th *TopicsHandler) requestTopicWorkers(ctx sdk.Context, topic emissionstyp
...
-	// iterate over all the reputer nonces to find if this is unfulfilled
-	for _, nonce := range topNReputerNonces {
-		nonceCopy := nonce
-		// Get previous losses from keeper, default: reputer - epochLength
-		lastReputerCommitBlockHeight := nonceCopy.ReputerNonce.BlockHeight - topic.EpochLength
-		if lastCommit.Nonce != nil {
-			Logger(ctx).Debug(fmt.Sprintf("Reputer last commit found, setting: %v", lastCommit))
-			lastReputerCommitBlockHeight = lastCommit.Nonce.BlockHeight
-		}
+	nonceCopy := reputerNonces.Nonces[0]
```
xmariachi, line 92: I think this one is trickier: because reputer nonces legitimately accumulate during ground_truth_lag, before they get to be processed. 
But they can also accumulate because of reputation rounds missed (just like inference rounds missed above).

So this code would also get stuck with the earliest reputation attempt (if new nonces are appended at end), 
So we should get those eligible first (ie those where gt is revealed already), and get the last one of those.

... or do like we were doing: do a number of them and retry, giving priority to new ones, and retrying those unfulfilled ones until they get pruned.

xmariachi (changes requested):

app/topics_handler.go
kpeluso: we could send out up to a batch per topic each time, then we're guaranteed to eventually catch up so long as that batch size >1

app/topics_handler.go
```diff
@@ -61,87 +61,84 @@ func (th *TopicsHandler) requestTopicWorkers(ctx sdk.Context, topic emissionstyp
...
-	// iterate over all the worker nonces to find if this is unfulfilled
-	for _, nonce := range sortedWorkerNonces {
-		nonceCopy := nonce
-		Logger(ctx).Debug(fmt.Sprintf("Current Worker block height has been found unfulfilled, requesting inferences %v", nonceCopy))
-		go generateInferencesRequest(ctx, topic.InferenceLogic, topic.InferenceMethod, topic.DefaultArg, topic.Id, topic.AllowNegative, *nonceCopy)
+	if len(workerNonces.Nonces) < 1 {
+		Logger(ctx).Warn("No unfulfilled worker nonces found for topic %d", topic.Id)
+		return
 	}
+	nonceCopy := workerNonces.Nonces[0]
```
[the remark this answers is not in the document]
kpeluso: because we're always adding new nonces to the head of the list, then this should be the latest, no? To be safe / cleaner design, we could abstract this into class methods on the `Nonces` type

app/topics_handler.go
kpeluso: maybe we also need a new keyset/map that tracks if the unfulfilled nonce was attempted yet

app/topics_handler.go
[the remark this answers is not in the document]
xmariachi: True, missed that line 
`nonces.Nonces = append([]*types.Nonce{nonce}, nonces.Nonces...)`
so it should work then.

So now the allowed length of this array is governed by the amount of epochLengths while a tx from workerLeader will be accepted. I think this can be lowered from the current value of `MaxUnfulfilledWorkerRequests=100`, to something like 2/3.

xmariachi:
This branch will not be merged for now.

kpeluso:
We'll keep retries as they're used by reputers [for this reason](https://github.com/allora-network/allora-chain/pull/418#discussion_r1674170762).

PR #437, opened 2024-07-16
Fix decay rate issue - multiple dripping calls

xmariachi:
Closes: #ORA-1841 

Fix iteration causing decay rate being applied multiple times

## Testing and Verifying

This change is a small change that reuses same unit tests.

  - [ ] Code comments?
  - [X ] N/A - It is a bugfix.

### DIFF
--- a/x/emissions/keeper/keeper.go
+++ b/x/emissions/keeper/keeper.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/binary"
 	"errors"
+	"fmt"
 
 	errorsmod "cosmossdk.io/errors"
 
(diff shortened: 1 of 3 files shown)


x/emissions/keeper/keeper.go
```diff
@@ -1746,6 +1747,8 @@ func (k *Keeper) DripTopicFeeRevenue(ctx context.Context, topicId TopicId, block
 		}
 		newTopicFeeRevenue = val.SdkIntTrim()
 	}
+
+	ctx.Logger().Debug(fmt.Sprintf("Dripping topic fee revenue: block %d, topicId %d, oldRevenue %v, newRevenue %v", ctx.BlockHeight(), topicId, topicFeeRevenue, newTopicFeeRevenue))
```
relyt29, line 1751: I am not aware of anywhere else where we are doing logger().Debug()

I'm not against doing it, seeing as we can always filter logs, but we should probably agree as a team that it is ok to leave debug logging statements in the codebase because I think up until now we had not

xmariachi: Good point.
My view: I don't see why we shouldn't follow logging best practices lol.
Especially when other modules are extremely verbose at DEBUG level.

relyt29 (approved):

kpeluso (approved):

PR #440, opened 2024-07-19
Remove `go func` parallelization in ABCI EndBlocker

relyt29:

This change simplifies a hard-to-read and hard-to-reason about piece of code that parallelized the picking of churn ready topics from the active set. After this PR the code is easier to read and understand, and has less likelihood of causing non-determinism issues. The tradeoff is that the code may run slightly slower in serial.

## Testing and Verifying

This change is not trivial, however it is a small diff / refactor PR and our existing test cases should be able to cover the change.

This PR does not have any implications for documentation.

### DIFF
--- a/test/invariant/invariants.patch
+++ b/test/invariant/invariants.patch
@@ -1,16 +1,17 @@
 diff --git a/x/emissions/module/abci.go b/x/emissions/module/abci.go
-index 1ca10a02..7a755bac 100644
+index 8d778144..523ac83f 100644
 --- a/x/emissions/module/abci.go
 +++ b/x/emissions/module/abci.go
-@@ -6,6 +6,7 @@ import (
- 	"sync"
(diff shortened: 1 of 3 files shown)


guilherme-brandao:
I would like to know more about the the potential non-determinism here.

In this specific case, the parallelization was added because we are dealing with data structures of different topics, so there's no risk of race conditions afaik.

relyt29:
non determinism because the order of operations is now non-deterministic as the different threads will interrupt and be ordered differently in the scheduler on different computers

according to my understanding the order of writes into the IAVL tree on disk does produce different trees for different orders of insertions, even if the data is the same

we have not seen evidence of this being a real issue on mainnet, but I advocate for removing the (extremely convoluted, hard to read) parallelization code anyways

guilherme-brandao (approved):

PR #442, opened 2024-07-19
Prevent potential race condition by making parallelized pass-by-copy rather than reference

relyt29:

This PR closes a subtle race condition bug that could potentially happen anywhere that for-loops are spawned with `go func` - the for loop operates on a reference which changes faster than the function call can potentially happen, causing a function to use the wrong parameter when referenced.

The solution is very simple, make a local copy of the variable and pass the resulting function by copy, rather than by reference - then there's no way the for loop can accidentally reference the wrong data. 

## Testing and Verifying

This change is hard to test, so we're not going to write a unit test to simulate it explicitly. The performance implicaitons of doing a pass-by-copy are trivial, therefore there is no reason to not do the safer thing. Our existing test suite should test that the change does not have correctness implications.

This PR has no implications for documentation.


guilherme-brandao (approved):
nice

PR #444, opened 2024-07-19
Remove Parallelization in topic handler over topics

relyt29:

this change removes parallelization in the topics handler over topics. Instead of parallelizing the math to craft the request for each topic, this change only allows the actual submission of HTTP requests to be parallelized. 

The reason for this is that it makes the code simpler to reason about especially with regards to race conditions across many threads accessing the same keeper

## Testing and Verifying

This code is covered by our existing unit tests. No functionality is changed.

This PR has no implications for our documentation

### DIFF
--- a/app/topics_handler.go
+++ b/app/topics_handler.go
@@ -2,7 +2,6 @@ package app
 
 import (
 	"fmt"
-	"sync"
 
 	"cosmossdk.io/log"
 	emissionstypes "github.com/allora-network/allora-chain/x/emissions/types"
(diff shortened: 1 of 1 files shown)


guilherme-brandao:
It seems like we could maintain the parallelization just by adding Mutex on specific keeper functions

relyt29:
I'm actually going to close this in favor of #458  - when 458 is merged, this issue becomes moot because the block proposer no longer has to fire off these kinds of requests to the off-chain components

PR #451, opened 2024-07-22
Collections.go Iterators need to be manually closed

relyt29:

The collections.go cosmos-sdk library provides a manual iteration function for iterating over all the keys of a collection. However the iterator must be manually closed by the caller. 

I suspect this probably would cause memory leaks if we didn't fix this issue, but deferring an iter.Close() call is trivial so we fix it here

## Testing and Verifying

In terms of the existing code functionality, this change is covered by our existing unit and integration tests.

This PR has no implications for our documentation


guilherme-brandao (approved):
lgtm! just need to update the branch

PR #460, opened 2024-07-24
High Precision Tests and Research Updates

guilherme-brandao:

**Fixes, improvements and updates**
-  Added new regrets (research update)
- Added  new type of loss and inference (research update)
- Updated forecast utility function (research update)
- Refactor to remove duplicate calculation of chi and gamma
- Fixed weighted inference calculation: We were using the same epsilon value that is set in the topic to avoid division by zero, which can be a high value considering the weights calculation. Therefore, I added the value of 1e-7.
- Fixed forecast entropy calculation: We were using the normal fractions instead of the post-EMA/smoothed fractions.

**High precision tests:**
**Section 3**
- Regrets
- Network Loss
- Network Inference

**Section 4**
- Scores
- Fractions
- Entropies
- Total Rewards per Task

## Testing and Verifying

*(Please pick one of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added unit test that validates ...*
  - *Added integration tests for end-to-end deployment with ...*
  - *Extended integration test for ...*
  - *Manually verified the change by ...*

  - [ ] Code comments?
  - [ ] N/A


test/testutil/genesis.go
```diff
@@ -0,0 +1,300 @@
+package testutil
+
```
relyt29, line 2: this file can be deleted, and will be in my PR so up to you if you want to remove it now or later

my apologies for merging it in, it was appendix code from something I had been working on

x/emissions/keeper/genesis.go
```diff
@@ -558,16 +558,6 @@ func (k *Keeper) InitGenesis(ctx context.Context, data *types.GenesisState) erro
 			}
 		}
 	}
-	//LatestOneInForecasterSelfNetworkRegrets []*TopicIdActorIdTimeStampedValue
```
relyt29, line 561: so we're deleting these lines from the genesis export, ok

I saw that there were some new additions to reputer.proto - shouldn't we be adding those in, here?

x/emissions/keeper/inference_synthesis/synth_palette_forecast_implied_test.go
```diff
@@ -110,7 +110,7 @@ func (s *InferenceSynthesisTestSuite) TestCalcForecastImpliedInferencesTwoWorker
 	}
 
 	expected := map[string]*emissionstypes.Inference{
-		"forecaster0": {Value: alloraMath.MustNewDecFromString("1.018004876148456175753781623")},
+		"forecaster0": {Value: alloraMath.MustNewDecFromString("1.055841253742177320400327600231111")},
```
relyt29, line 113: where are these random long numbers coming from? simulation csv?

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
```diff
@@ -188,8 +188,10 @@ func (p *SynthPalette) CalcWeightedInference(weights RegretInformedWeights) (Inf
 	}
 
 	// Normalize the running unnormalized network inference to yield output
-	if sumWeights.Lt(p.Epsilon) {
-		sumWeights = p.Epsilon
+	// TODO: Add this in the future as a parameter(?)
```
relyt29, line 191: can we create todos in linear not our codebase

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
```diff
@@ -188,8 +188,10 @@ func (p *SynthPalette) CalcWeightedInference(weights RegretInformedWeights) (Inf
...
+	weightsEpsilon := alloraMath.MustNewDecFromString("0.0000001")
```
relyt29, line 192: should this be a emissions module parameter, rather than something hardcoded

x/emissions/keeper/inference_synthesis/synth_palette_weight.go
relyt29: and i agree with the todo, it should be a emissions module parameter

x/emissions/keeper/keeper.go
```diff
@@ -163,8 +165,12 @@ type Keeper struct {
 	// map of (topic, forecaster, inferer) -> R^+_{ij_kk} regret of forecaster loss from comparing one-in loss with
 	// all network inferer (3rd index) regrets L_ij made under the regime of the one-in forecaster (2nd index)
 	latestOneInForecasterNetworkRegrets collections.Map[collections.Triple[TopicId, ActorId, ActorId], types.TimestampedValue]
-	// the forecaster (2nd index) regrets made under the regime of the same forecaster as a one-in forecaster
-	latestOneInForecasterSelfNetworkRegrets collections.Map[collections.Pair[TopicId, ActorId], types.TimestampedValue]
+
+	latestNaiveInfererNetworkRegrets               collections.Map[collections.Pair[TopicId, ActorId], types.TimestampedValue]
```
relyt29, line 169: if we're adding new data structures, then genesis.go _should_ be exporting them

x/emissions/types/keys.go
```diff
@@ -11,65 +11,70 @@ const (
...
+	TopicRewardNonceKey                               = collections.NewPrefix(48)
+	DelegateRewardPerShare                            = collections.NewPrefix(49)
+	PreviousPercentageRewardToStakedReputersKey       = collections.NewPrefix(50)
+	StakeRemovalsByBlockKey                           = collections.NewPrefix(51)
+	DelegateStakeRemovalsByBlockKey                   = collections.NewPrefix(52)
+	StakeRemovalsByActorKey                           = collections.NewPrefix(53)
+	DelegateStakeRemovalsByActorKey                   = collections.NewPrefix(54)
+	TopicLastWorkerCommitKey                          = collections.NewPrefix(55)
+	TopicLastReputerCommitKey                         = collections.NewPrefix(56)
+	TopicLastWorkerPayloadKey                         = collections.NewPrefix(57)
```
relyt29, line 72: I think this actually causes a problem for consensus on an upgrade from what I read somewhere if you change what keys are what so be VERY CAREFUL with this

I see these numbers have changed e.g. `TopicLastWorkerPayloadKey` is 57 here, but 58 elsewhere. I think we need to NOT be doing that since we're doing an upgrade. 

whatever thing you added, give it a number greater than the numbers we have, don't reassign numbers i'm like 80% confident that leads to consensus failures

relyt29 (changes requested):
good work. I left one critical comment, the rest are just take it or leave it advice

I reviewed it to the best of my ability given the time and stress constraints we have and to my pathetically weak eyes it looks ok

x/emissions/keeper/inference_synthesis/synth_palette_forecast_implied_test.go
guilherme-brandao: Just recalculated after we've made the change in the std dev

x/emissions/types/keys.go
guilherme-brandao: I didn't reassign numbers but good to know about this!

x/emissions/proto/emissions/v1/genesis.proto
```diff
@@ -135,16 +137,21 @@ message GenesisState {
   // all network inferer (3rd index) regrets L_ij made under the regime of the one-in forecaster (2nd index)
   repeated TopicIdActorIdActorIdTimeStampedValue latestOneInForecasterNetworkRegrets = 47;
   // the forecaster (2nd index) regrets made under the regime of the same forecaster as a one-in forecaster
-  repeated TopicIdActorIdTimeStampedValue latestOneInForecasterSelfNetworkRegrets = 48;
+  repeated TopicIdActorIdTimeStampedValue latestNaiveInfererNetworkRegrets = 48;
```
relyt29, line 140: i hope it's ok that `latestOneInForecasterSelfNetworkRegrets` -> `latestNaiveInfererNetworkRegrets` share the same number


but maybe this is a problem we can look at when we look at the upgrade, Fernando has more experience

relyt29 (approved):
:ship: :rocket:

PR #463, opened 2024-07-25
Improve Error Handling in FundTopic and RewardDelegateStake

relyt29:

This PR improves error handling in `FundTopic` and `RewardDelegateStake`

In the FundTopic case we return the error correctly but also return a non nil response for the RPC request, which may a) cause issues and confusion for consumers and b) prevent transactions from failing that should not be included in a block (not actually sure about this, just hypothecating)

In the case of RewardDelegateStake, we silently ignore errors allowing transactions to continue with errors, which allows Delegators to claim rewards multiple times. This is actually a serious theft-of-funds bug.

## Testing and Verifying

This change does not require additional test coverage.

This change does not have any documentation-related implications.


xmariachi (approved):
lgtm

PR #465, opened 2024-07-25
Silent errors in Remove{Delegate}Stake flow causes incorrect states

relyt29:

The stake removals are processed in ABCI End Blocker and if there is an error sending tokens or an error writing into the keeper state, the intent is to skip that stake removal and keep going. However, if the stake removal is incorrect, it actually can enter an error state where the tokens are sent, but the state machine is not updated, causing a theft of funds.

This PR uses a cache context to try writing the state into the state machine, and, if there is an error, skip writing the state.

## Testing and Verifying

Our existing test cases should confirm that this change does not change the overall flow of stake removal

There is no implication on documentation from this PR.

### DIFF
--- a/x/emissions/module/stake_removals.go
+++ b/x/emissions/module/stake_removals.go
@@ -25,6 +25,26 @@ func RemoveStakes(
 		return
 	}
 	for _, stakeRemoval := range removals {
+		// attempt writes in a cache context, only write finally if there are no errors
+		cacheSdkCtx, write := sdkCtx.CacheContext()
+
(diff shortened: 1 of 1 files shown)


RedBird96 (approved):
This is same logic for preventing reentrancy of smart contract...lol
Looks good to me!

xmariachi (approved):
lgtm

relyt29:
> This is same logic for preventing reentrancy of smart contract...lol Looks good to me!

not a coincidence!!

PR #466, opened 2024-07-25
GenerateForecastScores Should Update Forecasted Scores Not Inference Scores

relyt29:

This PR corrects a really nasty typo bug where we were updating inference scores in a forecaster function

## Testing and Verifying

Our test suite should have caught this. The fact that it didn't means we need further testing. However our existing tests will at least confirm that this PR is directionally correct...

Definitely with the addition of high-precision unit tests that match what the research team's python implementations numbers are will help with this issue, that's coming in a later PR.

This is an internal change with no documentation related implications


relyt29:
would particularly appreciate a review from @guilherme-brandao or @kpeluso - I think we had this wrong and never noticed, but maybe it's intentional that we were using the inference scores twice???

guilherme-brandao (approved):

PR #467, opened 2024-07-26
throw error during encoding if nil value trying to be encoded

relyt29:

Add an additional error when trying to encode data using quadruple.go, to make sure that nil values aren't stored or allowed.

## Testing and Verifying

This change is an additional error statement and the functionality should be covered by our existing tests.

There are no documentation implications for this PR


RedBird96 (approved):
lgtm

relyt29:
After digging some into the failing tests, it appears encoding nil values is a necessary feature if you want to support nil rangers that will iterate over all keys

I'm inclined to call this a wontfix until theres some substantiated bug that can actually clearly demonstrate a problem

PR #468, opened 2024-07-26
Sanity check create topic input sizes

relyt29:

This change validates additional parameters for the creation of new topics, namely it puts a limit on the size of string data like metadata, defaultArg etc

## Testing and Verifying

This change is a trivial rework without any additional test coverage.

no documentation implications


relyt29:
I lied. @kush-alloralabs you should probably want to write in the documentation that we are now limiting the length of certain fields on creation of new topics to 1024

PR #470, opened 2024-07-27
Michaelc/ora 2020 skim the top logic workers and reputers

RedBird96:

Remove skim top logic from close nonce

## Testing and Verifying

Need to make testing script

  - [ ] Code comments?
  - [ ] N/A


x/emissions/keeper/msgserver/msg_server_worker_payload.go
```diff
@@ -76,6 +76,13 @@ func (ms msgServer) InsertWorkerPayload(ctx context.Context, msg *types.MsgInser
...
 	if msg.WorkerDataBundle.InferenceForecastsBundle.Inference != nil {
 		inference := msg.WorkerDataBundle.InferenceForecastsBundle.Inference
+		isInfererRegistered, err := ms.k.IsWorkerRegisteredInTopic(ctx, topicId, inference.Inferer)
+		if err != nil {
+			return nil, errorsmod.Wrapf(err, "Error address is not registered in this topic")
+		}
+		if !isInfererRegistered {
+			return nil, errorsmod.Wrapf(err, "Error address is not registered in this topic")
+		}
 		err = ms.k.AppendInference(ctx, topicId, *nonce, inference)
```
kpeluso, line 86: for each actor (reputer, inferer, forecaster) we need to check if they already submitted data. doesn't matter at this time if we only take their first or latest response (assume all conditions for inclusion are still met), but we must be consistent with this policy

and certainly can't accept multiple bundles from the same actor for the same role at the time epoch

x/emissions/keeper/actor_utils/losses.go
```diff
@@ -98,77 +98,58 @@ func CloseReputerNonce(
...
-
-		lossBundlesFromTopReputers = append(lossBundlesFromTopReputers, lossBundlesByReputer[reputer])
-		stakesByReputer[reputer] = stake
+		stakesByReputer[bundle.ValueBundle.Reputer] = stake
 	}
+
 	// sort by reputer score descending
-	sort.Slice(lossBundlesFromTopReputers, func(i, j int) bool {
-		return lossBundlesFromTopReputers[i].ValueBundle.Reputer < lossBundlesFromTopReputers[j].ValueBundle.Reputer
+	sort.Slice(lossBundlesByReputer, func(i, j int) bool {
```
kpeluso, line 147: not sure you have to sort if you were already skimming the top as the bundles were being submitted...

x/emissions/keeper/actor_utils/worker.go
```diff
@@ -241,66 +200,22 @@ func insertForecastsFromTopForecasters(
...
-
-			/// Filtering done now, now write what we must for inclusion
-
-			// Get the latest score for each forecaster => only take top few by score descending
-			latestScore, err := k.GetLatestForecasterScore(ctx, topicId, forecast.Forecaster)
-			if err != nil {
-				continue
-			}
-			latestForecasterScores[forecast.Forecaster] = latestScore
+			latestForecaster = append(latestForecaster, forecast)
```
kpeluso, line 203: i have another comment about this, but this easily should happen when the bundle is being committed, not when it's being closed. this adds extra, unnecessary storage load if we allow duplicate entries prior to nonce close

x/emissions/keeper/actor_utils/worker.go
```diff
@@ -241,66 +200,22 @@ func insertForecastsFromTopForecasters(
...
-
 	// Though less than ideal because it produces less-acurate network inferences,
 	// it is fine if no forecasts are accepted
 	// => no need to check len(forecastsFromTopForecasters) == 0
 
 	// Ensure deterministic ordering
-	sort.Slice(forecastsFromTopForecasters, func(i, j int) bool {
-		return forecastsFromTopForecasters[i].Forecaster < forecastsFromTopForecasters[j].Forecaster
+	sort.Slice(latestForecaster, func(i, j int) bool {
+		return latestForecaster[i].Forecaster < latestForecaster[j].Forecaster
```
kpeluso, line 214: int his "1-tier", online design, we should only be adding values from actors of better previous score, meaning we "skim the top as we go", with each successive data bundle. this means that we may not need to sort at all, and we certainly shouldn't need to sort when we close the nonce

x/emissions/keeper/msgserver/msg_server_worker_payload.go
```diff
@@ -85,6 +92,13 @@ func (ms msgServer) InsertWorkerPayload(ctx context.Context, msg *types.MsgInser
...
 	if msg.WorkerDataBundle.InferenceForecastsBundle.Forecast != nil {
 		forecast := msg.WorkerDataBundle.InferenceForecastsBundle.Forecast
+		isForecasterRegistered, err := ms.k.IsWorkerRegisteredInTopic(ctx, topicId, forecast.Forecaster)
+		if err != nil {
+			return nil, errorsmod.Wrapf(err, "Error address is not registered in this topic")
+		}
+		if !isForecasterRegistered {
+			return nil, errorsmod.Wrapf(err, "Error address is not registered in this topic")
+		}
 		err = ms.k.AppendForecast(ctx, topicId, *nonce, forecast)
```
kpeluso, line 102: where are we "skimming the top as we go"? I also don't see any spot where we're enforcing a maximum number of bundles across actors per epoch (the actual "skimming" part. maybe I'm missing something?)?

kpeluso (changes requested):

x/emissions/keeper/msgserver/msg_server_worker_payload.go
xmariachi: The AppendInference method contains this logic afaiu.

x/emissions/module/autocli.go
```diff
@@ -748,7 +748,7 @@ func (am AppModule) AutoCLIOptions() *autocliv1.ModuleOptions {
 				},
 				{
 					RpcMethod: "InsertReputerPayload",
-					Use:       "insert-reputer-payload [sender] [reputer_data]",
+					Use:       "insert-reputer-payload [sender] [nonce] [topic_id] [reputer_data]",
```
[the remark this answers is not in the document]
RedBird96: Yes, I will update insert worker payload as well

x/emissions/keeper/msgserver/msg_server_worker_payload.go
RedBird96: yes, I will overwrite previous data

x/emissions/keeper/actor_utils/losses.go
RedBird96: When submitting data, only the append function is performed. Therefore, I think sorting should be done at close time.

xmariachi (approved):
lgtm

PR #478, opened 2024-08-01
Kenny/patch/add fee abstraction to txs

kpeluso:

* We were already sending registration, data ingress, and topic creation fees to ecosystem bucket, but we weren't updating effective revenue
* We weren't updating topic activation status correctly when new stake was placed

## Testing and Verifying

This change added tests and can be verified as follows:

*(example:)*
  - *Added unit test that validates ...*
  - *Added integration tests for end-to-end deployment with ...*
  - *Extended integration test for ...*

Adding ^these momentarily...

No

  - [ ] Code comments?
  - [ ] N/A


x/emissions/keeper/msgserver/msg_server_worker_payload.go
```diff
@@ -58,20 +56,14 @@ func (ms msgServer) InsertWorkerPayload(ctx context.Context, msg *types.MsgInser
...
-			"Error checking balance for sender: %v", &msg.Sender)
-	}
-	if !hasEnoughBal {
-		return nil, types.ErrDataSenderNotEnoughDenom
+		return nil, errorsmod.Wrapf(err, "Error getting params for sender: %v", &msg.Sender)
 	}
-
-	// Before creating topic, transfer fee amount from creator to ecosystem bucket
-	err = ms.k.SendCoinsFromAccountToModule(ctx, msg.Sender, mintTypes.EcosystemModuleName, sdk.NewCoins(fee))
+	err = sendEffectiveRevenueActivateTopicIfWeightSufficient(ctx, ms, msg.Sender, topicId, params.DataSendingFee, "insert worker payload")
```
RedBird96, line 64: Is there a reason to add topic fee revenue when inserting worker/reputation payload?

kpeluso: Responding for completeness here as well:

whenever we pay into the ecosystem bucket or add stake, both should impact the topic weight.
for revenue to impact topic weight, we must update effective revenue
and any increase to topic weight should be coupled with possible reactivation.
we automatically inactivate but “manually” activate (per relevant tx)
=> need to capture all the moments when reactivation is possible
inactivation is done automatically already in EndBlock

PR #480, opened 2024-08-02
Add back removed fields and restore the original order

fernandofcampos:
Closes: #XXX

Add back removed fields and restore the original order

## Testing and Verifying

*(Please pick one of the following options)*

This change is already covered by existing tests

 No

  - [ ] Code comments?
  - [x ] N/A


x/emissions/proto/emissions/v1/query.proto
```diff
@@ -912,6 +914,12 @@ message QueryTopicFeeRevenueResponse {
   ];
 }
 
+message QueryChurnableTopicsRequest {}
+
+message QueryChurnableTopicsResponse {
```
kpeluso, line 919: if we delete the query entirely, must the proto still stay?

kpeluso (changes requested):
1. that 1 question about needing to keep protos for removed queries
2. plz add `reserved` keyword where appropriate

x/emissions/proto/emissions/v1/node.proto
```diff
@@ -4,6 +4,9 @@ package emissions.v1;
 option go_package = "github.com/allora-network/allora-chain/x/emissions/types";
 
 message OffchainNode {
-  string node_address = 1;
-  string owner = 2;
+  string lib_p2p_key = 1;    // LibP2P key of the node
+  string multi_address = 2;  // Network address for accessing the node
+  string owner = 3;
+  string node_address = 4;
+  string node_id = 5;
```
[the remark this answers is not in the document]
fernandofcampos: To use reserved we would need to delete the fields. Can we delete the fields? Is there any client code using this, or it's just used internally? 

Deprecated only signals to clients that they should not use the field because it might get removed in the future.

x/emissions/proto/emissions/v1/query.proto
fernandofcampos: No, it can be removed. 

Question is, will you break anything if you remove the query?

fernandofcampos: Done

x/emissions/proto/emissions/v1/node.proto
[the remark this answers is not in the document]
kpeluso: with the "b7s removal" project we're updating the only "allora official" client, and it's prehistory in our ecosystem's history so not too worried about alternative community clients. valid concern otherwise

x/emissions/proto/emissions/v1/query.proto
kpeluso: same as above just reposting for completeness:
with the "b7s removal" project we're updating the only "allora official" client, and it's prehistory in our ecosystem's history so not too worried about alternative community clients. valid concern otherwise

kpeluso (approved):

PR #485, opened 2024-08-04
WIP Make Active Topic Set Management Scalable

kpeluso:
## Description

Currently active topics are in one large keyset that must be parsed in pages in order to find topics.
This is not scalable as the number of topics increases.
Furthermore, there are some minor issues with respect to how topic weight is calculated -- and EMA is often used where the non-EMA may be more appropriate.

This PR attempts to address these concerns by segmenting the list of active topics by block height of the next epoch.
So, there is no longer 1 keyset for all active topics.
Instead, we map block height to a bounded list of topics.
When a topic is (re)activated, we append it to this list but knock off the topic of lowest weight to maintain a bounded set size.
We also introduce a cache for topic weight and lowest topic weight per "active set for a block" to minimize compute.

## Link to Ticket or Issue resolved by this PR

https://linear.app/upshot/issue/ORA-2068/patch-active-topic-parsing

## Still Left Todo

* Implement GetTopicWeight() + cascade its usage + should couple with topic weight-related corrections from Sherlock
* Finish propagating per global param checklist for new param
* Error msgs + logging
* Actually apply the functions that are started here throughout EndBlock, etc.
* Testing + Documentation


kpeluso:
Should favor this solution instead of: https://github.com/allora-network/allora-chain/pull/450

kpeluso:
Resolved by both:
* https://github.com/allora-network/allora-chain/pull/542
* https://github.com/allora-network/allora-chain/pull/558

PR #481, opened 2024-08-02
Set initial target emission in block 0 to 0

guilherme-brandao:
## What is the purpose of the change

> Add a description of the overall background and high level changes that this PR introduces

*(E.g.: This pull request improves documentation of area A by adding ....)*

## Testing and Verifying

*(Please pick one of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added unit test that validates ...*
  - *Added integration tests for end-to-end deployment with ...*
  - *Extended integration test for ...*
  - *Manually verified the change by ...*

## Documentation and Release Note

  - [ ] Does this pull request introduce a new feature or user-facing behavior changes?

Where is the change documented?
  - [ ] Specification (`x/{module}/README.md`)
  - [ ] Allora documentation site `docs.allora.network` source code at: `https://github.com/allora-network/docs`
  - [ ] Code comments?
  - [ ] N/A

PR #486, opened 2024-08-04
[ORA-1613] Initial Value for emission per unit staked token = target

relyt29:

We calculate target emission and the emission rate on a monthly cadence. In the first month we have no previous history to use for the exponential moving average which returns e_i. Therefore, in that case, we use the target e_i instead of the exponential moving average, directly.

## Testing and Verifying

Our existing tests should at least cover that this does not break anything

No Documentation changes needed.


relyt29:
we should close #481 and use this instead

kpeluso (approved):

PR #515, opened 2024-08-11
Make Invariant Fuzzer runnable against release v0.3.0

relyt29:

Makes updates to the invariant / fuzzer test suite so that inferences can actually be sent during a fuzz run

No tickets created for this one

These changes are tests themselves. They have no implications relevant to documentation

### DIFF
--- a/test/invariant/inferences_test.go
+++ b/test/invariant/inferences_test.go
@@ -114,7 +114,7 @@ func insertWorkerBulk(
 	// Get Bundles
 	for _, worker := range workers {
 		workerData := generateSingleWorkerBundle(m, topic.Id, workerNonce, worker, workers)
-		wasErr = insertLeaderWorkerBulk(m, data, topic.Id, workerNonce, leaderWorker, workerData)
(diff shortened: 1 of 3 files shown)


xmariachi (commented):
Remove references to "leader" or "bulk"

guilherme-brandao (approved):

test/invariant/inferences_test.go
```diff
@@ -15,8 +15,8 @@ import (
 
 func doInferenceAndReputation(
 	m *testcommon.TestConfig,
-	leaderWorker Actor,
-	leaderReputer Actor,
+	_ Actor,
```
xmariachi, line 18: why are these params needed to be passed in at all if they are not used? Can't we remove these params?

xmariachi (approved):
lgtm, left one comment

test/invariant/inferences_test.go
relyt29: no they're part of an interface that is expected by the "list of actions a fuzz round can take"

some functions don't use them, some do, the ones that don't just ignore em

PR #519, opened 2024-08-14
Update Network Inference Calculation

guilherme-brandao:
Implements new recommendations from the research team, which include:

- Use the latest available network loss to calculate the network inference.
- If no previous network loss is available, calculate the median of the inferences to get the network inference.
- When calculating the forecast-implied inferences, if all inferers are new, calculate the median of the inferences of the forecasted elements to obtain the forecast-implied inference of the forecaster instead of calculating the mean.


relyt29:
this will need a network upgrade

pretty sure the query functions don't matter, those could be hot-patched

the inference synthesis requires a cosmovisor kill-old-binary, use new binary flow, pretty sure

but no state migrations, because nothing changed on disk

relyt29 (approved):
makes logical sense to me from a philosophy perspective, you'll have to see what test is breaking and fix it

x/emissions/keeper/inference_synthesis/synth_palette_forecast_implied.go
```diff
@@ -39,26 +39,30 @@ func (p *SynthPalette) CalcForecastImpliedInferences() (map[Worker]*emissionstyp
...
 			if p.AllInferersAreNew {
-				// If all inferers are new, take regular average of inferences
+				// If all inferers are new, calculate the median of the inferences
 				// This means that forecasters won't be able to influence the network inference when all inferers are new
-				// However this seeds losses for forecasters for future rounds
+				// However, this seeds losses for forecasters for future rounds
 
+				inferenceValues := make([]alloraMath.Dec, 0, len(sortedInferersInForecast))
 				for _, inferer := range sortedInferersInForecast {
 					if p.InferenceByWorker[inferer] != nil {
```
kpeluso, line 51: this not the best way to test for inclusion, right? cuz it throws. consider changing to `if _, ok := ...; ok {...}`

kpeluso (approved):

PR #520, opened 2024-08-14
Upgrade process improvement

okedeji:
To solidify the process of upgrading with docker. 
Now provide an image with tag appended with -upgrader

Yes tested manually. Downlaoded snapshot and then tested with the snapshot and it upgraded as expected


scripts/l1_node.sh
```diff
@@ -14,9 +14,18 @@ KEYRING_BACKEND=test                              #! Use test for simplicity, yo
...
-# SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-# "${SCRIPT_DIR}/restore_snapshot.sh"
+if [ "$RESTORE_S3_SNAPSHOT" == "true" ]; then
+    SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+    "${SCRIPT_DIR}/restore_snapshot.sh"
+fi
+
+BINARY=""
+if [ "$UPGRADE" == "true" ]; then
+    echo "You have set upgrade to true, please nake sure you are running the correct docker image (vx.x.x-upgrader)
```
fernandofcampos, line 24: Small typo ("nake")

docker-compose.yaml
```diff
@@ -2,16 +2,17 @@
 services:
   node:
     container_name: sample_validator
-    image: "alloranetwork/allora-chain:v0.2.14"
+    image: "alloranetwork/allora-chain:v0.3.1" # docker image without cosmovisor. use vx.x.x-upgrader for upgrade image
```
fernandofcampos, line 5: Shouldn't be v0.3.0?

fernandofcampos (commented):
Minor comments.

docker-compose.yaml
okedeji: so people are already using that v0.3.1 tag. I don't want to overide it such that it could break already running services

.github/workflows/build_push_upgrader_docker_hub.yml
```diff
@@ -0,0 +1,64 @@
+# This workflow will build and push a new container image to Amazon ECR,
+# and then will deploy a new task definition to Amazon ECS which will be run by Fargate when a release is created
+# This workflow will build and push a new container image to Docker Hub
+# when a release is created
+name: Build and Push docker image to Docker Hub
+
+on:
+  push:
```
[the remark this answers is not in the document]
okedeji: yeah this is brilliant.. and also will make tag and previous and upgrade binaries as input

scripts/l1_node.sh
```diff
@@ -75,12 +84,31 @@ if [ "x${STATE_SYNC_RPC1}" != "x" ]; then
...
+        --home=${APP_HOME} \
+        start \
+        --moniker=${MONIKER} \
+        --minimum-gas-prices=0${DENOM} \
+        --rpc.laddr=tcp://0.0.0.0:26657 \
+        --p2p.seeds=$SEEDS \
+        --p2p.persistent_peers $PEERS
+else
+    echo "Starting validator node without cosmovisor"
+    allorad \
```
xmariachi, line 106: Why it first runs with cosmovisor, then without?

okedeji: this is an if statement. if UPGRADE is true, then run with cosmovisor, else, run without as a normal binary

xmariachi: sorry! reviewed from mobile, I may have missed the else lol

docker-compose.yaml
okedeji: changed to v0.3.0 actually. we are only releasing the upgrader

PR #479, opened 2024-08-01
[WIP] High precision simulator comparison testing of section 5 calculations

relyt29:
## What is the purpose of the change

Adds unit tests to confirm that implementation of section 5 matches research simulations.

## Testing and Verifying

This change introduces new tests in emissions_test.go

## Documentation and Release Note

no Documentation changes.

PR #521, opened 2024-08-14
Mint module inflation rate and debug query query endpoint

relyt29:
## Changes

Adds a new debugging endpoint to the mint module query server that allows one to see detailed information that goes into the token inflation rate target adjustment calculation.

Also updates the "inflation" RPC endpoint to use the inflation over the circulating supply rather than all tokens currently in existence.

Sets the default parameters (not relevant for testnet that has already launched but will be relevant when it comes time to make a mainnet genesis) for blocks per month to 3 second block times

PROTO-1612

Tested manually, since they're query endpoint changes. Went over them live on testnet with Diederik as well to compare against simulator.

Will speak with Kush to let him know that this endpoint exists for documentation.


x/mint/keeper/query_server.go
```diff
@@ -46,10 +54,138 @@ func (q queryServer) Inflation(ctx context.Context, _ *types.QueryInflationReque
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get number of staked tokens")
+	}
+	lockedVestingTokensTotal, lockedVestingTokensPreseed,
+		lockedVestingTokensSeed, lockedVestingTokensTeam := GetLockedVestingTokens(
+		blocksPerMonth,
+		math.NewIntFromUint64(blockHeight),
+		moduleParams)
+	ecosystemLocked := ecosystemBalance.Add(ecosystemMintSupplyRemaining)
+	circulatingSupply := moduleParams.MaxSupply.Sub(lockedVestingTokensTotal).Sub(ecosystemLocked)
```
kpeluso, line 112: maybe we could call the function to get the circulating supply instead?

x/mint/keeper/query_server.go
```diff
@@ -46,10 +54,138 @@ func (q queryServer) Inflation(ctx context.Context, _ *types.QueryInflationReque
...
+		moduleParams.OneMonthSmoothingDegree,
+		previousRewardEmissionPerUnitStakedToken,
+	)
+	emissionPerMonth := GetTotalEmissionPerMonth(emissionPerUnitStakedToken, networkStakedTokens)
+	blockEmission := emissionPerMonth.
+		Quo(math.NewIntFromUint64(blocksPerMonth))
+	validatorCut := vPercent.Mul(blockEmission.ToLegacyDec()).TruncateInt()
+	alloraRewardsCut := blockEmission.Sub(validatorCut)
+
+	return &types.QueryEmissionInfoResponse{
```
kpeluso, line 163: v useful

kpeluso (commented):
Some small things I would change or want more clarity on. However curious to see how this impacts this PR's changes:
https://linear.app/alloralabs/issue/PROTO-1945/by-transferring-uallo-tokens-to-another-chain-via-ibc-the-reward

Obvi the calculation of circulating supply is potentially implicated by ^that ticket.
You might have averted that though with this PR's changes.

x/mint/keeper/query_server.go
```diff
@@ -46,10 +54,138 @@ func (q queryServer) Inflation(ctx context.Context, _ *types.QueryInflationReque
 		Mul(math.NewIntFromUint64(blocksPerMonth)).
 		Mul(math.NewInt(12)).
 		ToLegacyDec()
-	totalSupply := q.k.GetTotalCurrTokenSupply(ctx).Amount.ToLegacyDec()
-	inflation := EmissionPerYearAtCurrentBlockEmissionRate.Quo(totalSupply).MulInt64(100)
+	circulatingSupply, err := GetCirculatingSupply(ctx, q.k, moduleParams, blockHeight, blocksPerMonth)
+	if err != nil {
+		return nil, err
+	}
+	inflation := EmissionPerYearAtCurrentBlockEmissionRate.QuoInt(circulatingSupply).MulInt64(100)
```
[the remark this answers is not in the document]
relyt29: yes that's @jmdkastro's definition of circulating, not including ecosystem module's balance. I'm ambivalent

x/mint/keeper/query_server.go
relyt29: yeah idk. It's less efficient if we do that because the circulating supply function calls variables that we then use later so it's two lookups instead of one for the same data, but on the other hand the code looks cleaner to read, is one less place to have to worry about differences, etc

relyt29:
> curious to see how this impacts this PR's changes: https://linear.app/alloralabs/issue/PROTO-1945/by-transferring-uallo-tokens-to-another-chain-via-ibc-the-reward
> 
> Obvi the calculation of circulating supply is potentially implicated by ^that ticket. You might have averted that though with this PR's changes.

That ticket is dealt with in #479 - the circulating supply doesn't care about IBC at all after that PR.

x/mint/keeper/query_server.go
relyt29: oh sorry, thought you were asking about in abci.go `GetEmissionPerMonth` - I'm still on the fence there

for query server yes makes sense

kpeluso: few look ups vs abstraction for somewhat involved calc -- I lean abstraction

PR #524, opened 2024-08-15
Tune Emissions Module Parameters

relyt29:

Tweaks default values to emissions module parameters to values set from research simulations

https://linear.app/alloralabs/issue/PROTO-2052/apply-parameter-tunings

These parameter tunings should be covered by our existing unit and integration tests.

## Still Left Todo

See follow on PR that has protobuf-breaking changes and require module upgrade (v0.4.0)


kpeluso (approved):
Just have to remember to send the UpdateParams tx too when we're ready

PR #525, opened 2024-08-15
Remove min_effective_topic_revenue emissions module parameter

relyt29:

This deletes the `min_effective_topic_revenue` emissions module parameter. Justification: "... it’s redundant with min_topic_weight and I can imagine topics getting super low revenue but being valuable (high stake), so weight is the parameter to purge on."

https://linear.app/alloralabs/issue/PROTO-2052/apply-parameter-tunings

Covered by existing unit tests and integration tests.

## Still Left Todo

I am unsure if this is the correct way to delete a protobuf parameter, to be quite honest. Would appreciate @fernandofcampos making sure that I moved v2->v3 correctly for the network upgrade where this parameter will be deleted


x/emissions/migrations/v3/migrate.go
```diff
@@ -0,0 +1,92 @@
...
+	err := proto.Unmarshal(oldParamsBytes, &oldParams)
+	if err != nil {
+		return errors.Wrapf(err, "failed to unmarshal old parameters")
+	}
+
+	// DIFFERENCE BETWEEN OLD PARAMS AND NEW PARAMS:
+	// ADDED:
+	// 		N/A
+	// REMOVED:
+	// 		MinEffectiveTopicRevenue
```
kpeluso, line 43: perf

kpeluso (commented):
@relyt29 @fernandofcampos would be great to eventually autogenerate new proto versions + these migration files. Also have linter to throw if we're not properly doing a migration.

kpeluso (approved):

PR #528, opened 2024-08-16
Add forecast element limit

RedBird96:
Added cap to number of forecast elements per forecast. It's selected only top N elements by score

This also contains 
[PROTO-1958](https://linear.app/alloralabs/issue/PROTO-1958/insecure-implementation-of-transaction-validation)
[PROTO-1874](https://linear.app/alloralabs/issue/PROTO-1874/registerlegacyaminocodec-is-not-implemented-in-appmodule)

## Still Left Todo


x/emissions/types/params.go
```diff
@@ -53,6 +53,7 @@ func DefaultParams() Params {
 		MinEffectiveTopicRevenue:            alloraMath.MustNewDecFromString("0.00000001"), // we no stop dripping from the topic's effective revenue when the topic's effective revenue is below this
 		HalfMaxProcessStakeRemovalsEndBlock: uint64(40),                                    // half of the max number of stake removals to process at the end of the block, set this too big and blocks require too much time to process, slowing down consensus
 		DataSendingFee:                      cosmosMath.NewInt(10),                         // how much workers and reputers must pay to send payload
+		MaxTopForecasterElementsToSubmit:    uint64(3),                                     // top forecast elements by score
```
xmariachi, line 56: A forecaster can only retain 3 forecast elements in its forecast? 
We have a limit of 48 `MaxTopInferersToReward`. 
The way I see it, this value should be >= `MaxTopInferersToReward`

x/emissions/proto/emissions/v2/params.proto
```diff
@@ -134,5 +134,7 @@ message Params {
     (gogoproto.customtype) = "cosmossdk.io/math.Int",
     (gogoproto.nullable) = false,
     (amino.dont_omitempty) = true
-  ];  // payload sending fee for reputer or worker
+  ];
+  // payload sending fee for reputer or worker
+  uint64 max_top_forecaster_elements_to_submit = 45;  // max number of top forecasters by score to reward
```
kpeluso, line 139: would rename to `max_elements_per_forecast`

x/emissions/types/params.go
kpeluso, line 56: Default of 12 plz

x/emissions/keeper/msgserver/msg_server_stake.go
```diff
@@ -15,6 +15,10 @@ import (
 
 // Function for reputers to call to add stake to an existing stake position.
 func (ms msgServer) AddStake(ctx context.Context, msg *types.MsgAddStake) (*types.MsgAddStakeResponse, error) {
+	if err := msg.Validate(); err != nil {
```
kpeluso, line 18: did you do a full sweep of the msgserver? seems like we should also have this in multiple places even within each of the add/remove delegate functions and elsewhere

x/emissions/keeper/msgserver/msg_server_worker_payload.go
```diff
@@ -107,11 +107,27 @@ func (ms msgServer) InsertWorkerPayload(ctx context.Context, msg *types.MsgInser
...
 		// Remove duplicate forecast element
 		acceptedForecastElements := make([]*types.ForecastElement, 0)
 		seenInferers := make(map[string]bool)
+
+		for _, el := range forecast.ForecastElements {
+			score, err := ms.k.GetLatestInfererScore(ctx, forecast.TopicId, el.Inferer)
+			if err != nil {
+				continue
+			}
+			latestForecastScores[el.Inferer] = score
```
kpeluso, line 120: would instead call this `latestScoresForForecastedInferers` cuz otherwise it sounds like "forecaster scores" which are a different concept

x/emissions/module/module.go
```diff
@@ -46,8 +46,10 @@ func NewAppModule(cdc codec.Codec, keeper keeper.Keeper) AppModule {
 func (AppModule) Name() string { return types.ModuleName }
 
 // RegisterLegacyAminoCodec registers the state module's types on the LegacyAmino codec.
-// New modules do not need to support Amino.
-func (AppModule) RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {}
+func (AppModule) RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
+	// TODO: Implement full Amino codec registration if needed in the future
```
kpeluso, line 50: What's going on here?

kpeluso (changes requested):
Could use some answers to some questions, some small stylistic fixes, and think some validation needs to be applied more broadly.

x/emissions/types/params.go
kpeluso: Indeed. This is however the most expensive piece. And we'll be lowering 48-32 anyway. Asked to raise max number of elements to 12 => size of largest item in every ValueBundle is 12x12=144, which I think is a fine place to start given 32 inferers and unknown network compute bandwidth at this stage.

kpeluso (approved):

x/emissions/types/params.go
kpeluso: should be _at most_ `MaxTopInferersToReward` but yes what @RedBird96 said

PR #529, opened 2024-08-16
Fee Revenue Decay Rate set to per topic's epoch rather than global fixed constant

relyt29:

Currently, we use a global parameter to drip revenue to the same extent across all topics, regardless of their epoch length. This is somewhat unfair, as topics of larger epochs then get dripped comparatively less than those with shorter epoch lengths.

@jmdkastro proposes a more appropriate way of updating effective revenue, that actually takes the topic epoch length into consideration (through calculation of "epochs per week"). Relevant text from the whitepaper:

The quantity `C_{t,i}` decays _each epoch_ by subtracting an amount `∆ C_{t,i} = N_{epochs,w} * C_{t,i}` (where `N_{epochs,w}` is the number of epochs per week) and thus captures recent revenue.

This PR modifies the keeper.go `DripTopicFeeRevenue` function to do the following: identify if we have dripped yet this epoch. If we have not, then calculate the number of blocksPerWeek based on the blocksPerMonth, get the epochs per block from the topic, then do the math to calculate how much to decay for this epoch. Then it overwrites the fee revenue for this topic, and marks this epoch as having been decayed.

This change add a new data structure to the keeper called `lastDripBlock`. This change also deletes the global emissions module parameter `TopicFeeRevenueDecayRate`

PROTO-2150
RES-302

Our existing unit tests were adapted to fit this new change, as some of them had broke from the different algorithms.


x/emissions/keeper/keeper.go
```diff
@@ -156,6 +156,9 @@ type Keeper struct {
 	// map of (topic) -> unfulfilled nonces
 	unfulfilledReputerNonces collections.Map[TopicId, types.ReputerRequestNonces]
 
+	// map of (topic) -> last dripped block
+	lastDripBlock collections.Map[TopicId, BlockHeight]
```
fernandofcampos, line 160: Do we need a migration for this?

x/emissions/keeper/genesis.go
```diff
@@ -1345,6 +1367,25 @@ func (k *Keeper) ExportGenesis(ctx context.Context) (*types.GenesisState, error)
 		return nil, errors.Wrap(err, "failed to get previous percentage reward to staked reputers")
 	}
 
+	// openWorkerWindows
+	openWorkerWindows := make([]*types.BlockHeightAndTopicIds, 0)
```
guilherme-brandao, line 1371: Do we really need to create a specific structure just for the import/export? Is it the only way?

guilherme-brandao (approved):
Looks good! Just left a comment

x/emissions/keeper/genesis.go
relyt29: I don't really understand the import export, but based on my partial understanding yes, this is what we have to do

PR #535, opened 2024-08-20
Added telemetry and removed unnecessary logs

guilherme-brandao:
- Remove unnecessary `fmt.Print` statements
- Added telemetry + new endblock metric for emissions module


relyt29 (approved):
The integration test is failing because it's flakey. if you re run it it will likely pass

we should probably create a ticket to look into it at some point. I am not sure why the integration test has suddenly decided to be flakey when it wasn't before

PR #536, opened 2024-08-20
[PROTO-2211] Ensure all structs always have all field members present

relyt29:

This PR turns on the `exhaustruct` golang-ci linter.

The linter ensures that when a struct is created, all of its fields are initialized with it, for example below:

```golang
type Person struct {
    name: string;
    job: string;
    age: uint64;
}

//... somewhere else in the codebase

func foo() {
    joe := Person {
        name: "Joe"
    } // <--- never filled out job or age, make sure this never happens in our codebase
}
```

PROTO-2211

These changes are tested by our Github CI action using golang-ci. These should be non-logic changing changes and should have no impact on the business logic of our software. Thus our existing test suite should be sufficient, and this is internal so there is no need for documentation updates.


fernandofcampos:
It found "issues" in a few files. You can run `make lint` locally to check.

kpeluso:
Per @relyt29 's instructions, DO NOT MERGE THIS to `dev` until https://github.com/allora-network/allora-chain/pull/538 is merged to `dev` !

x/emissions/keeper/topic_activation.go
```diff
@@ -41,15 +41,21 @@ func (k *Keeper) GetActiveTopicIdsAtBlock(ctx context.Context, block BlockHeight
...
+func (k *Keeper) GetLowestActiveTopicWeightAtBlock(
+	ctx context.Context,
+	block BlockHeight,
+) (topicIdAndWeight types.TopicIdWeightPair, noPrior bool, err error) {
+	topicIdAndWeight, err = k.blockToLowestActiveTopicWeight.Get(ctx, block)
 	if err != nil {
 		if errors.IsOf(err, collections.ErrNotFound) {
-			return types.TopicIdWeightPair{}, true, nil
+			return types.TopicIdWeightPair{
+				TopicId: 0,
```
kpeluso, line 52: Consider dedicated `ZERO_TOPIC_ID` constant, but nbd

relyt29: there's only two places this occurs, in this file and in quantile_test.go, don't think a constant is warranted

kpeluso (approved):

x/emissions/keeper/inference_synthesis/forecast_implied.go
```diff
@@ -84,8 +88,12 @@ func CalcForecastImpliedInferences(args CalcForecastImpliedInferencesArgs) (
 				}
 
 				forecastImpliedInference := emissionstypes.Inference{
-					Inferer: forecaster,
-					Value:   medianValue,
+					TopicId:     args.TopicId,
```
xmariachi, line 91: In this case, I understand we are adding fields that are not used(they were not previously), am I right? 
Would it make sense to use nolint here? 
I worry it may make data grow even bigger. 
Also if those other fields are not used, makes me think if `Inference` is the right structure to use here.

relyt29: I do not endorse using nolint because it means leaving those values as zero or nil, when they should not be
here for example the topic id is indeed the args.TopicId, regardless of whether it's used

and you are correct, Inference is not the right structure to use here: PROTO-2655

xmariachi (approved):
Seeing how tedious it was to review all this, I can imagine about doing it. 
I can only thank you for this ^^

PR #537, opened 2024-08-21
Updating forecaster score ratio

guilherme-brandao:
Added missing update to the forecaster score ratio


relyt29 (approved):
thanks, I would have handled this myself but I didn't want to screw up where to do the update

PR #538, opened 2024-08-21
[PROTO-2211] Refactor Inference Synthesis to use Functions instead of "Builder Pattern"

relyt29:

in #536 we enable a linter to force all struct member fields to be present. The linter then clashes with the `NetworkInferenceBuilder` which uses object oriented style programming to first create a NetworkInferenceBuilder struct and then call functions on it to calculate various values and set fields on the struct. Obviously, not all struct member fields are present when you write code like this.

This PR rewrites `network_inference_builder.go` to operate in a procedural programming style instead. It removes the `NetworkInferenceBuilder` struct and the `SynthPalette` struct. The "builder pattern" and the object oriented style of programming where methods mutate the state of the struct "object" is removed. Instead, each network inference calculation is performed by a function that takes its arguments explicitly as opposed to a struct with many hidden fields unseen at time of function declaration. These functions do not mutate their inputs or make state changes that persist after the life of the function. This will assist readers of the code and linter tools to properly check that only necessary arguments are passed and eliminates the possibility of a whole class of bugs related to object mutation. This code also removes ~8 different Clone operations, which required O(n) copies of every array passed through inference synthesis, and 4 different sort operations over every inference, so this PR should improve performance as well.

The vast majority of changes in this PR are non-changing to the logic of inference synthesis, just a rewrite of the code structure. There are two logic-changing operations in this PR, detailed below:

### Logic Changing operations in this refactor

While working on this refactor, the refactor uncovered a bug where the synth palette clone function was not setting the `allInferersAreNew` field. This bug went undiscovered by our test suite. In fact, the test suite relied on the bug being present to pass some of the CSV tests. If we had the exhaustruct linter turned on we would have discovered this bug.

This refactor additionally uncovered a "bug" in the treatment of the error case where a loss bundle is passed to inference synthesis where a forecaster exists in the list of forecasts but not in a mapping of forecaster to their forecasts. After this PR, we now return an error from inference synthesis, rather than skipping that forecaster and doing a `continue` in the for loop.

PROTO-2211

These changes are tested by the tests in `x/emissions/keeper/inference_synthesis/network_inference_builder_test.go` and `x/emissions/keeper/queryserver/query_server_inferences.go` - that file includes the `GetSimulatedValuesGetterForEpochs` - so the refactored code is guaranteed to be within epsilon 5 of the simulator (and before this PR's) values, given sample data. This ensures this refactor is indeed only a refactor with no actual impact/code changes.

This is an internal change, so there's no impact on documentation.


x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -7,136 +7,99 @@ import (
...
-}
-
 // Calculates the network combined inference I_i, Equation 9
-func (b *NetworkInferenceBuilder) SetCombinedValue() *NetworkInferenceBuilder {
-	b.logger.Debug(fmt.Sprintf("Calculating combined inference for topic %v", b.palette.TopicId))
-	palette := b.palette.Clone()
+func GetCombinedInference(logger log.Logger, inPalette SynthPalette) (
+	weights RegretInformedWeights, combinedInference InferenceValue, err error) {
+	logger.Debug(fmt.Sprintf("Calculating combined inference for topic %v", inPalette.TopicId))
+	palette := inPalette.Clone()
```
guilherme-brandao, line 16: I think we don't need to clone the palettes in this case and in other cases below, please double check!

guilherme-brandao (commented):
directionally good! We should double-check error handling and the necessity of cloning palettes in each case (a benefit of this new, more functional approach).

x/emissions/keeper/inference_synthesis/network_inference_builder.go
relyt29: Looked into this, it seems the palette cloning is necessary - otherwise the CSV simulator values don't match - so something about the way the data is shared means the copy is necessary. I am looking more into it

relyt29:
still haven't figured out why we need clones before running `CalcWeightedInference` - I can't seem to find which line of code actually mutates the palette that is passed to it but clearly something is because when you remove the clones, the code fails tests

I have to step away from this for now - it is a distraction

relyt29:
the last commit renames files so if you want to see a closer commit diff look at all the commits before the last one

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -218,134 +337,244 @@ func (b *NetworkInferenceBuilder) SetOneOutInfererValues() *NetworkInferenceBuil
...
 	// Get one-in regrets for the forecaster and the inferers they provided forecasts for
-	for _, inferer := range palette.Inferers {
-		regret, _, err := palette.K.GetOneInForecasterNetworkRegret(palette.Ctx, palette.TopicId, oneInForecaster, inferer)
+	infererToRegretForSingleForecaster := make(map[Inferer]*Regret)
+	infererToInferenceForSingleForecaster := make(map[Inferer]*emissions.Inference)
+	for _, inferer := range inferers {
+		regret, _, err := k.GetOneInForecasterNetworkRegret(ctx, topicId, oneInForecaster, inferer)
 		if err != nil {
-			return alloraMath.Dec{}, errorsmod.Wrapf(err, "Error getting one-in forecaster regret")
+			return alloraMath.Dec{}, errorsmod.Wrapf(err, "Calc one in value error getting one-in forecaster regret")
```
guilherme-brandao, line 541: Error message out of context

guilherme-brandao (dismissed):

x/emissions/keeper/inference_synthesis/forecast_implied.go
```diff
@@ -0,0 +1,182 @@
...
+	infererToInference map[Inferer]*emissionstypes.Inference,
+	infererToRegret map[Inferer]*Regret,
+	forecasters []Forecaster,
+	forecasterToForecast map[Forecaster]*emissionstypes.Forecast,
+	forecasterToRegret map[Forecaster]*Regret,
+	networkCombinedLoss alloraMath.Dec,
+	epsilonTopic alloraMath.Dec,
+	pNorm alloraMath.Dec,
+	cNorm alloraMath.Dec,
+) (
```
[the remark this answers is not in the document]
relyt29: so I disagree with struct parameters and the reason why is because one of the linters, will pick up if an input is not being used and flag it for compile / fail CI. Even with exhaustruct on which is coming in my next PR, you won't get that if you start passing a struct which is basically a context struct. If at a later date somebody comes along and modifies this function to not need or use one of its parameters, the compiler will force them to update the function definition and adjust everywhere it's called.

Is this way uglier, yes, is it significantly uglier or harder to read, not really, just more scrolling down when you scroll through code. This is a more declarative approach. And escpecially with this important, math heavy inference synthesis code, I would rather be more declarative and bonedead obvious in what's going on

x/emissions/keeper/inference_synthesis/network_inference_builder.go
relyt29: what's the issue here? I think the error message is correct?

x/emissions/keeper/inference_synthesis/forecast_implied.go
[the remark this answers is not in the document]
relyt29: put it another way


nobody has ever gotten burned from having too many args to a function in their code

(actually this is not true in languages like C iirc, but the golang compiler can behind my back stuff this into a struct under the hood, shh)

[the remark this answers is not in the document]
guilherme-brandao: having a long list of arguments still increases the cognitive load for developers, especially for future maintenance. It’s harder to immediately understand what’s happening when the function signature is bloated imo, and it becomes more error-prone when passing the wrong values in large lists of parameters.

Is it impossible to get linter coverage in context struct? If yes, why not at least group static parameters into a logical struct? It would enhance maintainability in the long term and reduce refactor scope when adding or removing fields​.

I understand the desire to keep things "bone-dead obvious," but declarative doesn’t have to mean verbose. Grouping parameters into domain-specific structs still lets the code be clear while reducing the noise of many unrelated arguments. Being more declarative can also mean being more concise, and it's easier to reason about logical groups of data, especially in math heavy code​​.

While I agree that we should keep things declarative and obvious, I think organizing parameters into structs still helps maintain the code’s clarity and readability. Instead of scrolling through a large list of arguments, we can reason about the functionality more easily by working with grouped, coherent structs. The clarity here comes from proper modeling rather than simply expanding the number of args​​.

Also, go compiler might optimize large argument lists under the hood by "stuffing them into a struct", but this does not solve the core problem of code readability, clarity, and maintainability, which are human-facing issues.

x/emissions/keeper/inference_synthesis/network_inference_builder.go
guilherme-brandao: Here we are getting one-in regrets, not calculating

guilherme-brandao: yeah we calc one-in inferences, not regrets. The regrets we just fetch.

guilherme-brandao: ah ok! sorry just double-checked it, I'm good with this! But just make sure to follow the same pattern in other functions

x/emissions/keeper/inference_synthesis/forecast_implied.go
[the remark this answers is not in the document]
relyt29: yeah I don't agree with you and I don't view the readability impact as that bad, even in this explicit example CalcForecastImpliedInferences, the only kind of "tertiary" not directly used parameter is the logger. Everything else is pretty core and used directly in the function, and I wouldn't want it obscured in some struct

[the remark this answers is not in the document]
spooktheducks: @guilherme-brandao 
>  think it would be a good idea to have a shared struct with static properties only (e.g., logger, topicId, networkLoss, moduleParams, etc.).
Good thought in theory, let's keep going

@guilherme-brandao 
> It would avoid functions with this huge number of args
If we're at this point (imho probably so in this case), then the entire function signature needs to be a struct, and we need to enforce using named struct fields and `exhauststruct`.  It needs to be a stricter version of Python `kwargs` if it's going to provide guaranteed value to us on the team

@relyt29 
> one of the linters, will pick up if an input is not being used and flag it for compile / fail CI
What do you mean- which linter?  Opinion on keeping it enabled?  That sounds like a stupid fucking linter, but maybe I'm not understanding it fully.  Sometimes you have to pass structs into functions and not every member is used.  Am I misunderstanding?  Can we make it less stupid?  And what is its value proposition?  Unused args are reasonable only in the case of conforming to interfaces, but this sounds more rigid?  Sry lack the full picture

@guilherme-brandao 
> but declarative doesn’t have to mean verbose
Yeah, but explicit and verbose are not fully overlapping, and explicit is a must.  Go's culture leans towards explicit, and orthogonally, it also often leans towards verbose.  My advice: be explicit always, whether with Go or any other language.  Fuck magic.  Be verbose when being explicit requires it.  If we dislike verbosity, we can make simple IDE tools to hide it (I have made tools to hide `if err != nil` blocks when I press a simple keystroke -- in other words, let's think creatively about problems like these, don't change code because of our friction with it, let's change our view of code instead -- metamagic)

@guilherme-brandao 
> Grouping parameters into domain-specific structs still lets the code be clear
Agree.  This is good for human habits of mind.  If a linter dislikes it, disembowel the linter. Lose it or rewrite it.  Human cognitive sympathy with code is very important, especially on complex projects

@guilherme-brandao 
> Also, go compiler might optimize large argument lists under the hood by "stuffing them into a struct"
This is the least of our worries, tbh.  If you think it has become a problem, run `pprof` and prove it, and then we fix it.  No premature optimization, only data

Summary of my opinion:
- Be as explicit as possible, always
- Collapsing a large argument list into a struct (`MyFuncArgs{}` etc) is almost always good once the arg list gets too big, but we need to enforce 1) exhauststruct and 2) requiring field names on struct literals.  Think Python `kwargs` with `exhaust` rules
- Grouping related subdomains of a struct when it makes sense for human cognition is a win for mental energy
- However, if you're going to pass a pre-defined struct full of defaults into one of these subdomain fields, a lot of care should be taken (again, think about human cognition, and ways people might accidentally overlook the details of the interface you are creating by doing this).  This has to be a very well-named and well-defined default struct val.  For me, this is very much a case-by-case assessment, and I can't give many global rules
- Don't worry about struct-packing for performance improvements right now.  We have much bigger performance issues to solve; this will not move the needle on that at all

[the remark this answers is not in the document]
kpeluso: My thoughts are in the description of this PR, which presents and alternative path for the intention of this PR:
https://github.com/allora-network/allora-chain/pull/621

x/emissions/keeper/inference_synthesis/weight.go
```diff
@@ -11,10 +13,19 @@ import (
 
 // Given the current set of inferers and forecasters in the palette, calculate their
 // weights using the current regrets
-func (p *SynthPalette) CalcWeightsGivenWorkers() (RegretInformedWeights, error) {
+func calcWeightsGivenWorkers(
+	logger log.Logger,
```
xmariachi, line 17: Is it preferred to pass in the Logger than to pass the Context? I see merit to both cases but just to make sure we are aligned on this decision and aim to keep it as a new standard.

Using Logger is lighter and more explicit, and because we don't normally use advanced use of Context in keeper, then it is simpler. 
Using Context adds uniformity and flex through all cases, regardless you need sth else from the Context or not.

x/emissions/keeper/inference_synthesis/forecast_implied.go
```diff
@@ -0,0 +1,182 @@
...
+	"fmt"
+
+	errorsmod "cosmossdk.io/errors"
+	"cosmossdk.io/log"
+	alloraMath "github.com/allora-network/allora-chain/math"
+	emissionstypes "github.com/allora-network/allora-chain/x/emissions/types"
+)
+
+// Calculate the forecast-implied inferences I_ik given inferences, forecasts and network losses.
+// Calculates R_ijk, w_ijk, and I_ik for each forecast k and forecast element (forcast of worker loss) j
```
xmariachi, line 13: typo "forcast" -> "forecast"

x/emissions/keeper/inference_synthesis/weight.go
relyt29: no strong opinion, slight preference for the logger for the reasons you outline

note that in other parts of the codebase, the context is used for things unrelated to the logger (majority in functions that interact with the keeper which takes ctx as an argument), so I doubt that there are many cases we would find even if we looked for them where we could substitute a ctx for just the logger

x/emissions/keeper/inference_synthesis/forecast_implied.go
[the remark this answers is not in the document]
relyt29: resolution: Every function gets its own input struct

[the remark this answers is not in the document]
relyt29: resolved in [48c996d](https://github.com/allora-network/allora-chain/pull/538/commits/48c996d646c0d933b04b8be3e3fdf68c77b65a57)

relyt29:
# Option A
massive named struct arg for every function with a different function arguments list. downside is perhaps a bit verbose

```golang
type CalcForecastImpliedInferencesArgs struct {
	Logger               log.Logger
	TopicId              uint64
	AllInferersAreNew    bool
	Inferers             []Inferer
	InfererToInference   map[Inferer]*emissionstypes.Inference
	InfererToRegret      map[Inferer]*Regret
	Forecasters          []Forecaster
	ForecasterToForecast map[Forecaster]*emissionstypes.Forecast
	ForecasterToRegret   map[Forecaster]*Regret
	NetworkCombinedLoss  alloraMath.Dec
	EpsilonTopic         alloraMath.Dec
	PNorm                alloraMath.Dec
	CNorm                alloraMath.Dec
}

func CalcForecastImpliedInferences(args CalcForecastImpliedInferencesArgs) (
	forecasterToForecastImpliedInference map[Forecaster]*emissionstypes.Inference,
	infererToRegretOut map[Inferer]*Regret,
	forecasterToRegretOut map[Forecaster]*Regret,
	err error,
) {
```


# Option B

Every parameter gets its own function argument, downside is theoretically you could screw up the order of arguments and have a really nasty bug

```golang
func CalcForecastImpliedInferences(
	Logger log.Logger,
	TopicId uint64,
	AllInferersAreNew bool,
	Inferers []Inferer,
	InfererToInference map[Inferer]*emissionstypes.Inference,
	InfererToRegret map[Inferer]*Regret,
	Forecasters []Forecaster,
	ForecasterToForecast map[Forecaster]*emissionstypes.Forecast,
	ForecasterToRegret map[Forecaster]*Regret,
	NetworkCombinedLoss alloraMath.Dec,
	EpsilonTopic alloraMath.Dec,
	PNorm alloraMath.Dec,
	CNorm alloraMath.Dec,
) (
	forecasterToForecastImpliedInference map[Forecaster]*emissionstypes.Inference,
	infererToRegretOut map[Inferer]*Regret,
	forecasterToRegretOut map[Forecaster]*Regret,
	err error,
) {
```

# Option C 

Make some small struct that just holds tertiary information and leave the real important stuff as args.
Downside is you will certainly either pass
unnecessary args or have to make 5 different "context" type structs
downside two is that it obscures what the actual function arguments are, you have to go look up the struct definition

```golang
type InferenceSynthesisContext struct {
	Logger               log.Logger
	TopicId              uint64
	EpsilonTopic         alloraMath.Dec
	PNorm                alloraMath.Dec
	CNorm                alloraMath.Dec
}

func CalcForecastImpliedInferences(
    infSynthCtx InferenceSynthesisContext,
    Inferers             []Inferer
	InfererToInference   map[Inferer]*emissionstypes.Inference
	InfererToRegret      map[Inferer]*Regret
	AllInferersAreNew    bool
	Forecasters          []Forecaster
	ForecasterToForecast map[Forecaster]*emissionstypes.Forecast
	ForecasterToRegret   map[Forecaster]*Regret
	NetworkCombinedLoss  alloraMath.Dec
) (
	forecasterToForecastImpliedInference map[Forecaster]*emissionstypes.Inference,
	infererToRegretOut map[Inferer]*Regret,
	forecasterToRegretOut map[Forecaster]*Regret,
	err error,
) {
```


# Option D

Make a big ass Synth Palette struct that holds every possible argument for every possible function.
Downside a reader does not know what actual parts of the synth palette 
are used in the function they are looking at based on the function signature.

```golang
type SynthPalette struct {
	Ctx               sdk.Context
	K                 keeper.Keeper
	Logger            log.Logger
	TopicId           TopicId
	AllInferersAreNew bool
	// Should use this as a source of truth regarding for which inferers to have data calculated
	// i.e. if an inferer is not present here, calculate a network inference without their data
	// Must be unique values
	Inferers          []Worker
	InferenceByWorker map[Worker]*emissions.Inference
	// Must respect the order of sister `inferers` property
	InfererRegrets map[Worker]*alloraMath.Dec
	// Should use this as a source of truth regarding for which forecasters to have data calculated
	// i.e. if an forecaster is not present here, calculate a network inference without their data
	// Must be unique values
	Forecasters                      []Worker
	ForecastByWorker                 map[Worker]*emissions.Forecast
	ForecastImpliedInferenceByWorker map[Worker]*emissions.Inference
	// Must respect the order of sister `forecasters` property
	ForecasterRegrets   map[Worker]*alloraMath.Dec
	NetworkCombinedLoss Loss
	EpsilonTopic        alloraMath.Dec
	EpsilonSafeDiv      alloraMath.Dec
	PNorm               alloraMath.Dec
	CNorm               alloraMath.Dec
}


func CalcForecastImpliedInferences(
	palette SynthPalette,
)
	paletteOut SynthPalette,
	err error,
) {
```

xmariachi:
> # Option A
> massive named struct arg for every function with a different function arguments list. downside is perhaps a bit verbose
> 
> ```go
> type CalcForecastImpliedInferencesArgs struct {
> 	Logger               log.Logger
> 	TopicId              uint64
> 	AllInferersAreNew    bool
> 	Inferers             []Inferer
> 	InfererToInference   map[Inferer]*emissionstypes.Inference
> 	InfererToRegret      map[Inferer]*Regret
> 	Forecasters          []Forecaster
> 	ForecasterToForecast map[Forecaster]*emissionstypes.Forecast
> 	ForecasterToRegret   map[Forecaster]*Regret
> 	NetworkCombinedLoss  alloraMath.Dec
> 	EpsilonTopic         alloraMath.Dec
> 	PNorm                alloraMath.Dec
> 	CNorm                alloraMath.Dec
> }
> 
> func CalcForecastImpliedInferences(args CalcForecastImpliedInferencesArgs) (
> 	forecasterToForecastImpliedInference map[Forecaster]*emissionstypes.Inference,
> 	infererToRegretOut map[Inferer]*Regret,
> 	forecasterToRegretOut map[Forecaster]*Regret,
> 	err error,
> ) {
> ```
> 
> # Option B
> Every parameter gets its own function argument, downside is theoretically you could screw up the order of arguments and have a really nasty bug
> 
> ```go
> func CalcForecastImpliedInferences(
> 	Logger log.Logger,
> 	TopicId uint64,
> 	AllInferersAreNew bool,
> 	Inferers []Inferer,
> 	InfererToInference map[Inferer]*emissionstypes.Inference,
> 	InfererToRegret map[Inferer]*Regret,
> 	Forecasters []Forecaster,
> 	ForecasterToForecast map[Forecaster]*emissionstypes.Forecast,
> 	ForecasterToRegret map[Forecaster]*Regret,
> 	NetworkCombinedLoss alloraMath.Dec,
> 	EpsilonTopic alloraMath.Dec,
> 	PNorm alloraMath.Dec,
> 	CNorm alloraMath.Dec,
> ) (
> 	forecasterToForecastImpliedInference map[Forecaster]*emissionstypes.Inference,
> 	infererToRegretOut map[Inferer]*Regret,
> 	forecasterToRegretOut map[Forecaster]*Regret,
> 	err error,
> ) {
> ```
> 
> # Option C
> Make some small struct that just holds tertiary information and leave the real important stuff as args. Downside is you will certainly either pass unnecessary args or have to make 5 different "context" type structs downside two is that it obscures what the actual function arguments are, you have to go look up the struct definition
> 
> ```go
> type InferenceSynthesisContext struct {
> 	Logger               log.Logger
> 	TopicId              uint64
> 	EpsilonTopic         alloraMath.Dec
> 	PNorm                alloraMath.Dec
> 	CNorm                alloraMath.Dec
> }
> 
> func CalcForecastImpliedInferences(
> 	infSynthCtx InferenceSynthesisContext,
> 	Inferers             []Inferer
> 	InfererToInference   map[Inferer]*emissionstypes.Inference
> 	InfererToRegret      map[Inferer]*Regret
> 	AllInferersAreNew    bool
> 	Forecasters          []Forecaster
> 	ForecasterToForecast map[Forecaster]*emissionstypes.Forecast
> 	ForecasterToRegret   map[Forecaster]*Regret
> 	NetworkCombinedLoss  alloraMath.Dec
> ) (
> 	forecasterToForecastImpliedInference map[Forecaster]*emissionstypes.Inference,
> 	infererToRegretOut map[Inferer]*Regret,
> 	forecasterToRegretOut map[Forecaster]*Regret,
> 	err error,
> ) {
> ```
> 
> # Option D
> Make a big ass Synth Palette struct that holds every possible argument for every possible function. Downside a reader does not know what actual parts of the synth palette are used in the function they are looking at based on the function signature.
> 
> ```go
> type SynthPalette struct {
> 	Ctx               sdk.Context
> 	K                 keeper.Keeper
> 	Logger            log.Logger
> 	TopicId           TopicId
> 	AllInferersAreNew bool
> 	// Should use this as a source of truth regarding for which inferers to have data calculated
> 	// i.e. if an inferer is not present here, calculate a network inference without their data
> 	// Must be unique values
> 	Inferers          []Worker
> 	InferenceByWorker map[Worker]*emissions.Inference
> 	// Must respect the order of sister `inferers` property
> 	InfererRegrets map[Worker]*alloraMath.Dec
> 	// Should use this as a source of truth regarding for which forecasters to have data calculated
> 	// i.e. if an forecaster is not present here, calculate a network inference without their data
> 	// Must be unique values
> 	Forecasters                      []Worker
> 	ForecastByWorker                 map[Worker]*emissions.Forecast
> 	ForecastImpliedInferenceByWorker map[Worker]*emissions.Inference
> 	// Must respect the order of sister `forecasters` property
> 	ForecasterRegrets   map[Worker]*alloraMath.Dec
> 	NetworkCombinedLoss Loss
> 	EpsilonTopic        alloraMath.Dec
> 	EpsilonSafeDiv      alloraMath.Dec
> 	PNorm               alloraMath.Dec
> 	CNorm               alloraMath.Dec
> }
> 
> 
> func CalcForecastImpliedInferences(
> 	palette SynthPalette,
> )
> 	paletteOut SynthPalette,
> 	err error,
> ) {
> ```

I strongly favour explicit dependency definition (A,B,C) especially in logic-charged functions.

I also think option C is unsatisfying, since it does not clarify whether the context-like struct will be totally or partially used. 
Note: I would see another meaningful grouping, iff it makes sense to be used in other areas of the code, e.g. 
`InfererData` to contain
```
        AllInferersAreNew bool
        Inferers             []Inferer
	InfererToInference   map[Inferer]*emissionstypes.Inference
	InfererToRegret      map[Inferer]*Regret
```
(same for `ForecasterData` without the `AllInferersAreNew bool` )
But I would advise for this **only if** this is a meaningful grouping also in other parts of the code, ie if this inferer data parameters usually "travel and are used together". I think it is not the case, but I'm happy to be proved wrong :) 


And between A/B, I have lighter opinion, but ig the downside presented for option B also happens in option A (even if you `exhauststruct`, you may switch order of params, although because it forces to be in-your-face code `key: Value`, the chances are much lower). Also option A allows use of `exhauststruct` validation. So maybe because of that clarity would lean towards A.

guilherme-brandao:
As I said in private chat, I was thinking more like Option C, but with the common args between all funcs in a shared context.

```
type InferenceSynthesis struct {
	Logger               log.Logger
	TopicId               uint64
	Keeper               keeper.Keeper
	EpsilonTopic      alloraMath.Dec
	PNorm                alloraMath.Dec
	CNorm               alloraMath.Dec
}

func (is *InferenceSynthesis) CalcForecastImpliedInferences(
```

and the second best option for me is Option A.

xmariachi:
> As I said in private chat, I was thinking more like Option C, but with the common args between all funcs in a shared common context. The function signature cleaner, just with specific args for the function.
> 
> ```golang
> type InferenceSynthesis struct {
> 	Logger               log.Logger
> 	TopicId              uint64
> 	Keeper               keeper.Keeper
> 	EpsilonTopic         alloraMath.Dec
> 	PNorm                alloraMath.Dec
> 	CNorm                alloraMath.Dec
> }
> 
> func (is *InferenceSynthesis) CalcForecastImpliedInferences(
> ```
> 
> and the second best option for me is Option A.

I think making the keeper an element removes the explicit declaration of dependencies, back to being implicit (what from inside the keeper may be needed, or may be changed, is not clear from the declaration). Making it explicit also helps unit tests.

Principle of least privilege ftw

kpeluso:
We agreed to Option A on our last call. We're wasting cycles by continuing debate. "Disagree but commit"

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -173,186 +178,199 @@ func GetCalcSetNetworkRegrets(
...
 			if err != nil {
 				return errorsmod.Wrapf(err, "Error setting one-out inferer inferer regret")
 			}
 		}
 	}
 
+	sortedOneOutInfererForecasterValues := args.NetworkLosses.OneOutInfererForecasterValues
+	sort.Slice(sortedOneOutInfererForecasterValues, func(i, j int) bool {
+		return sortedOneOutInfererForecasterValues[i].Forecaster < sortedOneOutInfererForecasterValues[j].Forecaster
+	})
```
kpeluso, line 234: is this not already sorted? does it need sorting here?

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -173,186 +178,199 @@ func GetCalcSetNetworkRegrets(
...
 			return errorsmod.Wrapf(err, "Error setting inferer regret")
 		}
 	}
 
-	sort.Slice(networkLosses.OneOutInfererValues, func(i, j int) bool {
-		return networkLosses.OneOutInfererValues[i].Worker < networkLosses.OneOutInfererValues[j].Worker
+	sortedOneOutInfererValues := args.NetworkLosses.OneOutInfererValues
+	sort.Slice(sortedOneOutInfererValues, func(i, j int) bool {
+		return sortedOneOutInfererValues[i].Worker < sortedOneOutInfererValues[j].Worker
 	})
```
kpeluso, line 205: same wrt sorting -- already sorteD?

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -143,27 +147,28 @@ func GetCalcSetNetworkRegrets(
 		}
 	}
 
-	sort.Slice(networkLosses.ForecasterValues, func(i, j int) bool {
-		return networkLosses.ForecasterValues[i].Worker < networkLosses.ForecasterValues[j].Worker
+	sortedForecasterValues := args.NetworkLosses.ForecasterValues
+	sort.Slice(sortedForecasterValues, func(i, j int) bool {
+		return sortedForecasterValues[i].Worker < sortedForecasterValues[j].Worker
 	})
```
kpeluso, line 153: already sorted?

x/emissions/keeper/inference_synthesis/weight.go
```diff
@@ -140,46 +162,50 @@ func (p *SynthPalette) CalcWeightedInference(weights RegretInformedWeights) (Inf
...
-			if _, ok := weights.forecasters[forecaster]; !ok {
-				p.Logger.Debug(fmt.Sprintf("Cannot find forecaster in weights.forecasters in CalcWeightedInference %v", forecaster))
+			forecasterWeight, exists := weights.forecasters[forecaster]
+			if !exists {
+				logger.Debug(fmt.Sprintf("Cannot find forecaster in weights.forecasters in CalcWeightedInference %v", forecaster))
 				continue
 			}
-			if _, ok := p.ForecasterRegrets[forecaster]; !ok {
-				p.Logger.Debug(fmt.Sprintf("Cannot find forecaster in ForecasterRegrets in CalcWeightedInference %v", forecaster))
+			if _, exists := forecasterToRegret[forecaster]; !exists {
```
kpeluso, line 202: I believe resident Go sage @spooktheducks frowns upon this?

x/emissions/keeper/inference_synthesis/weight.go
```diff
@@ -140,46 +162,50 @@ func (p *SynthPalette) CalcWeightedInference(weights RegretInformedWeights) (Inf
...
-			if _, ok := weights.inferers[inferer]; !ok {
-				p.Logger.Debug(fmt.Sprintf("Cannot find inferer in weights.inferers in CalcWeightedInference %v", inferer))
+			infererWeight, exists := weights.inferers[inferer]
+			if !exists {
+				logger.Debug(fmt.Sprintf("Cannot find inferer in weights.inferers in CalcWeightedInference %v", inferer))
 				continue
 			}
-			if _, ok := p.InfererRegrets[inferer]; !ok {
-				p.Logger.Debug(fmt.Sprintf("Cannot find inferer in InfererRegrets in CalcWeightedInference %v", inferer))
+			if _, exists := infererToRegret[inferer]; !exists {
```
kpeluso, line 176: same

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -1,215 +1,352 @@
...
+		args.AllInferersAreNew,
+		remainingInferers,
+		remainingInfererToInference,
+		remainingInfererRegrets,
+		args.Forecasters,
+		remainingForecasterRegrets,
+		forecasterToForecastImpliedInference,
+		weights,
+		args.EpsilonSafeDiv,
+	)
```
kpeluso, line 283: if num args >3 or 4, plz write a struct. That is truly Option A

kpeluso (changes requested):

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -218,189 +355,503 @@ func (b *NetworkInferenceBuilder) SetOneOutInfererValues() *NetworkInferenceBuil
...
-	for _, inferer := range palette.Inferers {
-		regret, _, err := palette.K.GetOneOutForecasterInfererNetworkRegret(palette.Ctx, palette.TopicId, withheldForecaster, inferer)
-		if err != nil {
-			return alloraMath.Dec{}, errorsmod.Wrapf(err, "Error getting one-out inferer regret")
+			forecast, ok := args.ForecasterToForecast[forecaster]
+			if !ok {
+				args.Logger.Debug(fmt.Sprintf("Cannot find forecaster in ForecasterRegrets in UpdateForecastersInfo %v", forecaster))
+				continue
+			}
+			remainingForecasterToForecast[forecaster] = forecast
```
guilherme-brandao, line 408: I think using `continue` here will break stuff in further calculations; please double-check if it's not better to just throw an error.

x/emissions/keeper/inference_synthesis/network_regrets.go
relyt29: This is already sorted in https://github.com/allora-network/allora-chain/blob/f02b25e839ee32eda8f9c422ea44596c40a6ebe7/x/emissions/keeper/inference_synthesis/network_losses.go#L218

and 

https://github.com/allora-network/allora-chain/blob/f02b25e839ee32eda8f9c422ea44596c40a6ebe7/x/emissions/keeper/inference_synthesis/network_losses.go#L239-L249

will fix now

x/emissions/keeper/inference_synthesis/network_regrets.go
relyt29: you are right, sorted in synth.CalcNetworkLosses

x/emissions/keeper/inference_synthesis/network_regrets.go
relyt29: you are right sorted in synth.CalcNetworkLosses

x/emissions/keeper/inference_synthesis/weight.go
relyt29: will fix

x/emissions/keeper/inference_synthesis/weight.go
relyt29: will fix

x/emissions/keeper/inference_synthesis/network_inference_builder.go
relyt29: missed this spot, thought I had gotten everything, will address

x/emissions/keeper/inference_synthesis/network_inference_builder.go
relyt29: that's taken from this line here:

https://github.com/allora-network/allora-chain/blob/fd6486788fb2f37d78ce23281d7296c89586753f/x/emissions/keeper/inference_synthesis/synth_palette_weight.go#L264-L281

so it would be a change from what `dev` does. I am not against throwing an error here instead of continuing, but want to note that

x/emissions/keeper/inference_synthesis/weight.go
```diff
@@ -122,15 +136,26 @@ func (p *SynthPalette) CalcWeightsGivenWorkers() (RegretInformedWeights, error)
...
+	allInferersAreNew bool,
+	inferers []Worker,
+	workerToInference map[Worker]*emissionstypes.Inference,
+	infererToRegret map[Worker]*alloraMath.Dec,
+	forecasters []Worker,
+	forecasterToRegret map[Worker]*alloraMath.Dec,
+	forecasterToForecastImpliedInference map[Worker]*emissionstypes.Inference,
+	weights RegretInformedWeights,
+	epsilonSafeDiv alloraMath.Dec,
+) (InferenceValue, error) {
```
kpeluso, line 150: could consider a lint to catch these. >3 arg again

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -1,215 +1,358 @@
...
-		if inferer != withheldInferer {
+	remainingInfererToInference := make(map[Worker]*emissions.Inference)
+	remainingInfererRegrets := make(map[string]*alloraMath.Dec)
+	for _, inferer := range args.Inferers {
+		// over just the remaining inferers
+		if inferer != args.WithheldInferer {
 			remainingInferers = append(remainingInferers, inferer)
+			inference, ok := args.InfererToInference[inferer]
+			if !ok {
+				args.Logger.Debug(fmt.Sprintf("calcOneOutInfererInference() cannot find inferer in InferenceByWorker in UpdateInferersInfo %v", inferer))
```
kpeluso, line 215: `UpdateInferersInfo` not anymore...

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -218,189 +361,507 @@ func (b *NetworkInferenceBuilder) SetOneOutInfererValues() *NetworkInferenceBuil
...
+		return alloraMath.Dec{}, errorsmod.Wrapf(err, "CalcOneInValue() error getting one-in forecaster regret")
 	}
+	singleForecasterRegret[args.OneInForecaster] = &regret.Value
 
-	palette.ForecasterRegrets[oneInForecaster] = &regret.Value
+	// get self forecast list
+	singleForecaster := []Worker{args.OneInForecaster}
 
-	remainingForecaster := []Worker{oneInForecaster}
-	err = palette.UpdateForecastersInfo(remainingForecaster)
```
kpeluso, line 323: Surprising but welcome compute savings here and in `UpdateInferersInfo()` call below. good stuff

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -361,7 +343,12 @@ func GetCalcSetNetworkRegrets(
 	return nil
 }
 
-func CalcTopicInitialRegret(regrets []alloraMath.Dec, epsilon alloraMath.Dec, pNorm alloraMath.Dec, cNorm alloraMath.Dec) (alloraMath.Dec, error) {
+func CalcTopicInitialRegret(
+	regrets []alloraMath.Dec,
+	epsilon alloraMath.Dec,
+	pNorm alloraMath.Dec,
+	cNorm alloraMath.Dec,
+) (initialRegret alloraMath.Dec, err error) {
```
kpeluso, line 351: pushing it :) but sure

x/emissions/keeper/inference_synthesis/network_inference_builder.go
kpeluso: fixed in my next commit

x/emissions/keeper/inference_synthesis/weight.go
kpeluso: fixed in my next commit

kpeluso (approved):

x/emissions/keeper/inference_synthesis/network_inference_builder.go
relyt29: made the change, I agree it should be an error not skipped

kpeluso (approved):

guilherme-brandao (approved):

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -1,215 +1,358 @@
...
-	palette := b.palette.Clone()
-	totalInferers := palette.Inferers
+func calcOneOutInfererInference(args CalcOneOutInfererInferenceArgs) (
+	oneOutNetworkInferenceWithoutInferer alloraMath.Dec,
+	err error,
+) {
+	args.Logger.Debug(fmt.Sprintf(
+		"calcOneOutInfererInference() calculating one-out inference for topic %v withheld inferer %s", args.TopicId, args.WithheldInferer))
 
 	// Remove the inferer from the palette's inferers
```
xmariachi, line 205: comment: no references to palette. There are several around the file.

x/emissions/keeper/inference_synthesis/network_inferences.go
```diff
@@ -7,32 +7,26 @@ import (
...
-	int64,
-	int64,
-	error,
+	networkInferences *emissions.ValueBundle,
+	forecasterToForecastImpliedInference map[string]*emissions.Inference,
+	infererToWeight map[Inferer]Weight,
+	forecasterToWeight map[Forecaster]Weight,
+	inferenceBlockHeight int64,
+	lossBlockHeight int64,
+	err error,
```
xmariachi, line 28: nice naming of the output components

xmariachi (approved):
lgtm, great work

guilherme-brandao (approved):

xmariachi (approved):

PR #542, opened 2024-08-23
WIP Active topic management

RedBird96:

Currently active topics are in one large keyset that must be parsed in pages in order to find topics.
This is not scalable as the number of topics increases.
Furthermore, there are some minor issues with respect to how topic weight is calculated -- and EMA is often used where the non-EMA may be more appropriate.

This PR attempts to address these concerns by segmenting the list of active topics by block height of the next epoch.
So, there is no longer 1 keyset for all active topics.
Instead, we map block height to a bounded list of topics.
When a topic is (re)activated, we append it to this list but knock off the topic of lowest weight to maintain a bounded set size.
We also introduce a cache for topic weight and lowest topic weight per "active set for a block" to minimize compute.

https://linear.app/upshot/issue/ORA-2068/patch-active-topic-parsing

## Still Left Todo


x/emissions/keeper/keeper_test.go
```diff
@@ -2280,6 +2296,72 @@ func (s *KeeperTestSuite) TestGetActiveTopicsWithSmallLimitAndOffset() {
...
+	_ = setTopicWeight(topic2.Id, 20, 10)
+	_ = keeper.SetTopic(ctx, topic2.Id, topic2)
+	_ = keeper.ActivateTopic(ctx, topic2.Id)
+
+	// Fetch next page -- should only return topic 5
+	pagination := &types.SimpleCursorPaginationRequest{
+		Key:   nil,
+		Limit: 2,
+	}
+	activeTopics, _, err := keeper.GetIdsOfActiveTopics(ctx, pagination)
```
kpeluso, line 2337: Just for testing is one thing, but in general we shouldn't need `GetIdsOfActiveTopics()` anymore. Can update its existing query method on the autocli to be "get ids of active topics at block" too. Ideally we don't even need it in the tests

x/emissions/keeper/topic_activation.go
```diff
@@ -0,0 +1,307 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Inactivate the topic
+func (k *Keeper) InactivateTopic(ctx context.Context, topicId TopicId) error {
+	_, err := k.inactivateTopicWithoutMinWeightReset(ctx, topicId)
```
kpeluso, line 275: need to do the min weight reset if this function is still used, or declare the encapsulating function with a more appropriate name

x/emissions/keeper/topic_activation.go
```diff
@@ -0,0 +1,307 @@
...
+	return err
+}
+
+// If the topic weight is not less than lowest weight keep it as activated
+func (k *Keeper) UpdateActiveTopic(ctx context.Context, topicId TopicId) error {
```
kpeluso, line 280: The function name could be way clearer. Maybe "AttemptTopicReactivation"

kpeluso (commented):
Still looking for migration handler, its tests, and integration tests (esp what happens when there are more topics than there is space for at a block).

x/emissions/keeper/topic_activation.go
```diff
@@ -0,0 +1,346 @@
...
+	}
+	currentBlock := sdk.UnwrapSDKContext(ctx).BlockHeight()
+	epochEndBlock := currentBlock + topic.EpochLength
+
+	// Remove current active topic from block
+	err = k.removeCurrentTopicFromBlock(ctx, topicId, currentBlock)
+	if err != nil {
+		sdkCtx.Logger().Warn(fmt.Sprintf("Failed to remove current active topic from block %d", topicId))
+		return err
+	}
```
guilherme-brandao, line 294: Considering the name of the function `AttemptTopicReactivation`,  I can't understand  why we need to do this. Can you explain please?

x/emissions/keeper/topic_activation.go
```diff
@@ -0,0 +1,346 @@
...
+}
+
+// Inactivate the topic
+func (k *Keeper) InactivateTopic(ctx context.Context, topicId TopicId) error {
+	_, err := k.inactivateTopicWithoutMinWeightReset(ctx, topicId)
+	return err
+}
+
+// If the topic weight is not less than lowest weight keep it as activated
+func (k *Keeper) AttemptTopicReactivation(ctx context.Context, topicId TopicId) error {
```
guilherme-brandao, line 280: most of this function is the same of the `ActivateTopic`, prob it can be refactored

x/emissions/keeper/keeper_test.go
```diff
@@ -2192,10 +2203,17 @@ func (s *KeeperTestSuite) TestGetActiveTopics() {
...
+	s.Require().Len(activeTopics, 1, "Should retrieve exactly two active topics")
 
-	s.Require().Len(activeTopics, 2, "Should retrieve exactly two active topics")
+	pagination = &types.SimpleCursorPaginationRequest{
+		Key:   nil,
+		Limit: 10,
+	}
+	activeTopics, _, err = keeper.GetIdsActiveTopicAtBlock(ctx, 15, pagination)
+	s.Require().NoError(err, "Fetching active topics should not produce an error")
+	s.Require().Len(activeTopics, 1, "Should retrieve exactly two active topics")
```
kpeluso, line 2216: In both cases the error message is somewhat misleading `"Should retrieve exactly two active topics"` because you are returning exactly 1 topic per block, but I get what you mean after staring at it -- there are 2 active topics overall

x/emissions/migrations/v3/migrate.go
```diff
@@ -86,11 +93,184 @@ func MigrateParams(store storetypes.KVStore, cdc codec.BinaryCodec) error {
...
+		MaxElementsPerForecast:              defaultParams.MaxElementsPerForecast,
+		MaxActiveTopicsPerBlock:             defaultParams.MaxActiveTopicsPerBlock,
 	}
 
 	store.Delete(types.ParamsKey)
 	store.Set(types.ParamsKey, cdc.MustMarshal(&newParams))
 	return nil
 }
+
+func MigrateActiveTopics(store storetypes.KVStore, ctx sdk.Context, emissionsKeeper keeper.Keeper) error {
```
kpeluso, line 105: I may be mistaken but technically this wasn't necessary because after the migration, as we talked about, we could simply add even the slightest funding to each of the topics, right? Then the rest of your code would just make it all work. Given that this migration handler is all already here though, we'll keep it -- you've saved us a few steps.

x/emissions/module/rewards/topic_rewards.go
```diff
@@ -133,7 +85,8 @@ func PickChurnableActiveTopics(
 		// Loop over and run epochs on topics whose inferences are demanded enough to be served
 		// Check the cadence of inferences, and just in case also check multiples of epoch lengths
 		// to avoid potential situations where the block is missed
-		if k.CheckWorkerOpenCadence(block, topic) {
+		//if k.CheckWorkerOpenCadence(block, topic)
```
kpeluso, line 88: Think this is right. Just should specify why in the function declaration -- because the `block` arg is used to get active topics as of that block

x/emissions/module/rewards/topic_rewards.go
```diff
@@ -188,7 +141,8 @@ func PickChurnableActiveTopics(
 			}
 		}
 		// Check Reputer Close Cadence
-		if k.CheckReputerCloseCadence(block, topic) {
+		//if k.CheckReputerCloseCadence(block, topic)
```
kpeluso, line 144: This I think we still want to keep because there is a `ground_truth_lag` period of time in which there are worker nonces but no reputer nonces.

kpeluso (commented):
Will give final review tm.

x/emissions/migrations/v3/migrate.go
```diff
@@ -86,11 +93,184 @@ func MigrateParams(store storetypes.KVStore, cdc codec.BinaryCodec) error {
...
+		lowestWeightBytes, err := cuLowestWeight.Marshal()
+		if err != nil {
+			continue
+		}
+		blockToActiveStore.Set(blockHeightBytes, activeTopicsBytes)
+		blockLowestWeightStore.Set(blockHeightBytes, lowestWeightBytes)
+	}
+	_ = iterator.Close()
+	return nil
+}
```
guilherme-brandao, line 221: Any error in this function should panic; if an error occurs in the migration and we just `continue`, we will end up in an incorrect state.

x/emissions/migrations/v3/migrate_test.go
```diff
@@ -167,5 +170,107 @@ func (s *MigrationTestSuite) TestMigrate() {
...
+	for ; iterator.Valid(); iterator.Next() {
+		var msg types.TopicIds
+		err := proto.Unmarshal(iterator.Value(), &msg)
+		if err != nil {
+			continue
+		}
+		if len(msg.TopicIds) == 0 {
+			continue
+		}
+		s.Require().Len(msg.TopicIds, 3)
```
guilherme-brandao, line 231: 1. Should check for errors: `s.Require().NoError(err)`
2.This `len(msg.TopicIds) == 0` validation is causing a false positive; when you comment it out, the test fails.

x/emissions/module/rewards/topic_rewards.go
```diff
@@ -280,9 +240,17 @@ func GetAndUpdateActiveTopicWeights(
...
+				return nil, alloraMath.Dec{}, cosmosMath.Int{}, errors.Wrapf(err, "failed to inactivate topic")
+			}
+			continue
 		}
 
+		// Update topic active status
+		err = k.AttemptTopicReactivation(ctx, topicId)
+		if err != nil {
+			continue
+		}
```
guilherme-brandao, line 253: I think we should add a log error here, at least in the first and the last continue.

x/emissions/migrations/v3/migrate.go
```diff
@@ -86,11 +93,184 @@ func MigrateParams(store storetypes.KVStore, cdc codec.BinaryCodec) error {
...
+		err = stake.Unmarshal(topicStakeStore.Get(idArray))
+		if err != nil {
+			continue
+		}
+		var previousWeight = alloraMath.NewDecFromInt64(0)
+		err = previousWeight.Unmarshal(topicPreviousWeightStore.Get(idArray))
+		if err != nil {
+			continue
+		}
+		// Get topic's latest weight
```
guilherme-brandao, line 149: Nbd, but you can improve the code above: instead of doing `NewStore` then `Unmarshal`, you can just use the emissionsKeeper, e.g., `GetPreviousTopicWeight`. Same thing for setting, e.g: `SetPreviousTopicWeight`

x/emissions/migrations/v3/migrate_test.go
```diff
@@ -167,5 +170,107 @@ func (s *MigrationTestSuite) TestMigrate() {
...
+	iterator := blockToActiveStore.Iterator(nil, nil)
+	for ; iterator.Valid(); iterator.Next() {
+		var msg types.TopicIds
+		err := proto.Unmarshal(iterator.Value(), &msg)
+		if err != nil {
+			continue
+		}
+		s.Require().GreaterOrEqual(len(msg.TopicIds), 3)
+	}
+}
```
guilherme-brandao, line 202: 1. Instead of getting the store, you can just use `emissionsKeeper`, e.g: `GetActiveTopicIdsAtBlock`
2. Check for error instead of just `continue`

x/emissions/module/rewards/topic_rewards.go
xmariachi: You're right for reproducing the previous behaviour @kpeluso 
However opened a convo ab this separately on the desired behaviour here, so we can clearly reproduce it in docs with 100% certainty on intent.

x/emissions/module/rewards/topic_rewards.go
xmariachi, line 88: Some `Check*Cadence`  are no longer in use - if so, remove those.

x/emissions/module/rewards/topic_rewards.go
```diff
@@ -130,90 +83,85 @@ func PickChurnableActiveTopics(
...
-			}
-			// Add Worker Nonces
-			nextNonce := types.Nonce{BlockHeight: block}
-			err = k.AddWorkerNonce(ctx, topic.Id, &nextNonce)
-			if err != nil {
-				ctx.Logger().Warn(fmt.Sprintf("Error adding worker nonce: %s", err.Error()))
-				continue
-			}
-			ctx.Logger().Debug(fmt.Sprintf("Added worker nonce for topic %d: %v \n", topic.Id, nextNonce.BlockHeight))
+		ctx.Logger().Debug(fmt.Sprintf("ABCI EndBlocker: Worker open cadence met for topic: %v metadata: %s . \n",
```
xmariachi, line 87: This reference to ABCI Endblocker is no longer valid since this code has been moved to topic_rewards

x/emissions/module/rewards/topic_rewards.go
```diff
@@ -130,90 +83,85 @@ func PickChurnableActiveTopics(
...
-				continue
-			}
+		// Update the last inference ran
+		err = k.UpdateTopicEpochLastEnded(ctx, topic.Id, block)
+		if err != nil {
+			ctx.Logger().Warn(fmt.Sprintf("Error updating last inference ran: %s", err.Error()))
+			continue
+		}
+		// Add Worker Nonces
+		nextNonce := types.Nonce{BlockHeight: block}
```
xmariachi, line 98: I wonder whether topic_rewards.go is the right place to keep all of these logic that refers to topic and nonce management. One would think it is only about rewards.

x/emissions/module/rewards/topic_rewards.go
```diff
@@ -280,9 +234,18 @@ func GetAndUpdateActiveTopicWeights(
...
 			err := k.InactivateTopic(ctx, topic.Id)
-			return nil, alloraMath.Dec{}, cosmosMath.Int{}, errors.Wrapf(err, "failed to inactivate topic")
+			if err != nil {
+				return nil, alloraMath.Dec{}, cosmosMath.Int{}, errors.Wrapf(err, "failed to inactivate topic")
+			}
+			continue
 		}
 
+		// Update topic active status
+		err = k.AttemptTopicReactivation(ctx, topicId)
```
xmariachi, line 244: Should this be in an "else" to the inactivation above?

x/emissions/proto/emissions/v3/params.proto
```diff
@@ -143,4 +142,5 @@ message Params {
   ];
   // payload sending fee for reputer or worker
   uint64 max_elements_per_forecast = 45; // max number of top forecasters by score to reward
+  uint64 max_active_topics_per_block = 46;
```
relyt29, line 145: this should have a comment describing what it does / is

x/emissions/types/params.go
```diff
@@ -50,6 +49,7 @@ func DefaultParams() Params {
 		HalfMaxProcessStakeRemovalsEndBlock: uint64(40),                                   // half of the max number of stake removals to process at the end of the block, set this too big and blocks require too much time to process, slowing down consensus
 		DataSendingFee:                      cosmosMath.NewInt(10),                        // how much workers and reputers must pay to send payload
 		MaxElementsPerForecast:              uint64(12),                                   // top forecast elements by score
+		MaxActiveTopicsPerBlock:             uint64(1),                                    // maximum number of active topics per block
```
relyt29, line 52: we're going with a default of 1? is that advisable? only one topic that's actually being handled for each block?

x/emissions/keeper/keeper.go
```diff
@@ -1995,6 +1962,58 @@ func (k Keeper) GetIdsOfActiveTopics(ctx context.Context, pagination *types.Simp
 	}, nil
 }
 
+func (k Keeper) GetIdsActiveTopicAtBlock(ctx context.Context, blockHeight int64, pagination *types.SimpleCursorPaginationRequest) ([]TopicId, *types.SimpleCursorPaginationResponse, error) {
```
relyt29, line 1965: would love a comment describing purpose of this function here

x/emissions/keeper/msgserver/msg_server_params_test.go
```diff
@@ -8,43 +8,6 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-func (s *MsgServerTestSuite) TestUpdateParams() {
```
relyt29, line 11: is there a reason we delete this test? or was it moved to another file

x/emissions/keeper/queryserver/query_server_topics.go
```diff
@@ -177,3 +177,34 @@ func (qs queryServer) GetRewardableTopics(
 
 	return &types.QueryRewardableTopicsResponse{RewardableTopicIds: rewardableTopics}, nil
 }
+
+func (qs queryServer) GetActiveTopicsAtBlock(
```
relyt29, line 181: so you return a pagination cursor right, but how is the user supposed to query for another page of data? I don't see anything that takes in existing page cursors?

x/emissions/keeper/msgserver/msg_server_params_test.go
relyt29: but the title of the test is testUpdateParams, we should keep at least one test that checks that we have the ability to update params in general, even if we deleted the specific param referenced, can change which param is updated in the function

x/emissions/migrations/v3/migrate.go
```diff
@@ -50,7 +60,6 @@ func MigrateParams(store storetypes.KVStore, cdc codec.BinaryCodec) error {
 		Version:                             oldParams.Version,
 		MaxSerializedMsgLength:              oldParams.MaxSerializedMsgLength,
 		MinTopicWeight:                      oldParams.MinTopicWeight,
-		MaxTopicsPerBlock:                   oldParams.MaxTopicsPerBlock,
```
relyt29, line 53: this is missing a comment in the comment header below stating that MaxTopicsPerBlock was deleted

relyt29 (changes requested):

x/emissions/migrations/v3/migrate.go
relyt29: nevermind I'll just fix it in my PR

relyt29 (approved):

x/emissions/migrations/v3/migrate.go
kpeluso: fixed

x/emissions/keeper/keeper.go
```diff
@@ -1995,6 +1963,59 @@ func (k Keeper) GetIdsOfActiveTopics(ctx context.Context, pagination *types.Simp
 	}, nil
 }
 
+// Returns active topic ids at block using pagination
+func (k Keeper) GetIdsActiveTopicAtBlock(ctx context.Context, blockHeight int64, pagination *types.SimpleCursorPaginationRequest) ([]TopicId, *types.SimpleCursorPaginationResponse, error) {
```
kpeluso, line 1967: do we really need pagination here?

kpeluso: resolved in https://github.com/allora-network/allora-chain/pull/557 + removed GetActiveTopics

kpeluso (approved):

kpeluso:
Approved changes I have coming in: https://github.com/allora-network/allora-chain/pull/557

kpeluso:
Will also do changelog in a batch to add even more color to all recent PRs including this one

PR #543, opened 2024-08-23
Guilherme/proto 2239 fix bugs with payload validation and weight loss normalization

guilherme-brandao:
- Added missing validation for `Dec` to avoid `Infinite` numbers
- Fixed validation in weight loss normalization
- Remove unnecessary `Epsilon` fallback in network loss calculation
- Use `EpsilonSafeDiv`instead of `topic.Epsilon` for proximity to consensus (Eq32) division protection


x/emissions/types/msg_create_topic.go
```diff
@@ -28,13 +28,13 @@ func (msg *MsgCreateNewTopic) Validate() error {
 	if msg.WorkerSubmissionWindow > msg.EpochLength {
 		return errors.Wrap(sdkerrors.ErrInvalidRequest, "worker submission window cannot be higher than epoch length")
 	}
-	if msg.AlphaRegret.Lte(alloraMath.ZeroDec()) || msg.AlphaRegret.Gt(alloraMath.OneDec()) {
+	if msg.AlphaRegret.Lte(alloraMath.ZeroDec()) || msg.AlphaRegret.Gt(alloraMath.OneDec()) || msg.AlphaRegret.IsNaN() || !msg.AlphaRegret.IsFinite() {
```
kpeluso, line 31: why not use the validate method here as well?

x/emissions/types/validations.go
```diff
@@ -0,0 +1,19 @@
+package types
```
kpeluso, line 1: would have specifically called this file `dec .go` to be in line with others in this dir but nbd

kpeluso (approved):

x/emissions/types/validations.go
guilherme-brandao: the idea is to add more common validations here in the future

relyt29 (approved):

PR #545, opened 2024-08-26
Fix duplicate HTTP endpoints

fernandofcampos:

Some RPC methods share the same HTTP endpoint. This will cause a conflict because the HTTP server won't know which RPC to call when this endpoint is hit. We should have unique endpoints for each RPC method.

https://linear.app/alloralabs/issue/PROTO-2241/some-http-endpoints-at-querygo-are-not-unique

Yes, using existing tests.

## Still Left Todo


kpeluso (approved):

relyt29 (approved):
:ship: it

PR #550, opened 2024-08-27
fix reputer window + strict reputer window tests + lint

xmariachi:

* Fix - close reputer window on the upper part
* Add 4 tests to strictly define reputer window
* Add basic validation on appending reputer losses

PROTO-2246

- [X] If tested, please describe how. If not, why tests are not needed. -- Adds new tests
- [X] If documented, please describe where. If not, describe why docs are not needed. -- not needed: no new behaviour added, just fix existing one

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ Implements merit based sortition for top N inferers, forecasters, and reputers. 
 ### Fixed
 
 * [#544](https://github.com/allora-network/allora-chain/pull/544) Added check against zero-rewards after conversion to cosmosInt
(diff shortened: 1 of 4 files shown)


relyt29 (approved):
Strictly closing the reputer window maybe not necessary but seems like a common sense good thing to do

xmariachi:
> Strictly closing the reputer window maybe not necessary but seems like a common sense good thing to do

After the window is closed (and so the reputer nonce is closed), any remaining input for that nonce would not be used and would just add confusion to have them, potentially making others think (through chain queries / indexer) that this was inserted on time and thus should have received rewards.

PR #551, opened 2024-08-27
Implementation of a linter to list range loops over a map

fernandofcampos:

Implementing  a linter to list range loops over a map, so we can identify those cases more easily.

https://linear.app/alloralabs/issue/PROTO-2144/linter-to-throw-if-iterating-over-a-map

Yes. New tests were added to validate the linter. The new linter is not enabled on golangci.yml, so it won't interfere with the current lint checks. 

## Still Left Todo


linter/.golangci-maprange.yml
```diff
@@ -0,0 +1,25 @@
...
+
+issues:      
+  max-issues-per-linter: 10000
+  max-same-issues: 10000
+
+  exclude-files:
+    - ".*\\.pb\\.go"
+    - ".*\\.pb\\.gw\\.go"
+    - ".*\\.pulsar\\.go"
+    - ".*_mocks\\.go"
```
kpeluso, line 25: should you exclude the linter dir itself so you can keep your test?

fernandofcampos: It is not reporting anything on my tests, which is strange. Investigating why, golangci automatically excludes these dirs:
vendor$, third_party$, testdata$, examples$, Godeps$, builtin$

kpeluso (approved):

PR #562, opened 2024-08-30
Hot fix:  Ensuring worker presence between reputer reports

guilherme-brandao:
One of the properties of the reputer bundle was not being considered in this function, causing a chain halt!


xmariachi (approved):
lgtm

kpeluso (approved):

PR #564, opened 2024-08-30
Quo vs Mul to apply dripping

xmariachi:

Fixes hyper aggressive topic dripping
Adds a test, fixes another test that was not setting the `BlocksPerEpoch` on the params, getting a div by zero.

- [X] If tested, please describe how. If not, why tests are not needed. -- tested locally for now, in line with design
- [X] If documented, please describe where. If not, describe why docs are not needed. -- in line with design now
- [X] Added to `Unreleased` section of `CHANGELOG.md`? -- no need since it is a fix on new feature that already is in the Changelog.

### DIFF
--- a/x/emissions/keeper/keeper.go
+++ b/x/emissions/keeper/keeper.go
@@ -2048,6 +2048,10 @@ func (k *Keeper) DripTopicFeeRevenue(ctx sdk.Context, topicId TopicId, block Blo
 		return err
 	}
 	blocksPerEpoch := alloraMath.NewDecFromInt64(topic.EpochLength)
+	if blocksPerEpoch.IsZero() {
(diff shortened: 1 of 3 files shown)


kpeluso (approved):

relyt29:
The formula is written `ΔC_{t,i} = N_{epoch,w}^{-1} * C_{t,i}` where ΔC_{t,i} is this "change in fee revenue" aka "Drip," C_{t,i} is the topic fee revenue, N_{epoch,w} is the number of epochs per week. 


This PR is correct, I mistakenly wrote Mul instead of Quo, because the formula in the whitepaper is written "to the negative one power multiplied by" instead of "x divided by"

relyt29 (approved):

PR #567, opened 2024-08-31
only close worker nonces after topic window submission number of blocks

kpeluso:

Issue originally identified by @RedBird96 

Our devnet revealed the following suspicious coincidences in time of worker nonce creation and fulfillment:

image

and:
image

This was in addition to a glut logs on our workers that showed many attempts to commit data yet minimal success.
Additionally, many community members have been reporting inability to participate in testnet for quite some time.

It appears our worker nonces are sometimes being closed immediately after they're opened.

I don't understand why this is not universally the case, but my reading of the code is congruent with the observation that this is overwhelmingly the case, making it impossible to join a topic epoch unless you happen to time yourself exactly when the worker nonce is opened/closed.

## After Merge

Must also merge to `dev` branch.

## Still Left Todo

### DIFF
--- a/x/emissions/module/abci.go
+++ b/x/emissions/module/abci.go
@@ -82,7 +82,7 @@ func EndBlocker(ctx context.Context, am AppModule) error {
 				}
 				for _, nonce := range nonces.Nonces {
 					// If the nonce plus the worker submission window is greater than the current block height, then the window is still open => skip
-					if nonce.BlockHeight+topic.WorkerSubmissionWindow > blockHeight {
+					if am.keeper.BlockWithinWorkerSubmissionWindowOfNonce(topic, *nonce, blockHeight) {
 						sdkCtx.Logger().Debug(fmt.Sprintf("ABCI EndBlocker %d: Worker window still open for topic: %d, nonce: %v", blockHeight, topicId, nonce))
 						continue
 					}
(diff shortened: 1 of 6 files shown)


x/emissions/module/abci.go
```diff
@@ -75,7 +75,17 @@ func EndBlocker(ctx context.Context, am AppModule) error {
...
 				continue
 			} else {
+				topic, err := am.keeper.GetTopic(sdkCtx, topicId)
+				if err != nil {
+					sdkCtx.Logger().Warn(fmt.Sprintf("Error getting topic: %s", err.Error()))
+					continue
+				}
 				for _, nonce := range nonces.Nonces {
+					// If the nonce plus the worker submission window is greater than the current block height, then the window is still open => skip
+					if nonce.BlockHeight+topic.WorkerSubmissionWindow > blockHeight {
```
xmariachi, line 85: Great catch. 

Actually the function that was done to check that cadence `keeper.CheckWorkerCloseCadence` is not being used now. 
[code](https://github.com/allora-network/allora-chain/blob/334fbad25389f1a33553f7b11f72e907ed79bf87/x/emissions/keeper/keeper.go#L2668)

I find to have a function with a name like this signals the intention in a clearer way and helps readability.

Note: its companions `CheckReputerCloseCadence` , `CheckWorkerOpenCadence` are not used either now. 
If they are not used then they can be dropped.

xmariachi (approved):
The patch seems to detect and fix the behaviour. 
Preemptively approvd bc of that, though check comments ab code cleaning.

x/emissions/module/abci.go
kpeluso: cleaned in both places mentioned

PR #548, opened 2024-08-26
Merit Based Sortition / Permeability of Inferer, Forecaster, Reputer Active Set

relyt29:
## Purpose of Changes and their Description

Due to the scaling constraints of running a blockchain, we can't process all of the inferences, forecasts, and reputation bundles given to us for rewards.

At the time of payload acceptance (both worker and reputer), we only accept up to `MaxTopXToReward` inferences, forecasts, and reputations to process and store on chain. We use a skimming the top algorithm: we accept everything up until we hit the max amount permissible, and then any subsequent payloads replace the payload of the actor with the lowest Exponential Moving Average (EMA) score in the existing set we have already accepted. We do not filter at this point who is allowed to submit payloads other than only "top N by score" accepted.

We also avoid caculating EMA scores for every actor. Because the calculation of scores requires complete knowledge of the network loss for the epoch, EMA score calculation happens at time of reputer nonce closure (in endblock), upon the final set of payloads submitted at that point. 

We order EMA scores into "active" and "passive" sets of inferers, forecasters, and reputers using score EMAs and quantiles.  `FindTopNByScoreDesc` sorts the set of actors based on their scores (roughly, their distance from the combined network loss) and only keeps the `moduleParams.MaxTopXToReward` actors (where `X` is one of inferers, forecasters, or reputers). In order to smooth out the effect of any particular epoch in the scores of actors, we use Exponential Moving Averages of scores with a new alpha parameter `moduleParams.MeritSortitionAlpha` - this replaces the latest scores data structures in the keeper, so `latestXScoresByX` becomes `xScoreEma` in the keeper.

In order to allow for permeability in the active set, i.e. to allow new players without established reputation to get a chance to upload payloads, we take the `topic.ActiveXQuantile` quantile of all the scores (e.g. the 20th percentile), and assign everyone in the passive set the quantile score as their score. This introduces churn in the bottom quantile of the active set, allowing players without established reputation to break into the active set and submit payloads if they are decently good.

Bullet point list of changes follows:

* Added new fields to the Topic struct for active quantiles: topics now include `active_inferer/forecaster/reputer_quantile` parameter between 0 and 1 for permeability.
* Introduced a merit sortition alpha parameter for weighting previous score EMAs:
* Refactored the FindTopNByScoreDesc function to not rely on a heap data structure
* Moved score calculation logic into the actor_utils package and updated the rewards system to get score EMAs from the keeper at time of rewards, rather than calculate scores at time of rewards.
[...]

PR #568, opened 2024-08-31
add online, weak merit based sortition without new scores yet

kpeluso:
### Why "Weak"?

@jmdkastro formulated a well-defined theory of merit-based sortition plus simulation. This rhymes with that, but is not exactly that.

A consequence of https://github.com/allora-network/allora-chain/pull/567 will be way greater stagnation in the active set, as those with scores have an actual opportunity to make it into their topic's epochs. Hence, there is a need to quickly add a non-corrosive form of permeability to the network. This route maximizes our ability to serve large cohorts of workers while enabling for more sophisticated methods later, potentially as fast-follows.

#### Definition

* **Active set** = Your inference/forecast/reputer loss payload is considered as part of rewards/inference synthesis calculations
* **Passive set** = Your inference/forecast/reputer loss payload is not considered as part of rewards/inference synthesis calculations

We differentiate between these 2 sets because calculations for the active set are expensive => this must be bounded. However, we must cycle through participants, so we now need a name for those "passively considered" actors.

#### When scores are calculated in rewards

* An EMA score is updated for each actor of each type
* A topic-defined quantile of the active set is taken for each actor type (inferer, forecaster, reputer)
   * This will be used in the next epoch
   * For example if EMA score among active set is [10,20,30] then topic quantile might be 12 (linear interpolation)

#### In the next epoch

* We skim the top in an online fashion i.e.
   * If we have not yet met our limit per actor type per epoch, then we add the actor payload to the active set
   * If we have met our limit per actor, then we ask if the EMA score of the new actor is better than the lowest in the current active set.
      * If yes, then we pop the lowest scorer by EMA score out of the active set and update their EMA using the topic quantile of the previous epoch
      * If no, then we update the EMA score of the rejected actor using the topic quantile of the previous epoch

In summary, if you're in the active set, your EMA score gets updated based on your performance in this round, and if you're int he passive set, your EMA score gets updated based on the performance of the previous epoch's active set. In either case, we do calculations in an online fashion => the true limit of actors for which we can do calculations is the cumulative blockspace within a worker submission window.

## Still Left Todo

* Fix tests
* Add more coverage


x/emissions/keeper/keeper.go
```diff
@@ -826,6 +835,16 @@ func (k *Keeper) AppendInference(ctx context.Context, topicId TopicId, nonce typ
...
+	} else {
+		// Update EMA score using previous active set score at topic-defined quantile among the last active set of inferers
+		topic, err := k.GetTopic(ctx, topicId)
+		if err != nil {
+			return err
+		}
+		err = k.CalcAndSaveInfererScoreEmaWithLastSavedTopicQuantile(ctx, topic, block, inference.Inferer)
+		if err != nil {
+			return err
+		}
```
guilherme-brandao, line 847: We are not saving the inference of this inferrer anywhere; is that correct?

Additionally, I realized that in the `AppendInference` function, we are only saving the inferences in the KV store `allInferences` and not in `inferences`, which can affect some queries that use it.

I think we should use the `InsertInferences` method of the keeper.

x/emissions/module/rewards/scores.go
```diff
@@ -108,33 +109,25 @@ func GenerateReputerScores(
...
-		}
-		emaScore := types.Score{
-			TopicId:     topicId,
-			BlockHeight: block,
-			Address:     reputer,
-			Score:       emaScoreDec,
-		}
-		err = keeper.SetReputerScoreEma(ctx, topicId, reputer, emaScore)
+
+		err = keeper.CalcAndSaveReputerScoreEmaIfNewUpdate(ctx, topic, block, reputer, newScore)
```
guilherme-brandao, line 113: Here, it’s calculating and saving the scores' EMA, and right above, it’s inserting the "normal" scores into the `reputerScoresByBlock` KV store. Apparently, this KV store it’s not being used by anything apart from the query server (which I think should use the EMA kv store instead). Need to double-check if this  it’s still needed.

x/emissions/keeper/keeper.go
```diff
@@ -826,6 +835,16 @@ func (k *Keeper) AppendInference(ctx context.Context, topicId TopicId, nonce typ
 		newInferences.Inferences = append(newInferences.Inferences[:lowScoreIndex], newInferences.Inferences[lowScoreIndex+1:]...)
 		newInferences.Inferences = append(newInferences.Inferences, inference)
 		return k.allInferences.Set(ctx, key, newInferences)
+	} else {
+		// Update EMA score using previous active set score at topic-defined quantile among the last active set of inferers
+		topic, err := k.GetTopic(ctx, topicId)
+		if err != nil {
+			return err
+		}
+		err = k.CalcAndSaveInfererScoreEmaWithLastSavedTopicQuantile(ctx, topic, block, inference.Inferer)
```
xmariachi, line 844: if this is the N-th time the inferer is sending inference in same block (yes, paying for it), seems like this will this apply EMA over and over with the quantile on this same block. Seems like this should happen only once per block to avoid people paying for cleaning its own EMA.

x/emissions/module/rewards/scores.go
guilherme-brandao: Also the quantile is not being calculated using the EMA score, but the "normal"score

x/emissions/keeper/keeper.go
```diff
@@ -989,6 +1019,16 @@ func (k *Keeper) AppendReputerLoss(ctx context.Context, topicId TopicId, block B
 			newReputerLossBundles.ReputerValueBundles[lowScoreIndex+1:]...)
 		newReputerLossBundles.ReputerValueBundles = append(newReputerLossBundles.ReputerValueBundles, reputerLoss)
 		return k.allLossBundles.Set(ctx, key, newReputerLossBundles)
+	} else {
+		// Update EMA score using previous active set score at topic-defined quantile among the last active set of inferers
```
relyt29, line 1023: reputers not inferers

relyt29:
so this architecture design, the only way for a new participant to break in is to be first / early?

if they don't have a good score and they try to upload an inference, their TX is rejected/fails and nothing happens 

But if they're early at least they land into the active set and then are later booted out of it and given a quantile

when we spoke on thursday I had pitched an idea of allowing passive actors to upload and get their score set to the quantile even if they're not getting picked in the active set, but the issue was the abuse of the quantile multiple times, so you only allow someone to do that once per epoch

x/emissions/module/rewards/scores.go
kpeluso: fixed the quantile calculation

x/emissions/keeper/keeper.go
kpeluso: fixed both

kpeluso:
@relyt29 this implements the online, unbounded approach where we check if they submitted already in this epoch before including them again. We use the quantile from the previous epoch's active set to update everyone in the passive set. This does not favor those who came first -- anyone below the topic quantile by EMA score is liable to be sub'd out for someone with better EMA score if they present themselves. Their score will be updated using previous epoch's topic quantile, which is by definition better than an earlier active participant who scored below that quantile.

relyt29:
To be clear, this is the attack i am concerned about (we should write a unit test and confirm we prevent this case):

it is now epoch 5. The quantile score from epoch 4 is 75.

Alice has a current score of 50. TopN has been filled at this point in the epoch, and the cutoff score is say, 67. 

Alice submits an inference. Her score is lower than 67, so she's not allowed into the active set. The chain takes her score 50, and computes an EMA with 75 the quantile from last epoch. `75*0.9 + 0.1*50` = 72.5 which is now stored in the chain as alice's score

Alice, in the same epoch, submits her inference again. Her score is now above 67 so she's let into the active set.

OR, depending on the numbers maybe she can't make it into the active set, but she can force the algorithm to recalculate her EMA again with 75. She can do this as many times as she wants until her score is infinitesimally close to 75.


To be clear I am not alleging that we are susceptible to this attack, just that we need to adequately test for it. I do see we check if the inference is already accepted, but I don't see that the negative case-where Alice is not in the active set, updates that data structure so at this point i'm not convinced the attack is thwarted.

kpeluso:
@relyt29 Good call on the attack and thank you for explanation. I tested this as part of `TestAppendInference` unit test. Please review.

x/emissions/keeper/keeper.go
kpeluso: See my latest comment/convo with @relyt29

x/emissions/keeper/actor_utils/quantile_test.go
```diff
@@ -0,0 +1,104 @@
...
+	result, err = actorutils.GetQuantileOfScores(scores, quantile)
+	require.NoError(t, err)
+	expectedInt, err := expectedResult.Int64()
+	require.NoError(t, err)
+	actualInt, err := result.Int64()
+	require.NoError(t, err)
+	require.Equal(t, expectedInt, actualInt)
+}
+
+func TestGetQuantileOfScores2(t *testing.T) {
```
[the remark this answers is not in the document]
kpeluso: done

x/emissions/keeper/actor_utils/quantile.go
```diff
@@ -0,0 +1,99 @@
...
+			} else {
+				return 0
+			}
+		}
+	})
+
+	// If there are no scores then the quantile of scores is 0.
+	// This better ensures chain continuity without consequence because in this situation
+	// there is no meaningful quantile to calculate.
+	if len(scores) == 0 {
```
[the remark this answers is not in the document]
kpeluso: done

x/emissions/keeper/msgserver/msg_server_worker_payload.go
```diff
@@ -11,6 +11,9 @@ import (
 
 // A tx function that accepts a individual inference and forecast and possibly returns an error
 // Need to call this once per forecaster per topic inference solicitation round because protobuf does not nested repeated fields
+// Only 1 payload per registered worker is kept, ignore the rest. In particular, take the first payload from each
+// registered worker and none from any unregistered actor.
+// Signatures, anti-synil procedures, and "skimming of only the top few workers by EMA score descending" should be done here.
```
relyt29, line 16: s/synil/sybil/

x/emissions/keeper/actor_utils/worker.go
```diff
@@ -141,19 +140,14 @@ func insertInferencesFromTopInferers(
...
 	}
 
 	return acceptedInferers, nil
 }
 
-// Output a new set of forecasts where only 1 forecast per registered forecaster is kept,
-// ignore the rest. In particular, take the first forecast from each registered forecaster
-// and none from any unregistered forecaster.
-// Signatures, anti-synil procedures, and "skimming of only the top few workers by score
-// descending" should be done here.
```
relyt29, line 156: we should have some function description here if only for the principle of the thing

x/emissions/keeper/actor_utils/worker.go
```diff
@@ -102,11 +102,10 @@ func CloseWorkerNonce(k *keeper.Keeper, ctx sdk.Context, topicId keeper.TopicId,
...
 
-// Output a new set of inferences where only 1 inference per registered inferer is kept,
-// ignore the rest. In particular, take the first inference from each registered inferer
-// and none from any unregistered inferer.
-// Signatures, anti-synil procedures, and "skimming of only the top few workers by score
-// descending" should be done here.
+// It is assumed `inferences` come from unique, registered, top inferers by EMA score descending
+// It is also assumed that the inferences are for the correct topic and nonce
+// The returned map is the set of inferers that were accepted
+// The returned error is an error that occurred during the process
```
relyt29, line 108: can change return values to named variables for extra clarity

relyt29: line 119 outstanding todo

relyt29 (approved):
This PR builds off quite a bit of work from #548, which I wrote so I am familiar with it.

The rest looks good based on my (twice now) reviews. I think the attack I described earlier is covered as well, though I will be looking to add unit test coverage now

x/emissions/keeper/msgserver/msg_server_worker_payload.go
relyt29: self resolving

x/emissions/proto/emissions/v3/query.proto
```diff
@@ -302,11 +290,24 @@ service Query {
     option (google.api.http).get = "/emissions/v3/inference_scores_until_block/{topic_id}/{block_height}";
   }
 
+  rpc GetPreviousTopicQuantileForecasterScoreEma(QueryGetPreviousTopicQuantileForecasterScoreEmaRequest) returns (QueryGetPreviousTopicQuantileForecasterScoreEmaResponse) {
```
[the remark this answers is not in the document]
relyt29: working on this now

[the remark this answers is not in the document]
relyt29: #572

PR #570, opened 2024-08-31
WIP Test fixes for rewards_test - one test failing

xmariachi:
Fixes on the reward_test failing

## Still Left Todo

### DIFF
--- a/x/emissions/module/rewards/rewards_test.go
+++ b/x/emissions/module/rewards/rewards_test.go
@@ -634,7 +634,7 @@ func (s *RewardsTestSuite) setUpTopicWithEpochLength(
 		LossMethod:               "mse",
 		EpochLength:              epochLength,
 		GroundTruthLag:           epochLength,
-		WorkerSubmissionWindow:   epochLength,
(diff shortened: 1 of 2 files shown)


x/emissions/module/rewards/rewards_test.go
```diff
@@ -744,10 +743,11 @@ func (s *RewardsTestSuite) getRewardsDistribution(
 		})
 		require.NoError(err)
 	}
-
-	err = actorutils.CloseWorkerNonce(&s.emissionsKeeper, s.ctx, topicId, *inferenceBundles[0].Nonce)
-	s.Require().NoError(err)
-
+	// Advance to close the window
+	newBlock := blockHeight + topic.WorkerSubmissionWindow
```
RedBird96, line 747: how about to update `blockHeight` here not use `newBlock`? Because we need to increase this `blockHeight` at next.

xmariachi: hmmm... although we use blockHeight later as a ref for the reputer nonce (+ gt_lag) so I think I'd keep it as is. However they could bear better names.

PR #572, opened 2024-09-03
Add GetCurrentLowestXScore query function

relyt29:

This adds a query function that lets inferers, forecasters, and reputers to check what the cutoff boundary is for making it into the active set, before sending a transaction on chain

Adds `TestGetCurrentLowestXScore` functions in `x/emissions/keeper/queryserver/query_server_score_test.go`


fernandofcampos (approved):

x/emissions/keeper/queryserver/query_server_score.go
```diff
@@ -114,6 +115,46 @@ func (qs queryServer) GetWorkerInferenceScoresAtBlock(
 	return &types.QueryWorkerInferenceScoresAtBlockResponse{Scores: &workerInferenceScores}, nil
 }
 
+func (qs queryServer) GetCurrentLowestInfererScore(
```
RedBird96, line 118: This is different with `socre_utils/GetLowScoreFromAllInferences()`?. I think we can use this function for this query, isn't it?

x/emissions/keeper/queryserver/query_server_score.go
```diff
@@ -144,6 +185,46 @@ func (qs queryServer) GetWorkerForecastScoresAtBlock(
 	return &types.QueryWorkerForecastScoresAtBlockResponse{Scores: &workerForecastScores}, nil
 }
 
+func (qs queryServer) GetCurrentLowestForecasterScore(
```
RedBird96, line 187: Same above with `GetLowScoreFromAllForecasts()`?

x/emissions/keeper/queryserver/query_server_score.go
relyt29: we might be able to let me check

relyt29: yeah i think we can, will replace

relyt29:
@RedBird96 can you take another look

RedBird96 (approved):
lgtm

PR #575, opened 2024-09-03
Do not migrate active topic management 

relyt29:
## Purpose of Changes and their Description

This PR forgoes migrating which topics are active, in favor of an approach where we manually re-activate topics after the upgrade.

## Link(s) to Ticket(s) or Issue(s) resolved by this PR
#574 

## Are these changes tested and documented?

Re-uses existing tests.

PR #580, opened 2024-09-05
Add emissions v4 migration

guilherme-brandao:
- New migration to fix the incomplete v3 emissions module migration


relyt29 (commented):
can copy migration test over too from #575

relyt29 (approved):

PR #581, opened 2024-09-05
Proto changes v0.4.1

relyt29:
1. Update a minimal set of protos, rather than copying the entire v3 folder to a v4 folder
2. Rename MsgServer -> MsgServiceServer (to appease `buf` linter)


relyt29:
ok in order to get the linter to pass I had to rename a bunch of files so if you want to see the real diff it's just commit [a682432](https://github.com/allora-network/allora-chain/pull/581/commits/a6824329052827baff26e69c8300553c884da233)

PR #586, opened 2024-09-06
Update version names: v0.3.0->v0.4.0; v0.4.0->v0.4.1 . Register upgrade.

fernandofcampos:

Update version names: v0.3.0->v0.4.0; v0.4.0->v0.4.1 . Register upgrade.

## Still Left Todo


relyt29 (approved):
this looks good to me, we should check ConsensusVersion for the two modules as well

relyt29:
just checked, emissions is good, mint needs a version bump

relyt29:
actually, do we need to bump the module version if we're not doing a migration, just changing the RPC endpoints / business logic?

fernandofcampos:
> actually, do we need to bump the module version if we're not doing a migration, just changing the RPC endpoints / business logic?

Apparently not, but the documentation is not 100% clear: https://docs.cosmos.network/v0.50/learn/advanced/upgrade#tracking-module-versions

They say, for instance: "For example, if the bank module was version 2 and an upgrade introduces bank module 3, the Cosmos SDK upgrades the bank module and runs the "version 2 to 3" migration script."

Not sure if this is just a trigger for Cosmos to run the migrations, or if there's something else. 

Since we changed the rpc endpoints, can we consider this a breaking change?

relyt29:
I think it makes sense to keep the proto versioning in line with the module ConsensusVersion, though I see cosmos-sdk does not do that (their mint module is at consensus version 2, but their proto is at v1beta1)

Anyways I'm inclined to bump the version just on account

PR #579, opened 2024-09-05
Release v0.4.1

relyt29:
## Purpose of Changes and their Description

This is a small release to fix an incomplete migration in 0.4.0, as well as a new mint module endpoint to stimulate the recalculation of the inflation target at any time.

## Link(s) to Ticket(s) or Issue(s) resolved by this PR

See Changelog.txt

PR #587, opened 2024-09-06
Update Migration Tests to include NaN Initial Regrets Test

relyt29:

In #579 @xmariachi asked for someone to write migration tests that actually check that Initial Regrets set to NaN are correctly migrated by the 0.4.1 test. I went and wrote that test, then updated the 0.4.0 tests that had been wholly copied over and not edited at all, to actually test the things 0.4.1 wants to test (in fact they were not even testing the 0.4.1 migration before, they were importing the 0.4.0 migration and testing that migation instead)

#579 

This PR introduces four new tests for the 0.4.1 migration: `TestMigratedTopicWithNoProblems` `TestMigratedTopicWithNaNInitialRegret` `TestNotMigratedTopic` and `TestNotMigratedTopicWithNaNInitialRegret`

Whether something was "migrated" or "not migrated" has to do with whether the topic was successfully migrated during the 0.4.0 migration.

No changes to documentation

Not necessary to edit changelog, as this is meant to be merged as part of #579.


RedBird96 (approved):

xmariachi:
❤️ 
FYI:
My unit test request was more for the actual code cases than for the migration.
Ie complex tests that go beyond the execution of one function, but more complex tests that assert condition that last longer than a block.
 
Both are important ofc, but unit test last and define behaviour expectations, migration does not.
Ty

xmariachi:
❤️ 
FYI:
My unit test request was more for the actual code cases than for the migration.
Ie complex tests that go beyond the execution of one function, but more complex tests that assert conditions that last longer than a block.
 
Both are important ofc, but code unit tests "last" and define behavioural expectations, migration does not.

Ty for reading ;)

PR #588, opened 2024-09-06
Add Metadata Max Length Parameter (changes protobuf)

relyt29:

In #579 we discovered that we forgot to do adequate input sanitization on the parameters of a new topic being created for v0.4.0. This includes input sanitization for the new parameters, but also we were not checking the length of the string metadata field or the string LossMethod field. This PR enforces length boundaries, so that strings uploaded to the chain are limited to a 256 byte long string.

N/A

- [ ] If tested, please describe how. If not, why tests are not need
  We don't test this, probably should

todo
todo

## Still Left Todo


test/stress/create_topic_test.go
```diff
@@ -17,7 +17,7 @@ func createTopic(
 	creator NameAccountAndAddress,
 ) (topicId uint64) {
 	ctx := context.Background()
-	createTopicRequest := &emissionstypes.MsgCreateNewTopic{
+	createTopicRequest := &emissionstypes.CreateNewTopicRequest{
```
xmariachi, line 20: the new param introduced MaxStringLength does not seem to be checked against (ie expect rejection). If it is then I overlooked it.

xmariachi (commented):
Question about message type changes. The change of message names, will be a breaking compat change with offchain-node. 
Wondering whether this should be 0.5.0 and not 0.4.1 then, since this will not be "backward-compatible bug fixes".

cc @kpeluso  @fernandofcampos

test/stress/create_topic_test.go
relyt29: the check is on LossMethod and Metadata, both of which are acceptably short here

the tests are in `TestMsgCreateNewTopicTooLongMetadataFails` and `TestMsgCreateNewTopicTooLongLossMethodFails` in `x/emissions/keeper/msgserver/msg_server_topics_test.go`

sorry if I'm misunderstanding your question?

x/emissions/migrations/v4/migrate_test.go
```diff
@@ -62,13 +63,119 @@ func (s *EmissionsV4MigrationTestSuite) SetupTest() {
...
 
+// in this test we check that the emissions module params have been migrated
+// and the expected new field is set.
+func (s *EmissionsV4MigrationTestSuite) TestMigrateParams() {
+	storageService := s.emissionsKeeper.GetStorageService()
+	store := runtime.KVStoreAdapter(storageService.OpenKVStore(s.ctx))
+	cdc := s.emissionsKeeper.GetBinaryCodec()
+
+	defaultParams := types.DefaultParams()
+	paramsOld := v3types.Params{
```
RedBird96, line 79: This should be v2types, isn't it?

RedBird96 (approved):
Yes, I also agree with diego, that the message name changes could be moved to v0.5.0
Just left one comment and others looks good to me!

x/emissions/proto/emissions/v4/tx.proto
```diff
@@ -0,0 +1,411 @@
+syntax = "proto3";
```
kpeluso, line 1: wonder if this PR changes given https://github.com/allora-network/allora-chain/pull/589#pullrequestreview-2288953783

x/emissions/migrations/v4/migrate.go
```diff
@@ -6,7 +6,8 @@ import (
 	storetypes "cosmossdk.io/store/types"
 	alloraMath "github.com/allora-network/allora-chain/math"
 	"github.com/allora-network/allora-chain/x/emissions/keeper"
-	oldtypes "github.com/allora-network/allora-chain/x/emissions/migrations/v3/types"
+	v2types "github.com/allora-network/allora-chain/x/emissions/migrations/v3/types"
+	v3types "github.com/allora-network/allora-chain/x/emissions/migrations/v4/types"
```
kpeluso, line 10: you call it v3types but pointing at v4 ?

x/emissions/migrations/v4/migrate_test.go
kpeluso: think "v2types" may need to be renamed "v3types" but i share your question otherwise

test/stress/create_topic_test.go
xmariachi: ok i missed them sorry

x/emissions/migrations/v4/migrate.go
relyt29: yes correct, that is intentional

The types for v3, which is v0.4.0, which is now _in the past_ are stored in the migration folder for the v4 migration, because the v4 migration needs to remember the types from the past in order to migrate.
The types for v2, which is v0.3.0, which is _two migrations ago_ are stored in the migration folder for the v3 migration, because the v3 migration needs to remember the types from the past. In addition, because we messed up doing the v3 migration, now the v4 migration ALSO needs to reference those old types. Hence the naming.

I will rename it further to make it even more explicit.

x/emissions/migrations/v4/migrate_test.go
relyt29: see my comment on the other conversation.

this is the correct types.

Params needs to be migated from v3 types because v0.4.0 correctly migrated params from 2->3

Topics need to be migrated from the v2 types, because v0.4.0 failed to correctly migrate topics. Hence, the topics need to be migrated from 2 straight to 4.

kpeluso (approved):

x/emissions/migrations/v4/migrate_test.go
xmariachi: > Topics need to be migrated from the v2 types

Are we sure about this? Because topic types in testnet v0.4.0 are using v3 types too, aren't they?... problem is that one of them didnt get its values changed, and mig to v3 didn't fail because of the nature of the change (additive only) allowed for filling with default values and not break. 

In practice I guess both will work

relyt29: yeah since it takes the value from the GetTopic call, and the only thing that's used is the oldMsg.Id we could _probably_ use the v3 types. But using v2 types is safer because we can only guarantee that the v2 types will be present, we're not guaranteed that the v3 types will be present

probably it's not that important, but v2 is safer so that's what i went with

PR #590, opened 2024-09-06
Clear NaN in maps by deleting all map values during migration

relyt29:

This PR enforces the "clearing of poison" by removing all not a number (NaN) values in the chain. It does so by completely and totally deleting the contents of the inference scores, forecaster scores, emas, loss bundles etc maps.

Tested in `TestResetMapsWithNonNumericValues`

No documentation, no changes to change log as this is part of #579.

### DIFF
--- a/x/emissions/migrations/v4/migrate.go
+++ b/x/emissions/migrations/v4/migrate.go
@@ -8,6 +8,7 @@ import (
 	"github.com/allora-network/allora-chain/x/emissions/keeper"
 	v2types "github.com/allora-network/allora-chain/x/emissions/migrations/v3/types"
 	v3types "github.com/allora-network/allora-chain/x/emissions/migrations/v4/types"
+	"github.com/allora-network/allora-chain/x/emissions/types"
 	emissionstypes "github.com/allora-network/allora-chain/x/emissions/types"
 	"github.com/cosmos/cosmos-sdk/codec"
 	"github.com/cosmos/cosmos-sdk/runtime"
@@ -15,6 +16,13 @@ import (
 	"github.com/gogo/protobuf/proto"
 )
 
+const maxPageSize = uint64(10000)
+
+// MigrateStore migrates the store from version 3 to version 4
+// it does the following:
+// - migrates params
+// - migrates topics
+// - Deletes the contents of several maps that had NaN values in them
 func MigrateStore(ctx sdk.Context, emissionsKeeper keeper.Keeper) error {
 	ctx.Logger().Info("MIGRATING STORE FROM VERSION 3 TO VERSION 4")
 	storageService := emissionsKeeper.GetStorageService()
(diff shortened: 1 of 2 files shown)


x/emissions/migrations/v4/migrate.go
```diff
@@ -156,6 +174,61 @@ func MigrateTopics(
 	return nil
 }
 
+// Deletes all keys in the store with the given keyPrefix `maxPageSize` keys at a time
+func safelyClearWholeMap(store storetypes.KVStore, keyPrefix []byte) {
+	s := prefix.NewStore(store, keyPrefix)
```
xmariachi, line 179: are we deleting all from each store? Aren't we just attempting to delete topic 5 data?
(remember this will take some more time when running the update on nodes ... worth showing some progress log on the stores being cleaned maybe?)

relyt29: We're deleting everything, I think it's safer.

I don't think it will take as long because there's only a few days of inferences saved as opposed to months in the case of the previous migration

RedBird96 (approved):

x/emissions/migrations/v4/migrate.go
kpeluso: the number of inferences is the same because the pruning window is the same, but the number of calculated worker scores may be less

x/emissions/migrations/v4/migrate.go
```diff
@@ -156,6 +174,61 @@ func MigrateTopics(
...
+	safelyClearWholeMap(store, emissionstypes.AllLossBundlesKey)
+	safelyClearWholeMap(store, emissionstypes.NetworkLossBundlesKey)
+	safelyClearWholeMap(store, emissionstypes.InfererNetworkRegretsKey)
+	safelyClearWholeMap(store, emissionstypes.ForecasterNetworkRegretsKey)
+	safelyClearWholeMap(store, emissionstypes.OneInForecasterNetworkRegretsKey)
+	safelyClearWholeMap(store, emissionstypes.LatestNaiveInfererNetworkRegretsKey)
+	safelyClearWholeMap(store, emissionstypes.LatestOneOutInfererInfererNetworkRegretsKey)
+	safelyClearWholeMap(store, emissionstypes.LatestOneOutInfererForecasterNetworkRegretsKey)
+	safelyClearWholeMap(store, emissionstypes.LatestOneOutForecasterInfererNetworkRegretsKey)
+	safelyClearWholeMap(store, emissionstypes.LatestOneOutForecasterForecasterNetworkRegretsKey)
```
kpeluso, line 228: Please add logs to each line so we're able to track progress this time around.

kpeluso (changes requested):

x/emissions/migrations/v4/migrate.go
relyt29: addressed in [b9ade84](https://github.com/allora-network/allora-chain/pull/590/commits/b9ade8439d6adf7fa88f50b44cf72cf6bb502aa7)

kpeluso (approved):

xmariachi (approved):

PR #592, opened 2024-09-07
Remove rewardable topics

RedBird96:
Use an array of active topics to distribute rewards instead of rewardable topics. And remove add/clear rewardable topics at end block.
Also call `PickChurnableActiveTopics` before `EmitRewards` because need to set topic reward nonce

## Still Left Todo


x/emissions/module/rewards/rewards.go
```diff
@@ -36,12 +36,8 @@ func EmitRewards(
 		return errors.Wrapf(err, "failed to get module params")
 	}
 
-	rewardableTopics, err := k.GetRewardableTopics(ctx)
-	if err != nil {
-		return errors.Wrapf(err, "failed to get rewardable topics")
-	}
 	// Sorted, active topics by weight descending. Still need skim top N to truly be the rewardable topics
-	sortedRewardableTopics := alloraMath.GetSortedElementsByDecWeightDesc(rewardableTopics, weights)
+	sortedRewardableTopics := alloraMath.GetSortedElementsByDecWeightDesc(weights)
```
xmariachi, line 40: I understand that here weights contain the whole set of topics with weights (if not, then pls dismiss the rest of the comment A )

A: 
Then we take the top N by weight (N=1 or 2)
Then we attempt  to distribute rewards to that topic, where we check whetther this topic has an open reputer nonce.
If I'm reading this right and this is the case, we'll end up with the same problem as before: top N shadowing other topics in the selection, even if theiir "cadence" is not met this block, preventing them to generate rewards. 

In that case, it looks like the topics should check whether they have a topicRewardNonce before entering selection group.

B: 
if we only get active topics at that block: will not this be "advancing the payment of rewards" by 1 block, right? 

Please double check that no "nonce-related" code is affected by this order change, since now the nonces would be closed and given rewards on the same block, mostly if these checks happen on rewards code and expecting that to be for `block - 1`.

xmariachi (approved):
reviewed and tested ;)

x/emissions/module/abci.go
```diff
@@ -43,13 +43,6 @@ func EndBlocker(ctx context.Context, am AppModule) error {
...
 	sdkCtx.Logger().Debug(fmt.Sprintf("ABCI EndBlocker %d: Total Revenue: %v, Sum Weight: %v", blockHeight, totalRevenue, sumWeight))
 
-	// REWARDS (will internally filter any non-RewardReady topics)
-	err = rewards.EmitRewards(sdkCtx, am.keeper, blockHeight, weights, sumWeight, totalRevenue)
-	if err != nil {
-		sdkCtx.Logger().Error("Error calculating global emission per topic: ", err)
-		return errors.Wrapf(err, "Rewards error")
-	}
-
 	err = rewards.PickChurnableActiveTopics(
```
kpeluso, line 46: Should rename this to be `UpdateNoncesOfActiveTopics` and update the comments atop its declaration.

kpeluso (changes requested):
Think it's important to [update this](https://github.com/allora-network/allora-chain/pull/592/files#r1749312002) + other comments referencing concepts of earlier designs, but other than that this looks great.

x/emissions/module/rewards/rewards.go
RedBird96: Yes, this weights contains topicIds and weights. And I can't agree with your comment(A). If the topic does not met "cadence" can't be contained this weights array ig.

kpeluso (approved):
great!

PR #596, opened 2024-09-09
EMA: NaN protection

xmariachi:
Protect EMA against previous or incoming `NaN` values.

PROTO-2302 

- [X] If tested, please describe how. If not, why tests are not needed.  -- go test
- [X] If documented, please describe where. If not, describe why docs are not needed. -- trivial, not needed

### DIFF
--- a/math/utils.go
+++ b/math/utils.go
@@ -20,14 +20,14 @@ func CalcEma(
 	firstTime bool,
 ) (Dec, error) {
 	// If first iteration, then return just the new value
-	if firstTime || current.Equal(previous) || previous.isNaN {
-		if current.isNaN {
-			return ZeroDec(), errors.New("at least one ema operand should not be NaN")
-		}
+	if current.isNaN {
+		return ZeroDec(), errors.New("Current EMA operand should be NaN")
+	}
+	if firstTime || current.Equal(previous) {
 		return current, nil
 	}
-	if current.isNaN {
-		return previous, nil
+	if previous.isNaN {
+		return ZeroDec(), errors.New("Previous EMA operand should be NaN")
 	}
 	alphaCurrent, err := alpha.Mul(current)
 	if err != nil {
--- a/math/utils.go
+++ b/math/utils.go
@@ -2,9 +2,8 @@ package math
 
 import (
 	"cmp"
-	"sort"
-
 	"errors"
+	"sort"
 )
 
 // all exponential moving average functions take the form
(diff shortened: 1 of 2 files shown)


RedBird96 (approved):
lgtm

relyt29 (approved):
I'm not sure if it's great to allow current to be NaN, that feels like something we should catch before the Ema is happening or maybe throw and refuse to take the  Ema, but I guess it's fine

math/utils.go
```diff
@@ -21,9 +21,15 @@ func CalcEma(
...
 	// If first iteration, then return just the new value
-	if firstTime || current.Equal(previous) {
+	if firstTime || current.Equal(previous) || previous.isNaN {
+		if current.isNaN {
+			return ZeroDec(), errors.New("at least one ema operand should not be NaN")
+		}
 		return current, nil
 	}
+	if current.isNaN {
+		return previous, nil
```
kpeluso, line 31: No logging here?

kpeluso (changes requested):
Would rather see a more comprehensive and complete tracking of when NaNs are calculated plus assurances that we actually are not calculating NaNs, because we shouldn't be. That can be 2 separate tasks.

math/utils.go
xmariachi: yes, it could be added.
However this math package does not have any logging currently. Not sure what's the best way. 
Should we pass in a Context into the functions?

xmariachi:
> Would rather see a more comprehensive and complete tracking of when NaNs are calculated plus assurances that we actually are not calculating NaNs, because we shouldn't be. That can be 2 separate tasks.

Yes, I created separate ticket for that, because ensure

xmariachi:
sorry ab the close, I hit the wrong button by mistake.

math/utils.go
relyt29: alright i've thought further and I think that comment was too knee-jerk

we can pass the ctx in, logging is important - but i still feel that this is not the place where we need to be catching issues

it's a failure of somewhere else that we're thinking about this here

xmariachi: I agree that this type of utils library is ok to have no logging, just "result or error". 
And in that light,  I think calcEMA should also fail, and not correct, current or previous EMAs. No need for logging, just fail if any piece involved in the calcs is NaN. 
I will change that

xmariachi: Just pushed the change. This is also consistent with the rest of the EMA protection through the file.

PR #598, opened 2024-09-10
Math Operations on NaNs Should Return Errors

relyt29:

This PR introduces checks in the Allora math library when operating on NaN values. This PR makes it so that on addition, subtraction, `.Abs()`, etc of a value with a NaN, the addition or subtraction etc function returns an error.

This PR also takes a coverage based approach to write tests for many math functions that previously did not have coverage.

This PR also refactors various functions in the math package.

PROTO-2315

These changes are tested by our existing unit tests. In addition, I have added  tests to cover the new error cases when a math function ought to error.

Documentation changes are not required since allora math is an internally consumed library.

Added to changelog.


math/utils.go
```diff
@@ -94,7 +95,7 @@ func StdDev(data []Dec) (Dec, error) {
 	// Calculate the mean, excluding NaN values
 	for _, v := range data {
 		if v.isNaN { // Check if the value is NaN
-			continue // Skip NaN values
+			return Dec{}, errors.New("input data contains NaN values")
```
xmariachi, line 98: my other PR introduces earlier validation on these functions right on the start of the function, for all functions in this file.

relyt29: great, pulled your changes into my PR and merged them

xmariachi: ty. 
I also see that in loop cases like this, your code is more efficient, so feel free to keep it.

fernandofcampos (approved):

math/utils_test.go
```diff
@@ -283,6 +370,127 @@ func TestWeightedInferences2(t *testing.T) {
...
+	m := map[int32]*alloraMath.Dec{
+		1: &dec1,
+		2: &dec2,
+		3: &dec3,
+		4: &dec4,
+		5: &dec5,
+	}
+
+	expected := []int32{4, 1, 5, 2, 3}
+	keys := alloraMath.GetSortedElementsByDecWeightDesc(m)
```
xmariachi, line 494: Some of this unit tests do not include `NaN`s. 
IAs (Copilot too) do a good job at generating batteries of test cases (incl edge cases) for each function including numbers like "NaN" or "+/-Infinity" and providing expected results.

xmariachi (approved):
Great work!

math/utils_test.go
relyt29: yeah this test specifically is not meant to test NaNs, it's a happy path test

I didnt' include NaN tests for this function because the function itself doesn't check for NaNs

which I think is ok, because it's a sort function - it's not processing the number, let it sort the NaN

PR #603, opened 2024-09-13
Validate all Values for Correctness Prior to Storing them in the Keeper

relyt29:

In order to aid in ensuring correct state, and to practice defense-in-depth, we should check all values for correctness at the final step, right before the values are written into the keeper.

This code moves all the separate validation functions in the `types` folder into one `validations.go` file which is easier for someone looking for validations to go look for, rather than trying to remember the name of the type being validated.

It modifies all of the various keeper.go functions that do keeper.xyz.Set() functions. I went through and manually checked every single field in the keeper for where it's set and set up a validator prior to that line.

PROTO-2341, all of the NaN work in release v0.5.0

This code is tested by our existing unit tests, which proceed to fall over and took a week of rewriting to make pass.

This PR does not change any user or external facing code, so no documentation updates are needed.

PR added to the `Unreleased` section of `CHANGELOG.md`

## Still Left Todo

Unit tests are in #608 just to keep the PR diff smaller for reviewers.

PROTO-2369 was created by me to track one place in the code where we do incomplete validation in this PR: we modify loss bundles at the filtering step in CloseReputerNonce. This causes signature verification on the bundles to fail when it comes time to insert them into the database. The correct solution for this is to change the data type that we are storing these modified filtered loss bundles in. For now that can stay as future work.


x/emissions/types/topic.go
```diff
@@ -0,0 +1,150 @@
...
+		return errors.Wrap(sdkerrors.ErrInvalidRequest, "active reputer quantile must be between 0 and 1 inclusive")
+	}
+
+	return nil
+}
+
+// Validate checks if the given Topic is valid
+func (topic Topic) Validate(params Params) error {
+	if topic.Id == 0 {
+		return errors.Wrap(sdkerrors.ErrInvalidType, "topic id zero is reserved")
```
guilherme-brandao, line 65: do we really want this amount of redundancy? Topics just can be set by using the createTopic which already has basically the same validation.

x/emissions/keeper/keeper.go
```diff
@@ -2070,6 +2081,37 @@ func (k *Keeper) IsReputerRegisteredInTopic(ctx context.Context, topicId TopicId
...
+			return err
+		}
+	}
+	return k.blockToActiveTopics.Set(ctx, block, topicIds)
+}
+
+func (k *Keeper) SetTopicToNextPossibleChurningBlock(ctx context.Context, topicId TopicId, block BlockHeight) error {
+	if err := types.ValidateTopicId(topicId); err != nil {
+		return err
+	}
```
guilherme-brandao, line 2108: How can we ever get a topic id == 0 in these situations?

guilherme-brandao (commented):
I'm not sure about this amount of redundancy; I'm wondering if some of these validations are really necessary. Some validations like the topic Id in some of the keeper methods seems excessive. However, I don't have the full context of the issues that happened last week, so you can just choose to ignore what I'm saying.

x/emissions/types/topic.go
relyt29: Defense in depth 

I think the approach ought to be, do all this redundant stuff get the chain bulletproof, then later go back with an optimization pass

x/emissions/keeper/keeper.go
guilherme-brandao: This paranoia can go far xD But yeah, if we are planning to optimize this later, it's fine by me. It can be a pain in terms of maintenance if we keep this pattern all around. Also, I thought we'd just focus on values that are used in calculations and are suffering mutations... but again, if we plan to change this in the future, that's ok for now.

relyt29: yeah I think we will be able to remove some validations from other places if we want to keep them at the keeper

but yes I recognize not all of these validations are strictly necessary in all cases

but being thorough is better, after all we are building a financial system

relyt29:
The tests for this PR are going in #608

x/emissions/keeper/genesis.go
```diff
@@ -52,6 +53,9 @@ func (k *Keeper) InitGenesis(ctx context.Context, data *types.GenesisState) erro
 	//ActiveTopics []uint64
 	if len(data.ActiveTopics) != 0 {
 		for _, topicId := range data.ActiveTopics {
+			if topicId == 0 {
```
randygrok, line 56: you can reuse the  types.ValidateTopicId(topicAndActorId.TopicId)

x/emissions/keeper/genesis.go
```diff
@@ -678,6 +792,12 @@ func (k *Keeper) InitGenesis(ctx context.Context, data *types.GenesisState) erro
 	//TopicToNextPossibleChurningBlock []*topicBlock
 	if len(data.TopicToNextPossibleChurningBlock) != 0 {
 		for _, topicBlock := range data.TopicToNextPossibleChurningBlock {
+			if topicBlock.TopicId == 0 {
```
randygrok, line 795: same, can be part of method Validate() from the struct

x/emissions/types/topic.go
spooktheducks: Agree, we can always roll back later

relyt29:
rebased against dev to remove working commits

x/emissions/types/validations.go
```diff
@@ -1,18 +1,764 @@
...
+
 	"cosmossdk.io/errors"
+	cosmosMath "cosmossdk.io/math"
 	alloraMath "github.com/allora-network/allora-chain/math"
+	"github.com/cometbft/cometbft/crypto/secp256k1"
+	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
+/// EXTERNAL TYPE VALIDATIONS
```
guilherme-brandao, line 14: Why throw all the validations (dec, int, all types, etc) here?

relyt29: because I was finding it really hard to find them in the various disparate files when I was looking for them

if you're already in the codebase somewhere e.g. in worker_payload.go and you hit the "jump to code" command it'll automatically open the validate function in the right file and that's fine and easy

but if you're somewhere else thinking "I want to look at the validation for x type" you have to trawl through a crowded types folder full of different filenames, instead of just thinking "i wanted to look at the validations" go straight to validations.go

also if you want to look at multiple validations at the same time easier to have them in one spot then have to look for each separately

x/emissions/keeper/inference_synthesis/network_losses.go
```diff
@@ -231,6 +249,12 @@ func CalcNetworkLosses(
 	}
 
 	output := emissions.ValueBundle{
+		TopicId:             topicId,
+		ReputerRequestNonce: &emissions.ReputerRequestNonce{ReputerNonce: &emissions.Nonce{BlockHeight: blockHeight}},
+		// we re-use the value bundle for the network loss bundle
+		// this is bad practice and we should refactor: PROTO-2369
+		// for now set the Reputer address to a "zero" address
```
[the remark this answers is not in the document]
relyt29: Created PROTO-2464 to track that, thanks

x/emissions/types/validations.go
kpeluso: I softly don't support files like this (e.g. this and `keeper.go` both poorly formed files). A compromise here is to split this up into many files within a validations package

x/emissions/types/validations.go
```diff
@@ -1,18 +1,764 @@
...
+	if err := ValidateBech32(oc.Owner); err != nil {
+		return errors.Wrap(err, "offchain node owner is invalid")
+	}
+	return nil
+}
+
+/// PROTOBUF MESSAGE VALIDATIONS
+
+// validate that a register request follows the expected format
+func (msg *RegisterRequest) Validate() error {
```
kpeluso, line 643: note for future: wonder how we can write linter to check that every field of every struct was validated. tangentially realted to exhaustruct PR

x/emissions/keeper/genesis.go
```diff
@@ -1086,6 +1137,25 @@ func (k *Keeper) ExportGenesis(ctx context.Context) (*types.GenesisState, error)
 		previousForecastRewardFraction = append(previousForecastRewardFraction, &topicIdActorIdDec)
 	}
 
+	/* bug in genesis export, previousForecasterScoreRatio is not correct type in genesis.proto
```
kpeluso, line 1140: can't add as another type and decode here as correct type...?

x/emissions/types/validations.go
relyt29: I see where you're coming from @kpeluso with the "long-ass files" argument.

So long as these functions stay as methods of the structs that they process, they have to live in a types package, I can see us doing what you describe (a separate validations package) but then those need to be exported functions that take the things they're validating as args, as opposed to being methods on structs

x/emissions/keeper/genesis.go
relyt29: no it's a separate bug completely and I felt was out of scope for this PR. See PROTO-2349 for more details

x/emissions/types/validations.go
relyt29: I can implement that, if you really want in this PR

kpeluso (approved):
Would change some things organizationally but important to have asap

x/emissions/types/validations.go
kpeluso: it's fine. tbh my suggestion might've been worse than just leaving them as they were

guilherme-brandao (approved):

PR #604, opened 2024-09-13
Migration added stores, deletion reward, handle NaN, etc

xmariachi:

* Removal of additional stores in migration
* NaN Handling failing, 
* Ensure indexes of arrays match

* Rewards
  * Checks on empty array sizes before calling functions that will fail if so.
  * Per-topic refactor of rewards: failure of one topic will not make any other not be given rewards
  * Prune of  rewards data in case of failure anyway (otherwise it would linger forever - observed)
  * Ensure coefficients and reputers match and are valid
  
* autocli fix of `multi-coefficient` function

PROTO-2339
PROTO-2340

- [X] If tested, please describe how. If not, why tests are not needed.
Pre-Tested applying migration from 
Functionality doesn't change in practice, just fixes. 

Pending: reapply to early situation with bad data on 0.3.1 and test with a complete migration again.


x/emissions/keeper/keeper.go
```diff
@@ -1661,6 +1660,9 @@ func (k *Keeper) GetDelegateRewardPerShare(ctx context.Context, topicId TopicId,
 // Set the share on specific reputer and topicId
 func (k *Keeper) SetDelegateRewardPerShare(ctx context.Context, topicId TopicId, reputer ActorId, share alloraMath.Dec) error {
 	key := collections.Join(topicId, reputer)
+	if share.IsNaN() || !share.IsFinite() {
```
relyt29, line 1663: in the types package there is a function `validateDec` - you may need to uppercase it to export it but I would advise using that function rather than this in case we want to add additional requirements in the future

x/emissions/module/rewards/rewards.go
```diff
@@ -159,6 +121,44 @@ func EmitRewards(
 	return nil
 }
 
+// New function for processing each topic individually
+func processTopic(
```
relyt29, line 125: can we write a unit test for this

relyt29: also process topic is a generic name that isn't great at conveying meaning of what this function does

x/emissions/module/rewards/rewards.go
```diff
@@ -159,6 +121,44 @@ func EmitRewards(
...
+
+	topicReward := topicRewards[topicId]
+	if topicReward == nil {
+		Logger(ctx).Warn(fmt.Sprintf("Topic %d has no reward, skipping", topicId))
+		return alloraMath.ZeroDec(), nil
+	}
+
+	Logger(ctx).Debug(fmt.Sprintf("Generating rewards distribution for topic: %d, topicRewardNonce: %d, topicReward: %s", topicId, topicRewardNonce, topicReward))
+
+	// Distribute rewards between topic participants
```
relyt29, line 143:
```suggestion
	// Get the distribution of rewards across actor types and participants in this topic
```

x/emissions/module/rewards/rewards.go
relyt29:
```suggestion
func getDistributionAndPayoutRewardsToTopicActors(
```

maybe something like this, idk

relyt29 (approved):
small nit-picky comments

PR directionally looks good to me

x/emissions/keeper/keeper.go
```diff
@@ -1595,7 +1594,7 @@ func (k *Keeper) GetStakeReputerAuthority(ctx context.Context, topicId TopicId,
 // Includes the stake placed by delegators on the reputer in that topic.
 func (k *Keeper) SetStakeReputerAuthority(ctx context.Context, topicId TopicId, reputer ActorId, amount cosmosMath.Int) error {
 	key := collections.Join(topicId, reputer)
-	if amount.IsZero() {
+	if amount.IsNil() || amount.IsZero() {
```
kpeluso, line 1597: does nil actually have the same meaning as 0? and not more closer to how we're dealing with NaNs?

x/emissions/module/rewards/rewards.go
```diff
@@ -66,86 +66,48 @@ func EmitRewards(
 		return errors.Wrapf(err, "failed to calculate topic rewards")
 	}
 
-	// Calculate then pay out topic rewards to topic participants
-	totalRewardToStakedReputers := alloraMath.ZeroDec() // This is used to communicate with the mint module
-	for _, topicId := range sortedRewardableTopics {
-		Logger(ctx).Debug(fmt.Sprintf("Distributing rewards for topic %d", topicId))
```
kpeluso, line 72: why remove logs like this?

kpeluso (changes requested):
Most importantly... where did the call to `pruneRecordsAfterRewards()` go??

x/emissions/module/rewards/rewards.go
xmariachi: good calls

x/emissions/keeper/keeper.go
xmariachi: makes sense. 
And there is an additional benefit of exporting validations: that they can be used in allora-offchain-node too.

xmariachi:
> Most importantly... where did the call to `pruneRecordsAfterRewards()` go??

Was missing in case of success, added it back in. Good catch.

x/emissions/module/rewards/rewards.go
xmariachi: it's been moved inside the `getDistributionAndPayoutRewardsToTopicActors` function

x/emissions/keeper/keeper.go
xmariachi: This is cosmosMath, not alloraMath. In any case maybe it's more of a question for @relyt29 : what's the usage and validation policy regarding `cosmosMath.Int` ?
And then it should be applied to all cases, not just this one; and be included in the Data Validation PR.

x/emissions/module/rewards/rewards.go
```diff
@@ -66,86 +66,48 @@ func EmitRewards(
...
-				fmt.Sprintf(
-					"Failed to Generate Rewards for Topic, Skipping:\nTopic Id %d\nTopic Reward Amount %s\nError:\n%s\n\n",
-					topicId,
-					topicRewardString,
-					err.Error(),
-				),
-			)
+			Logger(ctx).Error(fmt.Sprintf("Failed to process rewards for topic %d: %s", topicId, err.Error()))
+			// Prune records after rewards anyway in case of error
+			err = pruneRecordsAfterRewards(ctx, k, moduleParams.MinEpochLengthRecordLimit, topicId, topicRewardNonce)
```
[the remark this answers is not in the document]
xmariachi: done

x/emissions/module/rewards/rewards.go
xmariachi: pending unit test

RedBird96 (approved):

kpeluso (approved):

PR #607, opened 2024-09-17
Add events for research monitoring

RedBird96:

Need to add some events for building the research monitoring suite.
For this I added events for forecasting task score, topic reward and tokenomics update.

PROTO-2353

## Still Left Todo


x/emissions/proto/emissions/v4/events.proto
```diff
@@ -0,0 +1,84 @@
...
+  ];
+}
+
+message EventNetworkLossSet {
+  uint64 topic_id = 1;
+  int64 block_height = 2;
+  emissions.v3.ValueBundle value_bundle = 3;
+}
+
+message EventForecastTaskScoreSet {
```
relyt29, line 58: I see an EventScoreSet and an EventForecastTaskScoreSet

is EventScoreSet for inferers, should it be called EventInferenceTaskScoreSet then?

Do we fire an event when reputers have their scores changed?


are these scores, or score EMAs?

RedBird96: `EventScoresSet` handles all performance scores for `inferes`, `forecasters` and `reputers`. And I created this `EventForecastTaskScoreSet` for forecast task score because this is used at only forecasts.
And all of these are score not EMAs. Do you think we need to emit EMA as well?

relyt29: well EMAs are what are actually used for inclusion on chain, so I would think they're more important than instantaneous scores but I'm also of the opinion that we should just log everything all the time

kpeluso: Could actually be a really good idea to emit the EMA score with a boolean that indicates if the actor is part of the active or passive set for that round.

So an event with schema: `actor, topic, epoch nonce, ema score, isActive`

Been getting questions related to being able to tell if one is in the active/passive set.

RedBird96: thanks guys
I will add this event

x/emissions/keeper/actor_utils/worker.go
```diff
@@ -97,6 +97,11 @@ func CloseWorkerNonce(k *keeper.Keeper, ctx sdk.Context, topicId keeper.TopicId,
 		return err
 	}
 
+	types.EmitNewWorkerLastCommitSetEvent(ctx, topic.Id, blockHeight, &nonce)
+	if err != nil {
```
kpeluso, line 101: where is the error here (and above) being reset??

x/mint/module/abci.go
```diff
@@ -88,6 +89,7 @@ func BeginBlocker(ctx context.Context, k keeper.Keeper) error {
 		if err != nil {
 			return err
 		}
+		types.EmitNewEcosystemTokenMintSetEvent(sdkCtx, blockHeight, tokensToMint)
```
kpeluso, line 92: I don't love how this is called only once per month but i see how it can get expensive to do it more than per block just for an event...

x/mint/proto/mint/v3/events.proto
```diff
@@ -0,0 +1,39 @@
+syntax = "proto3";
+package mint.v3;
```
kpeluso, line 2: We at minimum have to update `ConsensusVersion = 3` in `x/mint` when we do this. I followed up with this in my commit

x/emissions/keeper/actor_utils/worker.go
kpeluso: fixed in my commit

x/emissions/module/rewards/rewards_internal.go
```diff
@@ -163,9 +163,9 @@ func GetfUniqueAgg(numForecasters alloraMath.Dec) (alloraMath.Dec, error) {
 	return ret, nil
 }
 
-// GetFinalWorkerScoreForecastTask calculates the worker score in forecast task.
+// GetFinalWorkerPerformanceScore calculates the worker performance score.
 // T_ik
-func GetFinalWorkerScoreForecastTask(
+func GetFinalWorkerPerformanceScore(
```
kpeluso, line 168: better

x/mint/module/abci.go
kpeluso: @relyt29 fortunately also made `EmissionInfo` in `x/mint` query server so actually we can add that to our allora producer. Should be fine. And now this way we can have both

x/emissions/module/rewards/scores.go
```diff
@@ -209,6 +213,7 @@ func GenerateInferenceScores(
 	}
 
 	types.EmitNewInfererScoresSetEvent(ctx, newScores)
+	types.EmitNewActorEMAScoresSetEvent(ctx, types.ActorType_ACTOR_TYPE_INFERER_UNSPECIFIED, emaScores, activeArr)
```
kpeluso, line 216: You did this for all active set calculations, but forgot to for all passive set calculations.
Will do in my commit.

x/emissions/module/rewards/scores.go
```diff
@@ -305,7 +312,16 @@ func GenerateForecastScores(
 		return nil, err
 	}
 
+	// Calculate forecast task score and emit
+	forecastingTaskScore, err := networkLosses.NaiveValue.Sub(networkLosses.CombinedValue)
```
kpeluso, line 316: this isn't right... you shouldn't have to recalculate a whitepaper formula -- they should all already be there...

x/emissions/module/rewards/scores.go
```diff
@@ -233,7 +239,7 @@ func GenerateForecastScores(
 		newScore := types.Score{
 			TopicId:     topicId,
 			BlockHeight: block,
-			Address:     networkLosses.InfererValues[0].Worker,
```
kpeluso, line 236: Great bugfix...

kpeluso: cc @guilherme-brandao

x/mint/proto/mint/v3/events.proto
kpeluso: @relyt29 @guilherme-brandao Anything else we have to do here to upgrade the module. No migration needed I believe in this case.

PR #608, opened 2024-09-17
Tests for #603

relyt29:
I'm separating out the code to fix the tests because there are 265 failed tests to fix

:sob:


relyt29:
last test, on the second round, only 3 of the inferences are being accepted. Why don't know

relyt29:
integration test has been very flakey lately, not sure if related

PR #610, opened 2024-09-18
ensure every score blockheight is built with nonce not current height

kpeluso:

Scores were inconsistently using current height or nonce. I modified to just use nonces.

I used this to also modify how we were checking for idempotent submissions (1 submission per actor type per topic per epoch).

This is a simpler solution than https://github.com/allora-network/allora-chain/pull/609

[From conversation](https://github.com/allora-network/allora-chain/pull/609) with @xmariachi and @guilherme-brandao 

## Still Left Todo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Security
 
 * [#588](https://github.com/allora-network/allora-chain/pull/588)  Add String Max Length Module Parameter, Enforce Max String Length on Creation of New Topics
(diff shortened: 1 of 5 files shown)


xmariachi (approved):
Aligned on solution.

This may still cause inconsistent block behaviour on scores when actors move from passive to active set (and so may get new scores with a lower blockHeight).

* Note 1: As a better endgame solution, I think adding an additional txBlock field to the score would be more consistent with the scores-defined-separately for active-passive sets.

* Note 2: I think the EMAs are still done even if new scores' blocks are lower. This is consistent with current solution. If this is done somewhere else in the code (I haven't found it, this notion comes from a separate discussion), I think that condition should be removed.

guilherme-brandao (approved):

relyt29:
I didn't follow the whole conversation so I may be missing things, but this solutions _looks_ safe to me

relyt29 (approved):
You should add this to changelog.md under security

PR #613, opened 2024-09-25
Fix invariant tests

guilherme-brandao:
- Fixing context usage in invariant tests


relyt29 (approved):
looks reasonable to me, need to rebase against latest origin/dev and you'll have working integration tests

PR #615, opened 2024-09-26
Add worker experience to initial regret update

kpeluso:

1. Improve implementation of merit-based sortition -- Calculate percentile with instantaneous scores instead of EMA scores => Faster cycling through actors
2. Resolve runaway regrets
   1. Create migration to clear off regrets and initial regrets
   2. Apply formula developed by @jmdkastro [here](https://linear.app/alloralabs/issue/RES-381/investigate-initial-regret-issue-engineering#comment-2ec7fb4a)
   ```
   denominator = std(regrets[i-1, :]) + epsilon
   offset = cnorm - 8.25 / pnorm
   dummy_regret[i] = np.percentile(regrets[i-1, :], 25.) + offset * denominator
   ```
   which ensures the topic initial regret reaches an appropriate equilibrium.
4. Overwrite `topic.AllowNegative` to `false` in all topics -- was wrongly set as `true` somehow

* https://linear.app/alloralabs/issue/RES-381/investigate-initial-regret-issue-engineering#comment-2ec7fb4a
* https://linear.app/alloralabs/issue/PROTO-2455/add-experience-to-implement-updated-solution-for-sustain-the-regrets


x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -123,10 +123,18 @@ func GetCalcSetNetworkRegrets(
 		return errorsmod.Wrapf(err, "failed to get topic")
 	}
 
-	ShouldAddWorkerRegret := func(worker Worker, newParticipant bool) (bool, error) {
```
kpeluso: great catch

x/emissions/keeper/keeper.go
```diff
@@ -539,6 +539,9 @@ func (k *Keeper) GetCountInfererInclusionsInTopic(ctx context.Context, topicId T
 	key := collections.Join(topicId, inferer)
 	count, err := k.countInfererInclusionsInTopic.Get(ctx, key)
 	if err != nil {
+		if errors.Is(err, collections.ErrNotFound) {
```
kpeluso, line 542: also great catch

x/emissions/keeper/keeper.go
```diff
@@ -526,6 +533,52 @@ func (k *Keeper) DeleteUnfulfilledReputerNonces(ctx context.Context, topicId Top
 	return k.unfulfilledReputerNonces.Set(ctx, topicId, types.ReputerRequestNonces{})
 }
 
+/// INCLUSIONS
+
+func (k *Keeper) GetCountInfererInclusionsInTopic(ctx context.Context, topicId TopicId, inferer ActorId) (uint64, error) {
+	key := collections.Join(topicId, inferer)
+	count, err := k.countInfererInclusionsInTopic.Get(ctx, key)
+	if err != nil {
```
spooktheducks, line 541: little bit cleaner/more canonical way to write this:
```go
if errors.Is(err, collections.ErrNotFound) {
    return 0, nil
} else if err != nil {
    return 0, err
}
return count, nil
```

x/emissions/keeper/keeper.go
```diff
@@ -526,6 +533,52 @@ func (k *Keeper) DeleteUnfulfilledReputerNonces(ctx context.Context, topicId Top
...
+		if errors.Is(err, collections.ErrNotFound) {
+			return 0, nil
+		}
+		return 0, err
+	}
+	return count, nil
+}
+
+func (k *Keeper) IncrementCountInfererInclusionsInTopic(ctx context.Context, topicId TopicId, inferer ActorId) error {
```
spooktheducks, line 550: We can rely on these never being called from multiple goroutines, right?

x/emissions/keeper/keeper.go
```diff
@@ -526,6 +533,52 @@ func (k *Keeper) DeleteUnfulfilledReputerNonces(ctx context.Context, topicId Top
...
+	}
+	count++
+	return k.countInfererInclusionsInTopic.Set(ctx, key, count)
+}
+
+func (k *Keeper) GetCountForecasterInclusionsInTopic(ctx context.Context, topicId TopicId, forecaster ActorId) (uint64, error) {
+	key := collections.Join(topicId, forecaster)
+	count, err := k.countForecasterInclusionsInTopic.Get(ctx, key)
+	if err != nil {
+		if errors.Is(err, collections.ErrNotFound) {
```
spooktheducks, line 564: ditto re: structuring err blocks comment above

x/emissions/keeper/queryserver/query_server_inclusions_test.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"github.com/allora-network/allora-chain/x/emissions/types"
+)
+
+func (s *QueryServerTestSuite) TestGetCountInfererInclusionsInTopic() {
+	s.CreateOneTopic()
+	ctx := s.ctx
+	queryserver := s.queryServer
+	topicId := uint64(1)
+	inferer := "allo10es2a97cr7u2m3aa08tcu7yd0d300thdct45ve"
```
spooktheducks, line 12: Can we make a `testutil.RandomAddress()` func?  Or if this value needs to be exactly this because of test setup, can this be made into a testutils variable?

x/emissions/migrations/v5/migrate.go
```diff
@@ -0,0 +1,198 @@
...
+// ActiveInfererQuantile, ActiveForecasterQuantile, or ActiveReputerQuantile,
+// set those values to the default.
+func MigrateTopics(
+	ctx sdk.Context,
+	store storetypes.KVStore,
+	cdc codec.BinaryCodec,
+	emissionsKeeper keeper.Keeper,
+) error {
+	topicStore := prefix.NewStore(store, emissionstypes.TopicsKey)
+	iterator := topicStore.Iterator(nil, nil)
```
spooktheducks, line 57: `defer iterator.Close()`?

x/emissions/migrations/v5/migrate.go
```diff
@@ -0,0 +1,198 @@
...
+			return errorsmod.Wrapf(err, "failed to get topic")
+		}
+		newTopic := copyTopic(topic)
+		// fix egregious values
+		newTopic.InitialRegret = alloraMath.MustNewDecFromString("0")
+		// was wrongly set to true for existing topics
+		newTopic.AllowNegative = false
+		topicsToChange[string(iterator.Key())] = newTopic
+	}
+	_ = iterator.Close()
```
spooktheducks, line 84: Never omit `err` checking!

x/emissions/migrations/v5/migrate.go
```diff
@@ -0,0 +1,198 @@
...
+		// Delete the keys
+		for _, key := range keysToDelete {
+			s.Delete(key)
+		}
+	}
+}
+
+// Clear out poison NaN values on different inferences, scores etc
+func ResetMapsWithNonNumericValues(ctx sdk.Context, store storetypes.KVStore, cdc codec.BinaryCodec) {
+	ctx.Logger().Info("MIGRATION V4: RESETTING infererScoresByBlock MAP")
```
spooktheducks, line 129: Should these logs say `V5`?

x/emissions/migrations/v5/migrate_test.go
```diff
@@ -0,0 +1,388 @@
+package v4_test
```
spooktheducks, line 1: `v5`?

x/emissions/migrations/v5/migrate_test.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+
+	bz, err := proto.Marshal(&score)
+	s.Require().NoError(err)
+
+	mapStore := prefix.NewStore(store, key)
+	mapStore.Set([]byte("testKey"), bz)
+
+	// Sanity check
+	iterator := mapStore.Iterator(nil, nil)
```
spooktheducks, line 166: `defer` the close

x/emissions/migrations/v5/migrate_test.go
```diff
@@ -0,0 +1,388 @@
...
+	s.Require().True(iterator.Valid())
+	err = proto.Unmarshal(iterator.Value(), &score)
+	s.Require().NoError(err)
+	iterator.Close()
+
+	v5.ResetMapsWithNonNumericValues(s.ctx, store, cdc)
+
+	// Verify the store has been updated correctly
+	iterator = mapStore.Iterator(nil, nil)
```
spooktheducks, line 175: `defer` the close

x/emissions/migrations/v5/migrate_test.go
```diff
@@ -0,0 +1,388 @@
...
+	scores := emissionstypes.Scores{Scores: []*emissionstypes.Score{&score}}
+
+	bz, err := proto.Marshal(&scores)
+	s.Require().NoError(err)
+
+	mapStore := prefix.NewStore(store, key)
+	mapStore.Set([]byte("testKey"), bz)
+
+	// Sanity check
+	iterator := mapStore.Iterator(nil, nil)
```
spooktheducks, line 202: `defer` the close

x/emissions/migrations/v5/migrate_test.go
```diff
@@ -0,0 +1,388 @@
...
+	s.Require().True(iterator.Valid())
+	err = proto.Unmarshal(iterator.Value(), &scores)
+	s.Require().NoError(err)
+	iterator.Close()
+	s.Require().Len(scores.Scores, 1)
+
+	v5.ResetMapsWithNonNumericValues(s.ctx, store, cdc)
+
+	// Verify the store has been updated correctly
+	iterator = mapStore.Iterator(nil, nil)
```
spooktheducks, line 212: `defer` the close

x/emissions/migrations/v5/migrate_test.go
```diff
@@ -0,0 +1,388 @@
...
+) {
+	bundle := getBundle()
+	bz, err := proto.Marshal(&bundle)
+	s.Require().NoError(err)
+
+	mapStore := prefix.NewStore(store, key)
+	mapStore.Set([]byte("testKey"), bz)
+
+	// Sanity check
+	iterator := mapStore.Iterator(nil, nil)
```
spooktheducks, line 303: `defer` close

x/emissions/migrations/v5/migrate_test.go
```diff
@@ -0,0 +1,388 @@
...
+	s.Require().True(iterator.Valid())
+	err = proto.Unmarshal(iterator.Value(), &bundle)
+	s.Require().NoError(err)
+	iterator.Close()
+	s.Require().Equal(bundle, getBundle())
+
+	v5.ResetMapsWithNonNumericValues(s.ctx, store, cdc)
+
+	// Verify the store has been updated correctly
+	iterator = mapStore.Iterator(nil, nil)
```
spooktheducks, line 313: `defer` close

x/emissions/migrations/v5/migrate_test.go
```diff
@@ -0,0 +1,388 @@
...
+		},
+	}
+	bz, err := proto.Marshal(&reputerValueBundles)
+	s.Require().NoError(err)
+
+	mapStore := prefix.NewStore(store, key)
+	mapStore.Set([]byte("testKey"), bz)
+
+	// Sanity check
+	iterator := mapStore.Iterator(nil, nil)
```
spooktheducks, line 342: `defer` close

x/emissions/migrations/v5/migrate_test.go
```diff
@@ -0,0 +1,388 @@
...
+	s.Require().True(iterator.Valid())
+	err = proto.Unmarshal(iterator.Value(), &reputerValueBundles)
+	s.Require().NoError(err)
+	iterator.Close()
+	s.Require().Len(reputerValueBundles.ReputerValueBundles, 1)
+
+	v5.ResetMapsWithNonNumericValues(s.ctx, store, cdc)
+
+	// Verify the store has been updated correctly
+	iterator = mapStore.Iterator(nil, nil)
```
spooktheducks, line 352: `defer` close

x/emissions/migrations/v5/migrate_test.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+
+	bz, err := proto.Marshal(&timeStampedValue)
+	s.Require().NoError(err)
+
+	mapStore := prefix.NewStore(store, key)
+	mapStore.Set([]byte("testKey"), bz)
+
+	// Sanity check
+	iterator := mapStore.Iterator(nil, nil)
```
spooktheducks, line 376: `defer` close

x/emissions/migrations/v5/migrate_test.go
```diff
@@ -0,0 +1,388 @@
...
+	s.Require().True(iterator.Valid())
+	err = proto.Unmarshal(iterator.Value(), &timeStampedValue)
+	s.Require().NoError(err)
+	iterator.Close()
+
+	v5.ResetMapsWithNonNumericValues(s.ctx, store, cdc)
+
+	// Verify the store has been updated correctly
+	iterator = mapStore.Iterator(nil, nil)
```
spooktheducks, line 385: `defer` close

x/emissions/module/rewards/rewards.go
```diff
@@ -513,8 +513,20 @@ func payoutRewards(
 
 			if reward.Type == types.WorkerInferenceRewardType {
 				infererRewards = append(infererRewards, reward)
+
```
spooktheducks, line 516: Might be worth writing tests for these changes

spooktheducks (changes requested):

x/emissions/keeper/genesis.go
```diff
@@ -1803,6 +1821,42 @@ func (k *Keeper) ExportGenesis(ctx context.Context) (*types.GenesisState, error)
 		previousTopicQuantileReputerScoreEma = append(previousTopicQuantileReputerScoreEma, &topicIdAndDec)
 	}
 
+	countInfererInclusionsInTopic := make([]*types.TopicIdActorIdUint64, 0)
+	countInfererInclusionsInTopicIter, err := k.countInfererInclusionsInTopic.Iterate(ctx, nil)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to iterate count inferer inclusions in topic")
+	}
```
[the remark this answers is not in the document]
relyt29: that's probably worth another PR by itself, but good point

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -118,6 +118,35 @@ func GetCalcSetNetworkRegrets(
 		return networkLosses.InfererValues[i].Worker < networkLosses.InfererValues[j].Worker
 	})
 
+	topic, err := k.GetTopic(ctx, topicId)
+	if err != nil {
+		return errorsmod.Wrapf(err, "failed to get topic")
+	}
+
+	ShouldAddWorkerRegret := func(worker Worker, newParticipant bool, workerType emissions.ActorType) (bool, error) {
```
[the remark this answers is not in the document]
relyt29: also can we not used local variable functions and just write helper functions, this is not javascript

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -392,14 +445,13 @@ func CalcTopicInitialRegret(regrets []alloraMath.Dec, epsilon alloraMath.Dec, pN
 		return alloraMath.ZeroDec(), err
 	}
 
-	minimumRegret := alloraMath.ZeroDec()
-	for i, regret := range regrets {
-		if i == 0 || regret.Lt(minimumRegret) {
-			minimumRegret = regret
-		}
+	// Calculate percentile
+	percentile, err := alloraMath.GetQuantileOfDecs(regrets, alloraMath.MustNewDecFromString("0.25"))
```
relyt29, line 449: should this 25% be a module parameter or topic parameter?

x/emissions/keeper/keeper.go
```diff
@@ -526,6 +533,52 @@ func (k *Keeper) DeleteUnfulfilledReputerNonces(ctx context.Context, topicId Top
 	return k.unfulfilledReputerNonces.Set(ctx, topicId, types.ReputerRequestNonces{})
 }
 
+/// INCLUSIONS
+
+func (k *Keeper) GetCountInfererInclusionsInTopic(ctx context.Context, topicId TopicId, inferer ActorId) (uint64, error) {
```
relyt29, line 538: missing function header describing what the function does

x/emissions/keeper/keeper.go
relyt29, line 550: missing function header comment describing what function does

relyt29: ok rely is a strong word, but we definitely try not to use any goroutines anywhere in the code

it was something we got dinged for in a previous security audit

x/emissions/keeper/keeper.go
```diff
@@ -526,6 +533,52 @@ func (k *Keeper) DeleteUnfulfilledReputerNonces(ctx context.Context, topicId Top
...
+	key := collections.Join(topicId, inferer)
+	count, err := k.GetCountInfererInclusionsInTopic(ctx, topicId, inferer)
+	if err != nil {
+		return err
+	}
+	count++
+	return k.countInfererInclusionsInTopic.Set(ctx, key, count)
+}
+
+func (k *Keeper) GetCountForecasterInclusionsInTopic(ctx context.Context, topicId TopicId, forecaster ActorId) (uint64, error) {
```
relyt29, line 560: function header comment describing what this function does

relyt29: keeper_test.go unit test ensuring it works as expected

x/emissions/keeper/keeper.go
relyt29: keeper_test.go unit test ensuring it works as expected

x/emissions/keeper/keeper.go
relyt29: keeper_test.go unit test ensuring it works as expected

x/emissions/keeper/keeper.go
```diff
@@ -526,6 +533,52 @@ func (k *Keeper) DeleteUnfulfilledReputerNonces(ctx context.Context, topicId Top
...
+	if err != nil {
+		if errors.Is(err, collections.ErrNotFound) {
+			return 0, nil
+		}
+		return 0, err
+	}
+	return count, nil
+}
+
+func (k *Keeper) IncrementCountForecasterInclusionsInTopic(ctx context.Context, topicId TopicId, forecaster ActorId) error {
```
relyt29, line 572: function header comment describing what this function does

unit test in keeper_test ensuring it operates as expected

x/emissions/keeper/queryserver/query_server_inclusions_test.go
relyt29: this should be s.addrsStr[0] instead

every test suite has a handful of addrs on hand for testing with. Nothing needs to be hardcoded ever

x/emissions/keeper/queryserver/query_server_inclusions_test.go
```diff
@@ -0,0 +1,53 @@
...
+	s.Require().NoError(err)
+	s.Equal(results.Count, uint64(1))
+}
+
+func (s *QueryServerTestSuite) TestGetCountForecasterInclusionsInTopic() {
+	s.CreateOneTopic()
+	ctx := s.ctx
+	queryserver := s.queryServer
+	topicId := uint64(1)
+	forecaster := "allo1snm6pxg7p9jetmkhz0jz9ku3vdzmszegy9q5lh"
```
relyt29, line 36: same here, should be s.addrsStr[0]

x/emissions/migrations/v5/migrate.go
```diff
@@ -0,0 +1,198 @@
...
+	emissionstypes "github.com/allora-network/allora-chain/x/emissions/types"
+	"github.com/cosmos/cosmos-sdk/codec"
+	"github.com/cosmos/cosmos-sdk/runtime"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/gogo/protobuf/proto"
+)
+
+const maxPageSize = uint64(10000)
+
+// MigrateStore migrates the store from version 3 to version 4
```
relyt29, line 19: is that right? cause the migration is in a folder called v5

x/emissions/migrations/v5/migrate.go
```diff
@@ -0,0 +1,198 @@
...
+// it does the following:
+// - migrates params
```
relyt29, line 21: i don't see any params or topics migration, this comment header is just wrong

x/emissions/migrations/v5/migrate.go
```diff
@@ -0,0 +1,198 @@
...
+
+	ctx.Logger().Info("MIGRATING EMISSIONS MODULE FROM VERSION 4 TO VERSION 5 COMPLETE")
+	return nil
+}
+
+// migrate topics for this new version
+// iterate through all topics, keep all the old values of these topics
+// if a topic has a NaN value for InitialRegret, set the value to zero
+// if the topic doesn't have a value for MeritSortitionAlpha,
+// ActiveInfererQuantile, ActiveForecasterQuantile, or ActiveReputerQuantile,
```
relyt29, line 48: I don't see anything about active quantiles in the code

x/emissions/migrations/v5/migrate.go
```diff
@@ -0,0 +1,198 @@
...
+	iterator := topicStore.Iterator(nil, nil)
+
+	// iterate over the topic store by manually checking prefixes
+	// to make sure we absolutely do not miss any topics
+	// (as opposed to using collections.go API)
+	topicsToChange := make(map[string]emissionstypes.Topic, 0)
+	for ; iterator.Valid(); iterator.Next() {
+		ctx.Logger().Info("MIGRATION V5: Updating topic", iterator.Key())
+		iterator.Key()
+		var oldMsg oldV2Types.Topic
```
relyt29, line 66: do we need to use oldV2Types topics here? At the point of this upgrade, v0.5.0 has already been live

we can probably just use the collections api directly

x/emissions/migrations/v5/migrate.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// Clear out poison NaN values on different inferences, scores etc
+func ResetMapsWithNonNumericValues(ctx sdk.Context, store storetypes.KVStore, cdc codec.BinaryCodec) {
+	ctx.Logger().Info("MIGRATION V4: RESETTING infererScoresByBlock MAP")
+	safelyClearWholeMap(ctx, store, emissionstypes.InferenceScoresKey)
+	ctx.Logger().Info("MIGRATION V4: RESETTING forecasterScoresByBlock MAP")
+	safelyClearWholeMap(ctx, store, emissionstypes.ForecastScoresKey)
+	ctx.Logger().Info("MIGRATION V4: RESETTING reputerScoresByBlock MAP")
+	safelyClearWholeMap(ctx, store, emissionstypes.ReputerScoresKey)
```
relyt29, line 134: do we really need to clear out the score maps? regrets aren't scores?

x/emissions/migrations/v5/migrate_test.go
```diff
@@ -0,0 +1,388 @@
...
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/suite"
+
+	moduletestutil "github.com/cosmos/cosmos-sdk/types/module/testutil"
+
+	storetypes "cosmossdk.io/store/types"
+	cosmostestutil "github.com/cosmos/cosmos-sdk/testutil"
+)
+
+type EmissionsV4MigrationTestSuite struct {
```
relyt29, line 34: V5

x/emissions/migrations/v5/migrate_test.go
```diff
@@ -0,0 +1,388 @@
...
+	suite.Suite
+	ctrl *gomock.Controller
+
+	ctx             sdk.Context
+	storeService    store.KVStoreService
+	emissionsKeeper *keeper.Keeper
+}
+
+func TestEmissionsV4MigrationTestSuite(t *testing.T) {
```
relyt29, line 43: V5

x/emissions/migrations/v5/migrate_test.go
```diff
@@ -0,0 +1,388 @@
...
+		storeService,
+		accountKeeper,
+		bankKeeper,
+		authtypes.FeeCollectorName)
+
+	s.emissionsKeeper = &emissionsKeeper
+}
+
+// in this test, we check that an already migrated topic, that has all the new fields
+// for merit sortition, but has a NaN for initial regret, will have its initial regret
```
relyt29, line 71: I'm confused. Are we actually seeing NaN for initial regrets? I thought we were seeing very large values (huge enormous, negative values) which are non-NaN

x/emissions/module/rewards/scores.go
```diff
@@ -129,7 +129,7 @@ func GenerateReputerScores(
 	}
 
 	// Update topic quantile of EMA score
```
relyt29, line 131: comment is not updated

I'm also not convinced of the reasoning to use instantaneous scores over EMAs, which are more stable and less likely to be manipulated in the short term

x/emissions/module/rewards/scores.go
```diff
@@ -129,7 +129,7 @@ func GenerateReputerScores(
...
-	topicEmaScoreQuantile, err := actorutils.GetQuantileOfScores(emaScores, topic.ActiveReputerQuantile)
+	topicEmaScoreQuantile, err := actorutils.GetQuantileOfScores(newScores, topic.ActiveReputerQuantile)
```
relyt29, line 132: variable name is wrong assuming you intend to use instantaneous scores over ema scores

x/emissions/module/rewards/scores.go
```diff
@@ -203,7 +203,7 @@ func GenerateInferenceScores(
 	}
 
 	// Update topic quantile of EMA score
-	topicEmaScoreQuantile, err := actorutils.GetQuantileOfScores(emaScores, topic.ActiveInfererQuantile)
+	topicEmaScoreQuantile, err := actorutils.GetQuantileOfScores(newScores, topic.ActiveInfererQuantile)
```
relyt29, line 206: same comment as above

x/emissions/module/rewards/scores.go
```diff
@@ -303,7 +303,7 @@ func GenerateForecastScores(
 	}
 
 	// Update topic quantile of EMA score
-	topicEmaScoreQuantile, err := actorutils.GetQuantileOfScores(emaScores, topic.ActiveForecasterQuantile)
+	topicEmaScoreQuantile, err := actorutils.GetQuantileOfScores(newScores, topic.ActiveForecasterQuantile)
```
relyt29, line 306: same comment as above

x/emissions/migrations/v5/migrate.go
xmariachi: Yes I don't think we _need_ to clear scores or ema_scores afaik 

Same for coefficients, quantiles, reward fractions I think

x/emissions/migrations/v5/migrate_test.go
xmariachi: No NaNs afaik, just huge neg values

x/emissions/module/rewards/rewards.go
```diff
@@ -124,8 +124,8 @@ func EmitRewards(
 }
 
 // This function distributes and pays out rewards to topic actors based on their participation.
-// It returns the total reward distributed to actors
-func getDistributionAndPayoutRewardsToTopicActors(
+// It returns the total reward distributed to reputers
+func payoutRewardsToTopicActors(
```
relyt29, line 128: I disagree with this naming change as it obscures what this function does 

generating the distribution of scores is majorly important work. Naming this payoutRewards conflates the meaning of it with the `payoutRewards` function which unlike this function is just a dumb function that does `SendFromModuleToModule`

x/emissions/keeper/genesis.go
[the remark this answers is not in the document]
RedBird96: yes, I agree. This can be addressed at another PR

math/utils_test.go
```diff
@@ -494,3 +494,33 @@ func TestGetSortedElementsByDecWeightDesc(t *testing.T) {
 	keys := alloraMath.GetSortedElementsByDecWeightDesc(m)
 	require.Equal(t, expected, keys)
 }
+
+func TestGetQuantileOfDecs(t *testing.T) {
+	m := []alloraMath.Dec{
```
xmariachi, line 499: add tests where: 
* empty set
* 1 element set
* negative errors 
* inclusion of NaN/ negative errors in the set. 

These can be generated using AI in a table-based manner.

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -138,7 +144,19 @@ func GetCalcSetNetworkRegrets(
 		if err != nil {
 			return errorsmod.Wrapf(err, "Error setting inferer regret")
 		}
-		if !newParticipant {
+
+		shouldAddWorkerRegret, err := isExperiencedWorker(
```
relyt29, line 148: do we need to make research aware of the change here for the purposes of keeping the paper aligned with the codebase? 

@jmdkastro FYI now we will only use "experienced" workers for Network Regrets

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -392,17 +428,50 @@ func CalcTopicInitialRegret(regrets []alloraMath.Dec, epsilon alloraMath.Dec, pN
...
 	if err != nil {
 		return alloraMath.ZeroDec(), err
 	}
 
 	return initialRegret, nil
 }
+
+// Determine if a worker is an experienced worker by checking
+// the inclusions count is greater than 1/alpha_regret
+func isExperiencedWorker(
```
relyt29, line 447: can we write separate isExperiencedForecaster and isExperiencedInferer functions

I am afraid this actorType enum is a footgun waiting to happen

x/emissions/migrations/v5/migrate.go
relyt29: this is not resolved

Would like clarity on why this migration needs to use old types because I don't think the Topics type has changed from v0.5.0 -> v0.6.0 

Therefore the entire iterator.Valid() flow can use collections.go wrappers with cleaner APIs and safety rather than acessing the key value store directly

please let me know if this is not the case

x/emissions/migrations/v5/migrate_test.go
relyt29: so this comment can be updated then...

x/emissions/keeper/inference_synthesis/network_regrets.go
RedBird96: Yes, I think the research team has already tested this solution

RedBird96 (dismissed):

x/emissions/migrations/v5/migrate.go
RedBird96: I think we are using collections.go [here](https://github.com/allora-network/allora-chain/blob/b5ec36cd142b8d09c62f9de5b0caf3476fefa921/x/emissions/migrations/v5/migrate.go#L70) to get topic. And used `iterator.Valid` to get all topic Id ig

xmariachi (approved):
Good work, I approve from my end, though I strongly suggest to get additional approvals and resolve open convos.

relyt29:
PLEASE REBASE THIS BRANCH ONTO DEV AGAIN, AS DEV HAS BEEN PUSHED TO A NEW HEAD THAT REMOVES A BAD COMMIT
```
git fetch 
git rebase -i origin/dev
```
DO NOT UPDATE IN GITHUB UI, YOU WILL REINTRODUCE THE BAD COMMIT IN YOUR MERGE COMMIT - I REPEAT DO NOT MERGE

@RedBird96

x/emissions/keeper/inference_synthesis/network_regrets.go
relyt29: @kpeluso I think we need a new emissions module parameter for this, I think this should not be hardcoded to 25%

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -392,17 +426,66 @@ func CalcTopicInitialRegret(regrets []alloraMath.Dec, epsilon alloraMath.Dec, pN
...
+	numInclusionsDec, err := alloraMath.NewDecFromUint64(numInclusions)
+	if err != nil {
+		return false, errorsmod.Wrapf(err, "failed to get num inclusions dec")
+	}
+	oneOverAlpha, err := alloraMath.OneDec().Quo(topic.AlphaRegret)
+	if err != nil {
+		return false, errorsmod.Wrapf(err, "failed to get one over alpha")
+	}
+	return !newParticipant && numInclusionsDec.Gte(oneOverAlpha), nil
+}
```
relyt29, line 491:
```suggestion
func isExperiencedInferer(
    ctx context.Context,
    k keeper.Keeper,
    topic emissions.Topic,
    inferer Worker,
    newParticipant bool,
) (bool, error) {
    numInclusions := uint64(0)
    numInclusions, err := k.GetCountInfererInclusionsInTopic(ctx, topic.Id, inferer)
    if err != nil {
        return false, errorsmod.Wrapf(err, "failed to get inferer inclusions")
    }
    return isExperiencedActor(numInclusions)
}

// Determine if a forecaster is an experienced worker by checking
// the inclusions count is greater than 1/alpha_regret
func isExperiencedForecaster(
    ctx context.Context,
    k keeper.Keeper,
    topic emissions.Topic,
    forecaster Worker,
    newParticipant bool,
) (bool, error) { 
    numInclusions := uint64(0)
    numInclusions, err := k.GetCountForecasterInclusionsInTopic(ctx, topic.Id, forecaster)
    if err != nil {
        return false, errorsmod.Wrapf(err, "failed to get forecaster inclusions")
    }
    return isExperiencedActor(numInclusions)
}

// helper function for isExperiencedForecaster and isExperiencedInferer
// check if inclusions count is greater than 1/alpha_regret
func isExperiencedActor(numInclusions uint64) {
    numInclusionsDec, err := alloraMath.NewDecFromUint64(numInclusions)
    if err != nil {
        return false, errorsmod.Wrapf(err, "failed to get num inclusions dec")
    }
    oneOverAlpha, err := alloraMath.OneDec().Quo(topic.AlphaRegret)
    if err != nil {
        return false, errorsmod.Wrapf(err, "failed to get one over alpha")
    }
    return !newParticipant && numInclusionsDec.Gte(oneOverAlpha), nil
}
```

relyt29 (changes requested):
I am good on the migration code now, just one suggestion which is take it or leave it nitpick, and then I do think we need to have a separate controllable module parameter for that percentile

RedBird96 (approved):

x/emissions/keeper/inference_synthesis/network_regrets.go
relyt29: still need an answer @kpeluso , I am not happy about merging this without at least acknowledgement

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -348,7 +376,11 @@ func GetCalcSetNetworkRegrets(
 
 	// Recalculate topic initial regret
 	if len(workersRegrets) > 0 {
-		updatedTopicInitialRegret, err := CalcTopicInitialRegret(workersRegrets, epsilon, pNorm, cNorm)
+		params, err := k.GetParams(ctx)
+		if err != nil {
+			return errorsmod.Wrapf(err, "Error getting params")
```
relyt29, line 381:
```suggestion
			return errorsmod.Wrapf(err, "GetCalcSetNetworkRegrets error getting params")
```

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -348,7 +376,11 @@ func GetCalcSetNetworkRegrets(
...
+		}
+		updatedTopicInitialRegret, err := CalcTopicInitialRegret(workersRegrets, epsilon, pNorm, cNorm, params.RegretPercentile, params.PnormSafeDiv)
```
relyt29, line 383:
```suggestion
		updatedTopicInitialRegret, err := CalcTopicInitialRegret(workersRegrets, epsilon, pNorm, cNorm, params.InitialRegretQuantile, params.PnormSafeDiv)
```

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -392,17 +435,64 @@ func CalcTopicInitialRegret(regrets []alloraMath.Dec, epsilon alloraMath.Dec, pN
 		return alloraMath.ZeroDec(), err
 	}
 
-	minimumRegret := alloraMath.ZeroDec()
-	for i, regret := range regrets {
-		if i == 0 || regret.Lt(minimumRegret) {
-			minimumRegret = regret
-		}
+	// Calculate percentile
```
relyt29, line 438: it's a quantile not a percentile, it's not scaled to 100 percent it's between 0 and 1

x/emissions/proto/emissions/v5/params.proto
```diff
@@ -0,0 +1,155 @@
...
+    (cosmos_proto.scalar) = "cosmos.Int",
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.nullable) = false,
+    (amino.dont_omitempty) = true
+  ];
+  // payload sending fee for reputer or worker
+  uint64 max_elements_per_forecast = 45; // max number of top forecasters by score to reward
+  uint64 max_active_topics_per_block = 46; // max number of active topics per block
+  uint64 max_string_length = 47; // max permittible length of strings uploaded to the chain
+  string regret_percentile = 48 [ // percentile value for getting regret
```
relyt29, line 147:
```suggestion
  string InitialRegretQuantile = 48 [ // quantile value for getting initial regret during network regret calculation
```

relyt29 (dismissed):

RedBird96 (approved):

x/emissions/proto/emissions/v5/query.proto
```diff
@@ -0,0 +1,1241 @@
+syntax = "proto3";
```
kpeluso, line 1: Do we need query.proto v5?

x/emissions/proto/emissions/v5/tx.proto
```diff
@@ -0,0 +1,419 @@
+syntax = "proto3";
```
kpeluso, line 1: Do we need a brand new tx.proto v5?

x/emissions/proto/emissions/v4/query.proto
```diff
@@ -399,6 +399,34 @@ service QueryService {
     option (cosmos.query.v1.module_query_safe) = true;
     option (google.api.http).get = "/emissions/v4/next_churning_block_by_topic_id/{topic_id}";
   }
+
+  rpc GetCountInfererInclusionsInTopic(GetCountInfererInclusionsInTopicRequest) returns (GetCountInfererInclusionsInTopicResponse) {
```
kpeluso, line 403: wonder how necessary this was if we're making v5 query.proto?

x/emissions/proto/emissions/v5/tx.proto
```diff
@@ -0,0 +1,419 @@
...
+  repeated uint64 max_string_length = 47;
+  repeated string initial_regret_quantile = 48 [
+    (gogoproto.customtype) = "github.com/allora-network/allora-chain/math.Dec",
+    (gogoproto.nullable) = false
+  ];
+  repeated string pnorm_safe_div = 49 [
+    (gogoproto.customtype) = "github.com/allora-network/allora-chain/math.Dec",
+    (gogoproto.nullable) = false
+  ];
+}
```
kpeluso, line 185: You failed to add the new parameters here, per the global parameter checklist:
https://www.notion.so/upshothq/Adding-New-Global-Parameters-Checklist-6216783e3b4f4280a2fda6f3487f0164

kpeluso:
overall new changes lgtm

x/emissions/proto/emissions/v5/query.proto
RedBird96: Yes, it is because we added new module params(`InitialRegretQuantile`, `PnormSafeDiv`) so need to update query and tx proto

x/emissions/proto/emissions/v4/query.proto
RedBird96: Yes, I think we don't need to add this query on v4 anymore since we are using v5

spooktheducks (changes requested):

x/emissions/keeper/inference_synthesis/network_regrets.go
RedBird96: it's added as module param

math/utils_test.go
```diff
@@ -494,3 +494,33 @@ func TestGetSortedElementsByDecWeightDesc(t *testing.T) {
 	keys := alloraMath.GetSortedElementsByDecWeightDesc(m)
 	require.Equal(t, expected, keys)
 }
+
+func TestGetQuantileOfDecs(t *testing.T) {
```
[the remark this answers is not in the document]
RedBird96: I created some functions for non-happy paths. I will merge these functions to table driven test

x/emissions/keeper/inference_synthesis/network_regrets_test.go
```diff
@@ -82,10 +82,11 @@ func (s *InferenceSynthesisTestSuite) TestGetCalcSetNetworkRegretsTwoWorkers() {
 	require := s.Require()
 	k := s.emissionsKeeper
 
-	topicId := uint64(2)
+	topicId := uint64(1)
```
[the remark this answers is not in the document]
RedBird96: Through this `TestGetCalcSetNetworkRegretsTwoWorkers` function, we can test only topic which has more than 2 experienced actors update regrets

x/emissions/migrations/v5/migrate.go
```diff
@@ -0,0 +1,231 @@
...
+	nextTopicId, err := emissionsKeeper.GetNextTopicId(ctx)
+	if err != nil {
+		return err
+	}
+	// iterate all topics to migrate using collections.go api
+	topicsToChange := make(map[string]emissionstypes.Topic, 0)
+	for id := uint64(1); id < nextTopicId; id++ {
+		idByte := make([]byte, 8)
+		binary.BigEndian.PutUint64(idByte, id)
+		ctx.Logger().Info(fmt.Sprintf("MIGRATION V5: Updating topic:%d", id))
```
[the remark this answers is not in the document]
RedBird96: Yes, there is a string + argument. However, this argument must be a string type, so I think converting it to a string and using it is the same as using sprintf. That's why cosmos sdk itself uses this way i guess.

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -392,17 +435,64 @@ func CalcTopicInitialRegret(regrets []alloraMath.Dec, epsilon alloraMath.Dec, pN
...
-	initialRegret, err := minimumRegret.Add(offSetTimesDenominator)
+	initialRegret, err := percentile.Add(offSetTimesDenominator)
 	if err != nil {
 		return alloraMath.ZeroDec(), err
 	}
 
 	return initialRegret, nil
 }
+
+// Determine if a inferer is an experienced worker by checking
```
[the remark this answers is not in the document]
RedBird96: This is private function so I didn't make test function. However I think this test function could be replaced with a test for the function that calls this, isn't it?

x/emissions/migrations/v5/migrate.go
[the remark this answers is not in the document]
spooktheducks: I looked into this further, and it's actually what the stdlib `slog` package does.  An `n % 2 == 0` varargs array of `label, value`.  Which is better than format args.  Use structured logging.  This isn't the way anymore.  No more Sprintf in the codebase

x/emissions/keeper/inference_synthesis/network_regrets_test.go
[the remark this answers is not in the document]
spooktheducks: Comment it in the code.  Thoroughly.  It's useless here on a random Github PR

x/emissions/keeper/inference_synthesis/network_regrets_test.go
```diff
@@ -668,8 +671,164 @@ func (s *InferenceSynthesisTestSuite) TestCalcTopicInitialRegret() {
...
+		pNorm,
+		cNorm,
+		percentileRegert,
+		pnormDiv,
+	)
+	require.NoError(err)
+	testutil.InEpsilon5(s.T(), calculatedInitialRegret, "0.3354820760526412097325669544281814")
+}
+
+func (s *InferenceSynthesisTestSuite) TestUpdateTopicInitialRegret() {
```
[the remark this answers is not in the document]
spooktheducks: I think this guy resolves conversations for a living.  Does he have a LinkedIn?  Many founders want to know

x/emissions/keeper/inference_synthesis/network_regrets_test.go
```diff
@@ -668,8 +671,164 @@ func (s *InferenceSynthesisTestSuite) TestCalcTopicInitialRegret() {
...
+		epsilon,
+	)
+	require.NoError(err)
+
+	topic, err = s.emissionsKeeper.GetTopic(s.ctx, topicId)
+	require.NoError(err)
+	require.NotEqual(topic.InitialRegret, initialRegret)
+}
+
+func (s *InferenceSynthesisTestSuite) TestNotUpdateTopicInitialRegret() {
```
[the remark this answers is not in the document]
spooktheducks: Unresolving because you did literally nothing

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -392,17 +435,64 @@ func CalcTopicInitialRegret(regrets []alloraMath.Dec, epsilon alloraMath.Dec, pN
...
+	numInclusions, err := k.GetCountForecasterInclusionsInTopic(ctx, topic.Id, forecaster)
+	if err != nil {
+		return false, errorsmod.Wrapf(err, "failed to get forecaster inclusions")
+	}
+	return isExperiencedActor(newParticipant, numInclusions, topic.AlphaRegret)
+}
+
+// helper function for isExperiencedForecaster and isExperiencedInferer
+// check if inclusions count is greater than 1/alpha_regret
+func isExperiencedActor(newParticipant bool, numInclusions uint64, alpha alloraMath.Dec) (bool, error) {
```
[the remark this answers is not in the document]
spooktheducks: Unresolving because nothing was done.  Don't resolve things that you've done nothing about.

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -392,17 +435,64 @@ func CalcTopicInitialRegret(regrets []alloraMath.Dec, epsilon alloraMath.Dec, pN
...
+	numInclusions, err := k.GetCountInfererInclusionsInTopic(ctx, topic.Id, inferer)
+	if err != nil {
+		return false, errorsmod.Wrapf(err, "failed to get inferer inclusions")
+	}
+	return isExperiencedActor(newParticipant, numInclusions, topic.AlphaRegret)
+}
+
+// Determine if a forecaster is an experienced worker by checking
+// the inclusions count is greater than 1/alpha_regret
+func isExperiencedForecaster(
```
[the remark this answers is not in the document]
spooktheducks: Unresolving because nothing was done.  Don't resolve things that you've done nothing about.

x/emissions/keeper/inference_synthesis/network_regrets.go
spooktheducks: Didn't do anything about the feedback.  Great chap.  Unresolving.

x/emissions/keeper/inference_synthesis/network_regrets.go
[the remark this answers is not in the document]
spooktheducks: Not in my book.  We're professionally paranoid around these parts

math/utils_test.go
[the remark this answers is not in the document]
spooktheducks: Unresolving because nothing was done.  Don't resolve things that you've done nothing about.

x/emissions/keeper/inference_synthesis/network_regrets.go
spooktheducks: @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso @kpeluso

spooktheducks (changes requested):

spooktheducks:
Not tolerating this kind of response to extremely thorough PR feedback anymore.

x/emissions/keeper/inference_synthesis/network_regrets.go
relyt29: @spooktheducks I convinced RedBird to do it already, it should be in this PR now

math/utils_test.go
relyt29: this was never done @kpeluso (i think)

x/emissions/proto/emissions/v5/tx.proto
relyt29: yes, because we add params and new queries

x/emissions/proto/emissions/v3/genesis.proto
```diff
@@ -219,6 +219,12 @@ message TopicIdActorIdScore {
   Score Score = 3;
 }
 
+message TopicIdActorIdUint64 {
```
relyt29, line 222: I suspect this should not be here and should be only in v5/genesis.proto

it does not make sense why we are editing old protos

someone please justify this to me if it's necessary else lets make sure this is in v5, and if so delete from here

x/emissions/module/rewards/rewards_test.go
```diff
@@ -2861,6 +2861,25 @@ func (s *RewardsTestSuite) TestRewardForTopicGoesUpWhenRelativeStakeGoesUp() {
 	err = s.emissionsAppModule.EndBlock(s.ctx)
 	require.NoError(err)
 
+	worker1InclusionNum, err := s.emissionsKeeper.GetCountInfererInclusionsInTopic(s.ctx, topicId0, s.addrsStr[workerIndexes[0]])
+	require.NoError(err)
+	require.Positive(worker1InclusionNum)
```
relyt29, line 2866: can we do better than just checking it's positive, this seems very lazy

we should know what we expect the number to be during this (and every other rewards) test

x/emissions/module/rewards/rewards.go
relyt29: Narrator: he wrote exactly one test, and it wasn't very thorough

x/emissions/proto/emissions/v3/genesis.proto
kpeluso: removed

x/emissions/keeper/inference_synthesis/network_regrets.go
kpeluso: indeed afaict

math/utils_test.go
kpeluso: I think it is done actually. It doesn't show as "outdated" here because those particular lines are unchanged, but below these lines you see all the extra cases.

kpeluso: Preemptively resolving but feel free to pop open if issue

x/emissions/keeper/inference_synthesis/network_regrets.go
kpeluso: this actually was changed

spooktheducks (approved):

x/emissions/migrations/v5/migrate.go
```diff
@@ -0,0 +1,232 @@
...
+func safelyClearWholeMap(ctx sdk.Context, store storetypes.KVStore, keyPrefix []byte) {
+	s := prefix.NewStore(store, keyPrefix)
+
+	// Loop until all keys are deleted.
+	// Unbounded not best practice but we are sure that the number of keys will be limited
+	// and not deleting all keys means "poison" will remain in the store.
+	for {
+		// Gather keys to eventually delete
+		iterator := s.Iterator(nil, nil)
+		defer iterator.Close()
```
fernandofcampos, line 172: You are closing the iterator twice.

fernandofcampos (changes requested):

fernandofcampos (dismissed):

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -116,6 +118,11 @@ func GetCalcSetNetworkRegrets(args GetCalcSetNetworkRegretsArgs) error {
 
 	workersRegrets := make([]alloraMath.Dec, 0)
 
+	topic, err := args.K.GetTopic(args.Ctx, args.TopicId)
+	if err != nil {
+		return errorsmod.Wrapf(err, "failed to get topic")
+	}
+
```
guilherme-brandao, line 125: You already have the `AlphaRegret` in the args; you don't need to fetch the topic again.

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -136,7 +143,18 @@ func GetCalcSetNetworkRegrets(args GetCalcSetNetworkRegretsArgs) error {
 		if err != nil {
 			return errorsmod.Wrapf(err, "Error setting inferer regret")
 		}
-		if !newParticipant {
+
+		shouldAddWorkerRegret, err := isExperiencedInferer(
+			args.Ctx,
+			args.K,
+			topic,
```
guilherme-brandao, line 150: Pass the `TopicId` and `AlphaRegret` instead of the whole topic

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -161,7 +179,18 @@ func GetCalcSetNetworkRegrets(args GetCalcSetNetworkRegretsArgs) error {
 		if err != nil {
 			return errorsmod.Wrapf(err, "Error setting forecaster regret")
 		}
-		if !newParticipant {
+
+		shouldAddWorkerRegret, err := isExperiencedForecaster(
+			args.Ctx,
+			args.K,
+			topic,
```
guilherme-brandao, line 186: same

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -330,7 +359,18 @@ func GetCalcSetNetworkRegrets(args GetCalcSetNetworkRegretsArgs) error {
 
 	// Recalculate topic initial regret
 	if len(workersRegrets) > 0 {
-		updatedTopicInitialRegret, err := CalcTopicInitialRegret(workersRegrets, args.EpsilonTopic, args.PNorm, args.CNorm)
+		params, err := args.K.GetParams(args.Ctx)
+		if err != nil {
+			return errorsmod.Wrapf(err, "GetCalcSetNetworkRegrets error getting params")
+		}
```
guilherme-brandao, line 365: You can also avoid this query by adding the `InitialRegretQuantile` and `PnormSafeDiv` in the args since we already have the params in `CloseReputerNonce`.

Also, please follow the pattern:
`PnormSafeDiv` -> `PNormSafeDiv`

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -343,11 +383,19 @@ func GetCalcSetNetworkRegrets(args GetCalcSetNetworkRegretsArgs) error {
...
+// dummy_regret[i] = np.percentile(regrets[i-1, :], 25.) + offset * denominator
+// It is assumed that the regrets are filtered by experience for each actor
+// i.e. if they have not been included in the topic for enough epochs, their regret is ignored.
 func CalcTopicInitialRegret(
 	regrets []alloraMath.Dec,
 	epsilon alloraMath.Dec,
 	pNorm alloraMath.Dec,
 	cNorm alloraMath.Dec,
+	quantileRegret alloraMath.Dec,
+	pNormDiv alloraMath.Dec,
```
guilherme-brandao, line 398: `pNormSafeDiv`

x/emissions/keeper/actor_utils/quantile.go
```diff
@@ -16,85 +14,9 @@ func GetQuantileOfScores(
...
-	}
-	ret, err := lowerScore.Score.Add(product)
-	if err != nil {
-		return alloraMath.Dec{}, err
+	decScores := make([]alloraMath.Dec, len(scores))
+	for i, score := range scores {
+		decScores[i] = score.Score
 	}
-	return ret, nil
+	return alloraMath.GetQuantileOfDecs(decScores, quantile)
```
guilherme-brandao, line 21: I think we can remove this file `quantile.go`.

guilherme-brandao (dismissed):

math/utils.go
```diff
@@ -484,3 +485,81 @@ func SumDecSlice(x []Dec) (Dec, error) {
 	}
 	return sum, nil
 }
+
+func GetQuantileOfDecs(
```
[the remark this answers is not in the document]
kpeluso: umm??
image

x/emissions/keeper/inference_synthesis/network_regrets.go
kpeluso: good point!

x/emissions/keeper/inference_synthesis/network_regrets.go
kpeluso: fixed in next commit

x/emissions/keeper/inference_synthesis/network_regrets.go
kpeluso: fixed in next commit

x/emissions/keeper/actor_utils/quantile.go
kpeluso: umm??
image

x/emissions/keeper/inference_synthesis/network_regrets.go
kpeluso: added the first point in my commit!

kpeluso: fixed second in 2nd commit

math/utils.go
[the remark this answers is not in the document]
kpeluso: and
image

math/utils_test.go
relyt29: oh I see, not really labeled in the comment. If it's cool I'm going to push a commit

relyt29: ok I didn't see that

I think while we're trying to write quality code, we can do better than this

1. I would like to see a comment on top of each `data` field dictating what this test case is (empty set, 1 element set, etc etc) - I didn't even realize we had covered the test cases because they blended in with the code. So just adding a comment above each table element labeling which test case it is (alternatively can put the name first so it's the first thing the eye is drawn to, and improve what the name field actually says, because several of these are non-descriptive names "mixed data"- mixed data of what? mixed negative and positive data)
2. The infinity case that diego listed is not covered
3. I believe we need to add a `t.Parallel()` somewhere in here
4. Why is `TestGetQuantileOfDecsWithInvalidQuantile` a separate test, can we just combine it...?

x/emissions/proto/emissions/v5/genesis.proto
```diff
@@ -0,0 +1,388 @@
...
+  // map of (topic, reputer) -> listening coefficient
+  repeated TopicIdActorIdListeningCoefficient reputer_listening_coefficient = 16;
+  // map of (topic, reputer) -> previous reward (used for EMA)
+  repeated TopicIdActorIdDec previous_reputer_reward_fraction = 17;
+  // map of (topic, worker) -> previous reward for inference (used for EMA)
+  repeated TopicIdActorIdDec previous_inference_reward_fraction = 18;
+  // map of (topic, worker) -> previous reward for forecast (used for EMA)
+  repeated TopicIdActorIdDec previous_forecast_reward_fraction = 19;
+  // map of (topic, forecaster) -> ratio of forecaster score
+  repeated TopicIdAndDec previous_forecaster_score_ratio = 20;
```
relyt29, line 67: should close PROTO-2349, but I don't see any corresponding change in genesis.go, did we end up fixing this case?

x/emissions/types/params.go
```diff
@@ -181,6 +183,12 @@ func (p Params) Validate() error {
 	if err := validateMaxStringLength(p.MaxStringLength); err != nil {
 		return errorsmod.Wrap(err, "params validation failure: max string length")
 	}
+	if err := validateInitialRegretQuantile(p.InitialRegretQuantile); err != nil {
+		return errorsmod.Wrap(err, "params validation failure: regret percentile")
```
relyt29, line 187:
```suggestion
		return errorsmod.Wrap(err, "params validation failure: regret quantile")
```

x/emissions/types/params.go
```diff
@@ -568,6 +558,22 @@ func validateMaxStringLength(_ uint64) error {
 	return nil
 }
 
+func validateInitialRegretQuantile(i alloraMath.Dec) error {
```
relyt29, line 561: can we get some nice function header comments, like every other `validate` function in this file has 

AI will autogenerate this for you

ought to add a linter rule to throw if any function in the repo doesn't have function header

x/emissions/migrations/v5/migrate.go
```diff
@@ -0,0 +1,232 @@
...
+	ResetMapsWithNonNumericValues(ctx, store, cdc)
+
+	ctx.Logger().Info("MIGRATING EMISSIONS MODULE FROM VERSION 4 TO VERSION 5 COMPLETE")
+	return nil
+}
+
+// migrate params for this new version
+// the changes are the addition of InitialRegretQuantile,PNormSafeDiv
+func MigrateParams(store storetypes.KVStore, cdc codec.BinaryCodec) error {
+	oldParams := oldV4Types.Params{} //nolint: exhaustruct
```
relyt29, line 55:
```suggestion
	oldParams := oldV4Types.Params{} //nolint: exhaustruct // empty struct used by cosmos-sdk Unmarshal below
```

x/emissions/migrations/v5/migrate_test.go
```diff
@@ -0,0 +1,319 @@
...
+	s.Require().Equal(paramsExpected.MaxActiveTopicsPerBlock, params.MaxActiveTopicsPerBlock)
+	s.Require().Equal(paramsExpected.MaxStringLength, params.MaxStringLength)
+	s.Require().Equal(paramsExpected.InitialRegretQuantile, params.InitialRegretQuantile)
+	s.Require().Equal(paramsExpected.PNormSafeDiv, params.PNormSafeDiv)
+	s.Require().Equal(paramsExpected, params)
+}
+
+// in this test, we check that an already migrated topic, that only initialRegret
+// set to 0 but everything else will remain the same
+func (s *EmissionsV5MigrationTestSuite) TestMigratedTopicWithNaNInitialRegret() {
```
relyt29, line 185: I thought we had gotten rid of NaNs by now...?

relyt29 (dismissed):
Getting pretty close to ready for merge

mostly stylistic stuff and a few questions

thank you for your hard work, this is a big PR

x/emissions/migrations/v5/migrate_test.go
kpeluso: It's re-using the same functions as before because we still care to clear out fields. We're not clearing out NaNs necessarily but near-infinitely negative regrets that have developed, which shouldn't happen again once this PR is merged.

x/emissions/proto/emissions/v5/genesis.proto
kpeluso: hmm i did update both of the `genesis.go` files...

x/emissions/keeper/actor_utils/quantile.go
kpeluso: updated in upcoming commit

math/utils.go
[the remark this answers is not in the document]
kpeluso: updated in upcoming commit

fernandofcampos (approved):

guilherme-brandao (approved):

PR #620, opened 2024-09-30
Non-deferred close checker

spooktheducks:
Run this on CI.  Catches non-deferred `.Close()` calls, understands deferred func literals containing simple `.Close()` calls as well as simple `defer` statements


relyt29:
This is great, few things:

1. Nothing exists in golang-ci for this very case? I'd prefer to push the work of maintaining such a script to golang-ci people
2. This code is great but doesn't actually add it to our CI, shouldn't we do that in this PR here

kpeluso:
Should add to `makefile` as well

relyt29:
PLEASE REBASE THIS BRANCH ONTO DEV AGAIN, AS DEV HAS BEEN PUSHED TO A NEW HEAD THAT REMOVES A BAD COMMIT
```
git fetch 
git rebase -i origin/dev
```
DO NOT UPDATE IN GITHUB UI, YOU WILL REINTRODUCE THE BAD COMMIT IN YOUR MERGE COMMIT - I REPEAT DO NOT MERGE

@spooktheducks

x/utils/migutils/migutils.go
```diff
@@ -0,0 +1,57 @@
+package migutils
+
+import (
+	errorsmod "cosmossdk.io/errors"
+	"cosmossdk.io/store/prefix"
+	storetypes "cosmossdk.io/store/types"
+)
+
+// Deletes all keys in the store with the given keyPrefix `maxPageSize` keys at a time
+func SafelyClearWholeMap(store storetypes.KVStore, keyPrefix []byte, maxPageSize uint64) error {
```
relyt29, line 10: should this have a test to make sure it works as expected?

x/utils/fn/fn.go
```diff
@@ -0,0 +1,307 @@
+package fn
```
relyt29, line 1: I don't believe this package should live under the /x/ folder since that's cosmos-sdk nomenclature for cosmos-sdk modules, which this is not

Also none of these functions have tests?

also can we call the package allorautils or something, `fn` is such a short non descriptive name

also maybe this and the migutils should go in a separate PR so this PR is more tightly scoped to just a change about non-deferred close checking

relyt29 (changes requested):

x/utils/fn/fn.go
kpeluso, line 1: what tyler said. also I'm generally opposed to adding in unused code, if for no other reason other than audits get more expensive for no reaason

x/utils/migutils/migutils.go
kpeluso: +1

x/utils/migutils/migutils.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// Loop until all keys are deleted.
+	// Unbounded not best practice but we are sure that the number of keys will be limited
+	// and not deleting all keys means "poison" will remain in the store.
+	for {
+		more, err := clearPage()
+		if err != nil {
+			return err
+		}
+		if !more {
```
kpeluso, line 52: thought you were an `else if` guy in this situation?

x/utils/migutils/migutils.go
kpeluso: Safer than before for sure tho!

kpeluso (commented):
Basically agree w @relyt29 overall. Will recheck with those comments resolved.

x/utils/migutils/migutils.go
spooktheducks: 99% of the time -- this was a rare case where i just felt like it was a little clearer.  waffled back and forth on it for a few minutes

x/utils/fn/fn.go
spooktheducks: removed, but really would like to keep `migutils` in here since the code didn't actually change at all -- it's just ported verbatim to reduce duplication

relyt29 (approved):
looks good to me. Additional question not needed to be dealt with in this PR: we have the `linter/maprange.go` custom linter - where is that being run, and can we unify how our custom linters are run?

kpeluso (approved):

PR #621, opened 2024-10-03
resolve exhaustruct on synth while preserving builder

kpeluso:

This is my alternative to https://github.com/allora-network/allora-chain/pull/538

I have absolutely been burned before for having too many arg params. I've fixed many bugs directly resulting from this. The pattern employed there is really not great in my opinion.

In this PR I ready synth for exhaustruct while preserving the patterns.

## Still Left Todo

* I have not applied exhaustruct remediation to tests
* Testing and remediation of rest of code for exhaustruct (though perhaps for another PR)
* Other improvements and any bug fixes @relyt29 identified in https://github.com/allora-network/allora-chain/pull/538 should be included here
* The optimization @guilherme-brandao identified in our chats (to remove duplicate valiations and clone operation) could neatly be included here as well

### DIFF
--- a/x/emissions/keeper/inference_synthesis/network_inference_builder.go
+++ b/x/emissions/keeper/inference_synthesis/network_inference_builder.go
@@ -15,6 +15,7 @@ type NetworkInferenceBuilder struct {
 	logger  log.Logger
 	palette SynthPalette
 	// Network Inferences Properties
+	nonce                      *emissions.Nonce
 	inferences                 []*emissions.WorkerAttributedValue
 	forecastImpliedInferences  []*emissions.WorkerAttributedValue
 	weights                    RegretInformedWeights
@@ -34,9 +35,18 @@ func NewNetworkInferenceBuilderFromSynthRequest(
 		return nil, errorsmod.Wrapf(err, "Error building palette from request")
 	}
 	return &NetworkInferenceBuilder{
-		ctx:     req.Ctx,
-		logger:  Logger(req.Ctx),
-		palette: palette,
+		ctx:                        req.Ctx,
+		logger:                     Logger(req.Ctx),
+		palette:                    palette,
+		nonce:                      req.Nonce,
+		inferences:                 []*emissions.WorkerAttributedValue{},
+		forecastImpliedInferences:  []*emissions.WorkerAttributedValue{},
+		weights:                    RegretInformedWeights{inferers: nil, forecasters: nil},
+		combinedInference:          InferenceValue{},
+		naiveInference:             InferenceValue{},
+		oneOutInfererInferences:    []*emissions.WithheldWorkerAttributedValue{},
+		oneOutForecasterInferences: []*emissions.WithheldWorkerAttributedValue{},
+		oneInInferences:            []*emissions.WorkerAttributedValue{},
 	}, nil
 }
 
@@ -394,13 +404,17 @@ func (b *NetworkInferenceBuilder) CalcAndSetNetworkInferences() *NetworkInferenc
 func (b *NetworkInferenceBuilder) Build() *emissions.ValueBundle {
 	// Build value bundle to return all the calculated inferences
 	return &emissions.ValueBundle{
-		TopicId:                b.palette.TopicId,
-		CombinedValue:          b.combinedInference,
-		InfererValues:          b.inferences,
-		ForecasterValues:       b.forecastImpliedInferences,
-		NaiveValue:             b.naiveInference,
-		OneOutInfererValues:    b.oneOutInfererInferences,
-		OneOutForecasterValues: b.oneOutForecasterInferences,
-		OneInForecasterValues:  b.oneInInferences,
+		TopicId:                       b.palette.TopicId,
+		Reputer:                       "",
+		ReputerRequestNonce:           &emissions.ReputerRequestNonce{ReputerNonce: b.nonce},
+		CombinedValue:                 b.combinedInference,
+		InfererValues:                 b.inferences,
+		ForecasterValues:              b.forecastImpliedInferences,
+		NaiveValue:                    b.naiveInference,
+		OneOutInfererValues:           b.oneOutInfererInferences,
+		OneOutForecasterValues:        b.oneOutForecasterInferences,
+		OneInForecasterValues:         b.oneInInferences,
+		OneOutInfererForecasterValues: []*emissions.OneOutInfererForecasterValues{},
+		ExtraData:                     make([]byte, 0),
 	}
 }
--- a/x/emissions/keeper/inference_synthesis/network_inferences.go
+++ b/x/emissions/keeper/inference_synthesis/network_inferences.go
@@ -51,11 +51,9 @@ func GetNetworkInferences(
 		}
 	}
 
-	networkInferences := &emissions.ValueBundle{
-		TopicId:          topicId,
-		InfererValues:    make([]*emissions.WorkerAttributedValue, 0),
-		ForecasterValues: make([]*emissions.WorkerAttributedValue, 0),
-	}
+	val := GenerateEmptyValueBundle()
+	networkInferences := &val
+	networkInferences.TopicId = topicId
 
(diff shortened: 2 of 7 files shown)


relyt29:
This proposal is effectively no change on inference synthesis, and keeps all of the parts of inference synthesis that are written completely differently than any other part of the codebase (nowhere else do we use this factory builder object oriented pattern, nowhere else has a palette).

If I look at a function today in the inference synthesis package I have *no idea* what specific parts of the synth palette it operates on from looking at the function header because the synth palette and network inference builder structs hide the actual elements used in the function.

This, paired with lack of exhaust struct, has caused bugs that affect the accuracy of inferences. This also has affected performance due to clones and passing of unnecessary values to every function rather than just what the function needs

kpeluso:
@relyt29 
* Much of the rest of the codebase would benefit from more consistency in its code organization
   * This pattern could apply to rewards too, albeit with less benefit imo
* When we update state with any of the keeper's setters, we also technically don't know how much of the state is actually updated
   * ...if it were not for clear comments and function names. So maybe it's just a matter of that
   * The thing with these synth functions is that they basically all take the same set of args anyway, so when we know when we know the others, and safer to keep all props properly updated. That ends up being a benefit to this code because if one function doesn't directly use every method, it often happens that a function called in its scope still does
* The thing with the "cloning vs passing it in" is that when we're passing in by value, as in your PR, we have to realloc the memory for each new scope it's passed to anyway, which is the same complexity as cloning, as we do here. So I'm actually skeptical of the optimization you found, and confident that if you did find one, we could implement it under this framework all the same.

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -34,9 +35,18 @@ func NewNetworkInferenceBuilderFromSynthRequest(
...
 	}
 	return &NetworkInferenceBuilder{
-		ctx:     req.Ctx,
-		logger:  Logger(req.Ctx),
-		palette: palette,
+		ctx:                        req.Ctx,
+		logger:                     Logger(req.Ctx),
+		palette:                    palette,
+		nonce:                      req.Nonce,
+		inferences:                 []*emissions.WorkerAttributedValue{},
```
spooktheducks, line 42: Fwiw, the zero value of a slice is `nil`, which will still satisfy `exhauststruct`.  If you do it your way, it'll allocate some memory/`cap`that you might not use.  I'd advise waiting to `make(...)` slices like these until you need them and maybe even have the benefit of knowing how large to make them

Same goes for maps.  Zero value is `nil`

This is because slices are pointers to a struct under the hood, as are maps and channels.  Everything else is pretty much what it appears to be

x/emissions/keeper/inference_synthesis/network_inferences.go
```diff
@@ -141,23 +140,36 @@ func GetNetworkInferences(
...
-
-		networkInferences = &emissions.ValueBundle{
-			TopicId:       topicId,
-			CombinedValue: singleInference.Value,
-			InfererValues: []*emissions.WorkerAttributedValue{
-				{
-					Worker: singleInference.Inferer,
-					Value:  singleInference.Value,
-				},
+		networkInferencesVal := GenerateEmptyValueBundle()
```
spooktheducks, line 143: Making a zero-value struct constructor is much simpler this way:
```go
var x emissions.ValueBundle
x.foo = ...
```
No constructor function needed.

Deeper point-- making a zero-value constructor that passes exhauststruct but then assigns individual values seems to be avoiding the point of exhauststruct.

I think we're going for more semantic clarity.  `yield` @relyt29 right?

x/emissions/keeper/inference_synthesis/network_inferences.go
```diff
@@ -141,23 +140,36 @@ func GetNetworkInferences(
...
-			OneOutForecasterValues: []*emissions.WithheldWorkerAttributedValue{},
-			OneInForecasterValues:  []*emissions.WorkerAttributedValue{},
 		}
+		networkInferences = &networkInferencesVal
 	}
 
 	return networkInferences, forecastImpliedInferencesByWorker, infererWeights, forecasterWeights, inferenceBlockHeight, lossBlockHeight, nil
 }
+
+func GenerateEmptyValueBundle() emissions.ValueBundle {
```
spooktheducks, line 160: This reduces to
```go
var x emissions.ValueBundle
return x
```

x/emissions/keeper/inference_synthesis/network_losses.go
```diff
@@ -263,6 +263,7 @@ func CalcNetworkLosses(
 		OneOutInfererValues:           oneOutInfererLosses,
 		OneOutForecasterValues:        oneOutForecasterLosses,
 		OneInForecasterValues:         oneInForecasterLosses,
+		ExtraData:                     make([]byte, 0),
```
spooktheducks, line 266: The true zero value of a slice is `nil`

x/emissions/keeper/inference_synthesis/synth_palette_forecast_implied.go
```diff
@@ -61,8 +61,12 @@ func (p *SynthPalette) CalcForecastImpliedInferences() (map[Worker]*emissionstyp
 				}
 
 				forecastImpliedInference := emissionstypes.Inference{
-					Inferer: forecaster,
-					Value:   medianValue,
+					Inferer:     forecaster,
+					Value:       medianValue,
+					TopicId:     p.TopicId,
+					BlockHeight: p.Nonce.BlockHeight,
+					ExtraData:   make([]byte, 0),
```
spooktheducks, line 68: nil

x/emissions/keeper/inference_synthesis/synth_palette_forecast_implied.go
```diff
@@ -127,8 +131,12 @@ func (p *SynthPalette) CalcForecastImpliedInferences() (map[Worker]*emissionstyp
 						return nil, errorsmod.Wrapf(err, "error calculating forecast value")
 					}
 					forecastImpliedInference := emissionstypes.Inference{
-						Inferer: forecaster,
-						Value:   forecastValue,
+						Inferer:     forecaster,
+						Value:       forecastValue,
+						TopicId:     p.TopicId,
+						BlockHeight: p.Nonce.BlockHeight,
+						ExtraData:   make([]byte, 0),
```
spooktheducks, line 138: nil

spooktheducks (changes requested):

spooktheducks:
> * The thing with the "cloning vs passing it in" is that when we're passing in by value, as in your PR, we have to realloc the memory for each new scope it's passed to anyway, which is the same complexity as cloning, as we do here. So I'm actually skeptical of the optimization you found, and confident that if you did find one, we could implement it under this framework all the same.

Ensuring data immutability in Go is pretty simple -- here are the rules:
1) If your struct doesn't contain pointers, maps, slices, or channels, just pass them by value (those types are all pointers to structs)
2) You can't clone a channel.  That's completely insane and would create ruptures in spacetime
3) One can clone slices and maps with the respective stdlib `slices` and `maps` packages, or just write `for` loops if one has the muscle mass to support it
4) If you have pointers to structs, make sure you deference them when cloning, like this:
```go
childStructToClone := *x.foo
clonedStruct.foo = &childStructToClone
```

Therefore, if you're writing a `.Clone()` method:
1) use a non-pointer receiver so all of the fields are cloned automatically
2) any of the fields that are pointers, maps, or slices -- clone those manually

It tends to only be a few lines of code when you have to write a method for this.

x/emissions/keeper/inference_synthesis/network_inferences.go
relyt29: correct, I actually don't mind the pattern usually but given golang's inflexibility / immaturity of linters to properly handle this case, I think we should force ourselves to just never use this pattern and instead create local variables and then assign them once the full set of all fields is prepared

kpeluso:
Closing in favor of https://github.com/allora-network/allora-chain/pull/538

PR #622, opened 2024-10-03
Add metrics on tx/query

RedBird96:

Add counter, gauges metrics using telemetry  on every query/txs

## Still Left Todo

### DIFF
--- a/x/emissions/keeper/msgserver/msg_server_demand.go
+++ b/x/emissions/keeper/msgserver/msg_server_demand.go
@@ -9,8 +9,10 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-func (ms msgServer) FundTopic(ctx context.Context, msg *types.FundTopicRequest) (*types.FundTopicResponse, error) {
-	err := ms.k.ValidateStringIsBech32(msg.Sender)
+func (ms msgServer) FundTopic(ctx context.Context, msg *types.FundTopicRequest) (_ *types.FundTopicResponse, err error) {
+	defer metrics.RecordMetrics("FundTopic", time.Now(), func() bool { return err == nil })
+
+	err = ms.k.ValidateStringIsBech32(msg.Sender)
 	if err != nil {
 		return nil, err
 	}
--- a/x/emissions/keeper/msgserver/msg_server_demand.go
+++ b/x/emissions/keeper/msgserver/msg_server_demand.go
@@ -2,6 +2,7 @@ package msgserver
 
 import (
 	"context"
+	"time"
 
 	"github.com/allora-network/allora-chain/x/emissions/types"
 	"google.golang.org/grpc/codes"
--- a/x/emissions/keeper/msgserver/msg_server_demand.go
+++ b/x/emissions/keeper/msgserver/msg_server_demand.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"time"
 
+	"github.com/allora-network/allora-chain/x/emissions/metrics"
 	"github.com/allora-network/allora-chain/x/emissions/types"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
--- a/x/emissions/keeper/msgserver/msg_server_demand.go
+++ b/x/emissions/keeper/msgserver/msg_server_demand.go
@@ -11,7 +11,7 @@ import (
 )
 
 func (ms msgServer) FundTopic(ctx context.Context, msg *types.FundTopicRequest) (_ *types.FundTopicResponse, err error) {
-	defer metrics.RecordMetrics("FundTopic", time.Now(), func() bool { return err == nil })
+	defer metrics.RecordMetrics("FundTopic", time.Now(), &err)
 
 	err = ms.k.ValidateStringIsBech32(msg.Sender)
 	if err != nil {
--- a/x/emissions/metrics/labels.go
+++ b/x/emissions/metrics/labels.go
@@ -1,14 +1,6 @@
 package metrics
 
 const (
-	CONCURRENT     = "concurrent"
-	SYNCHRONOUS    = "synchronous"
-	OCC_CONCURRENT = "optimistic_concurrency"
-	GovMsgInBlock  = "gov-msg-in-block"
-	FailedToBuild  = "failed-to-build"
-)
-
-const (
 	INFERER_SCORE_EVENT       = "inferer_score_event"
 	FORECASTER_SCORE_EVENT    = "forecaster_score_event"
 	REPUTER_SOCRE_EVENT       = "reputer_score_event"
--- a/x/emissions/metrics/labels.go
+++ b/x/emissions/metrics/labels.go
@@ -1,0 +1,24 @@
+package metrics
+
+const (
+	CONCURRENT     = "concurrent"
+	SYNCHRONOUS    = "synchronous"
+	OCC_CONCURRENT = "optimistic_concurrency"
+	GovMsgInBlock  = "gov-msg-in-block"
+	FailedToBuild  = "failed-to-build"
+)
+
+const (
+	INFERER_SCORE_EVENT       = "inferer_score_event"
+	FORECASTER_SCORE_EVENT    = "forecaster_score_event"
+	REPUTER_SOCRE_EVENT       = "reputer_score_event"
+	NETWORK_LOSS_EVENT        = "network_loss_event"
+	INFERER_REWARD_EVENT      = "inferer_reward_event"
+	FORECASTER_REWARD_EVENT   = "forecaster_reward_event"
+	REPUTER_REWARD_EVENT      = "reputer_reward_event"
+	WORKER_LAST_COMMIT_EVENT  = "worker_last_commit_event"
+	REPUTER_LAST_COMMIT_EVENT = "reputer_last_commit_event"
+	FORECAST_TASK_SCORE_EVENT = "forecast_task_score_event"
+	TOPIC_REWARD_EVENT        = "topic_reward_event"
+	WORKER_EMA_SCORE_EVENT    = "worker_ema_score_event"
+)
--- a/x/emissions/metrics/metrics_util.go
+++ b/x/emissions/metrics/metrics_util.go
@@ -1,3 +1,5 @@
+// This is copied from the sei-chain
+// source: https://github.com/sei-protocol/sei-chain/blob/main/utils/metrics/metrics_util.go
 package metrics
 
 import (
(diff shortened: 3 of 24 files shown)


x/emissions/metrics/metrics_util.go
```diff
@@ -0,0 +1,199 @@
...
+	)
+}
+
+// allora_epoch_new
+func SetEpochNew(epochNum uint64) {
+	metrics.SetGauge(
+		[]string{"allora", "epoch", "new"},
+		float32(epochNum),
+	)
+}
```
guilherme-brandao, line 90: I didn't understand the purpose of this one. I guess we can remove it.

x/emissions/metrics/metrics_util.go
```diff
@@ -0,0 +1,199 @@
...
+// Metric Name:
+//
+//	allora_websocket_connect
+func IncWebsocketConnects() {
+	telemetry.IncrCounterWithLabels(
+		[]string{"allora", "websocket", "connect"},
+		1,
+		nil,
+	)
+}
```
guilherme-brandao, line 102: Are we using this?

guilherme-brandao: Please remove what we are not using

x/emissions/types/events.go
```diff
@@ -22,6 +24,7 @@ func EmitNewForecasterScoresSetEvent(ctx sdk.Context, scores []Score) {
 	if len(scores) < 1 {
 		return
 	}
+	metrics.IncrProducerEventCount(metrics.FORECASTER_SCORE_EVENT)
 	err := ctx.EventManager().EmitTypedEvent(NewScoresSetEventBase(ActorType_ACTOR_TYPE_FORECASTER, scores))
```
guilherme-brandao, line 28: I don't think we need to keep both... cc @kpeluso

guilherme-brandao (dismissed):

x/emissions/metrics/metrics_util.go
```diff
@@ -0,0 +1,199 @@
...
+//	allora_loadtest_produce_count
+func IncrProducerEventCount(msgType string) {
+	telemetry.IncrCounterWithLabels(
+		[]string{"allora", "loadtest", "produce", "count"},
+		1,
+		[]metrics.Label{telemetry.NewLabel("msg_type", msgType)},
+	)
+}
+
+func RecordMetrics(apiMethod string, connectionType string, startTime time.Time, success bool) {
```
kpeluso, line 196: I don't think the `connectionType` is never not `"rpc"` in your PR here, nor do I think that this code knows if it's being invoked via RPC or API.

=> I'd remove it

If it is kept, I'd make it an enum. I don't know what the other values would be though because of what I just said

x/emissions/metrics/labels.go
```diff
@@ -0,0 +1,24 @@
+package metrics
+
+const (
+	CONCURRENT     = "concurrent"
+	SYNCHRONOUS    = "synchronous"
+	OCC_CONCURRENT = "optimistic_concurrency"
+	GovMsgInBlock  = "gov-msg-in-block"
+	FailedToBuild  = "failed-to-build"
+)
```
kpeluso, line 9: are these used?

x/emissions/metrics/labels.go
```diff
@@ -0,0 +1,24 @@
...
+
+const (
```
kpeluso, line 11: are these used?

x/emissions/metrics/metrics_util.go
```diff
@@ -0,0 +1,199 @@
...
+	)
+}
+
+// Measures the RPC request latency in milliseconds
+// Metric Name:
+//
+//	allora_rpc_request_latency_ms
+func MeasureRpcRequestLatency(endpoint string, connectionType string, startTime time.Time) {
+	metrics.MeasureSinceWithLabels(
+		[]string{"allora", "rpc", "request", "latency_ms"},
```
kpeluso, line 174: for all these functions, if `connectionType` is kept, would you not be passing it into `[]string{...}` instead of  `"rpc"` ?

x/emissions/metrics/metrics_util.go
```diff
@@ -0,0 +1,199 @@
+package metrics
```
kpeluso, line 1: not the biggest fan of adding code we don't use or need. at best, we pay more for audits

ideally we keep and use more of them where it makes sense asap, but perhaps beyond this PR

kpeluso, line 1: if this was all pulled from somewhere, you should add comments of linksk to all sources for future devs so they can better understand the context and provenance of the design decisions used here (and where it was borrowed from). Also, attribution is important for egos and sometimes legal reasons

kpeluso (changes requested):
almost

ideally we can use more of those metrics too. maybe others can contribute ideas as to how @xmariachi @guilherme-brandao @relyt29 @spooktheducks @fernandofcampos

x/emissions/types/events.go
kpeluso: I'm not opposed to what @RedBird96 is doing here actually -- this provides a decent sanity check for any indexer nop

x/emissions/keeper/msgserver/msg_server_demand.go
```diff
@@ -2,13 +2,21 @@ package msgserver
...
+	"github.com/allora-network/allora-chain/x/emissions/metrics"
 
 	"github.com/allora-network/allora-chain/x/emissions/types"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
 
-func (ms msgServer) FundTopic(ctx context.Context, msg *types.FundTopicRequest) (*types.FundTopicResponse, error) {
+func (ms msgServer) FundTopic(ctx context.Context, msg *types.FundTopicRequest,
+) (
```
spooktheducks, line 15: I know this is a nitpick, but please don't format things like this.  Either

`func foobar(a string, b int) (float, bool)`
or
```
func foobar(
    a string,
    b int,
) (float, bool)
```

x/emissions/keeper/msgserver/msg_server_demand.go
```diff
@@ -2,13 +2,21 @@ package msgserver
...
+	_ *types.FundTopicResponse,
+	returnErr error,
```
spooktheducks, line 17: These errors will always be `nil` in your `defer` calls.  So you'll need to rework the whole PR unfortunately.

The reason: `defer` function arguments (and receivers, fwiw) are evaluated at the point where the `defer` call is first encountered in the code, and *not* when the deferred function actually runs (i.e., when leaving the outer function due to `panic` or `return`).

Check it out: https://go.dev/play/p/MQOXA_m1Z61 -- it's `nil`.  Same pattern, and the error is shown as `nil` by the `defer` function, but non-`nil` by the parent function.  This is because on the line where the `defer` happens, `err` is definitely `nil`.  It doesn't matter what `err` is when the function returns/panics.

However, you can still do what you're trying to do.  Here are two options:
1) by using a reference/pointer: https://go.dev/play/p/kQFWMC5BqTu
2) by using a closure: https://go.dev/play/p/EguVnE0g0CM

I propose it as a useful exercise for you to try to develop a way of expressing this logic that is correct, and also feels nice to use.  You'll learn a bit about the language in doing so.

Explaining why this behavior of the language is such a shitshow is probably out of scope for this review, but maybe we can have a quick internal call with everybody to talk about it.  This is not a fun little corner of Go!  It's complete trash!  So it's best to rely on patterns and an internal `utils` library for this, and maybe even a custom linter.  Very easy to get it wrong.  Even the most senior devs get it wrong all the time.

x/emissions/metrics/metrics_util.go
spooktheducks: @kpeluso can you own or hand off looking into the licensing here?  who is owning this PR atm?

x/emissions/types/events.go
```diff
@@ -69,34 +76,39 @@ func EmitNewReputerAndDelegatorRewardsSettledEvent(ctx sdk.Context, blockHeight
 	if len(rewards) < 1 {
 		return
 	}
+	metrics.IncrProducerEventCount(metrics.REPUTER_REWARD_EVENT)
```
spooktheducks, line 79: These really should match the function name right?

spooktheducks (changes requested):
1) There's a fundamental problem with all of the `defer` calls.  The PR actually buries the lede even deeper as far as I can tell.
2) Also, there are no tests.  Please write tests.  I will block 99% of PRs moving forward that do not have reasonable tests.

x/emissions/keeper/msgserver/msg_server_demand.go
spooktheducks: I'll give you a hint, btw.  Here's a util func I carry around to every project I work on:
```go
func AddStack(err *error) {
	if *err != nil {
		*err = errors.WithStack(*err)
	}
}
```
To use it:
```go
func blah() (err error) {
    defer utils.AddStack(&err)
    
    err = doThing()
    if err != nil {
        return err
    }
}
```

x/emissions/metrics/metrics_util.go
```diff
@@ -0,0 +1,58 @@
...
+//	allora_loadtest_produce_count
+func IncrProducerEventCount(msgType string) {
+	telemetry.IncrCounterWithLabels(
+		[]string{"allora", "loadtest", "produce", "count"},
+		1,
+		[]metrics.Label{telemetry.NewLabel("msg_type", msgType)},
+	)
+}
+
+func RecordMetrics(apiMethod string, startTime time.Time, success func() bool) {
```
fernandofcampos, line 55: All cases are passing the same success function: "func() bool { return err == nil }"

Maybe we can have another version of this that takes err as a parameter and checks it, to simplify usage.

xmariachi (commented):
Reviewed, agree with the intention and the impl looks sound in general.
I lack context to approve though; and yes, testing in the new code could be added.

x/emissions/metrics/metrics_util.go
kpeluso: and wraps the defer and time.now() calls too

kpeluso (dismissed):

x/emissions/metrics/metrics_util.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/cosmos/cosmos-sdk/telemetry"
+	metrics "github.com/hashicorp/go-metrics"
+)
+
+// Measures RPC endpoint request throughput
+// Metric Name:
+//
+//	allora_rpc_request_counter
+func IncrementRpcRequestCounter(endpoint string, success bool) {
```
spooktheducks, line 17:
```suggestion
func IncrementRpcRequestCounter(endpoint string, err *error) {
    success := *err == nil
```

x/emissions/keeper/msgserver/msg_server_demand.go
```diff
@@ -2,14 +2,18 @@ package msgserver
...
+	"github.com/allora-network/allora-chain/x/emissions/metrics"
 	"github.com/allora-network/allora-chain/x/emissions/types"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
 
-func (ms msgServer) FundTopic(ctx context.Context, msg *types.FundTopicRequest) (*types.FundTopicResponse, error) {
-	err := ms.k.ValidateStringIsBech32(msg.Sender)
+func (ms msgServer) FundTopic(ctx context.Context, msg *types.FundTopicRequest) (_ *types.FundTopicResponse, err error) {
+	defer metrics.RecordMetrics("FundTopic", time.Now(), func() bool { return err == nil })
```
spooktheducks, line 14:
```suggestion
	defer metrics.RecordMetrics("FundTopic", time.Now(), &err)
```

kpeluso (approved):

spooktheducks (approved):

PR #623, opened 2024-10-04
Insert Worker and Reputer Payload Refactor

guilherme-brandao:
In the latest update with the addition of merit-based sortition, the complexity of `InsertWorkerPayload` has increased significantly, and it's one of our most used methods (if not the most used).

This PR aims to improve the performance of how we are processing the insertion of new inferences, removing redundant persists, optimizing data structures, etc.

In this new implementation, we created new KV Stores:
- `LowestInfererScoreEma`, `LowestForecasterScoreEma`, and `LowestReputerScoreEma` - lowest score between the active actors
- `ActiveInferers`, `ActiveForecasters`, and `ActiveReputers` - actual active actors which the submission will be rewarded
- `reputerLosses`- individual losses submitted by each reputer

New insert flow overview:
- Instead of using all actual losses/inferences/forecasts submitted to find the lowest score among included actors, we store an Active<ActorType> to identify which addresses are currently active for the topic. This means that the worker/reputer/forecaster has submitted a valid bundle/inference/forecast and has a high enough score to be considered active.
- The lowest score is updated if a new actor with a higher score than the current lowest successfully submits losses/inferences/forecasts.
- During the closing of nonces (`CloseWorkerNonce` and `CloseReputerNonce`), we select all contributions from the active actors and build the set of losses/inferences/forecasts that will be rewarded. We also reset the Active<ActorType> KVStore for the next reward cycle.

This approach comes with the following benefits:
- We don't need to decode the entire set of losses/inferences/forecasts and retrieve all scores to find the lowest score every time a new actor submits.
- We can easily reject submissions by comparing the score of the actor submitting the contribution with the lowest score.
- We eliminate redundant data persistence that was previously occurring.
- Lowest scores are easily retrieved using the KVStore Get method.
- We can add and remove actors in the active set more efficiently, without needing to handle a large array of objects. Now we just need to `Add` or `Remove` from the KVStore.

https://linear.app/alloralabs/issue/PROTO-2528/optimize-insertworkerpayload-transaction

## Still Left Todo
- Implement new tests


x/emissions/types/keys.go
```diff
@@ -86,4 +86,8 @@ var (
 	PreviousTopicQuantileInfererScoreEmaKey           = collections.NewPrefix(71)
 	PreviousTopicQuantileForecasterScoreEmaKey        = collections.NewPrefix(72)
 	PreviousTopicQuantileReputerScoreEmaKey           = collections.NewPrefix(73)
+	QualifiedInferersKey                              = collections.NewPrefix(74)
```
xmariachi, line 89: I think the name should be "Active", not "Qualified", throughout the codebase, to keep consistency.

x/emissions/keeper/keeper.go
```diff
@@ -1044,24 +1070,27 @@ func (k *Keeper) AppendInference(
 }
 
 // Insert a complete set of inferences for a topic/block. Overwrites previous ones.
```
xmariachi, line 1072: This comment is now outdated, right? Belongs to the next one below "insertActiveInferences" maybe?

x/emissions/keeper/actor_utils/worker.go
```diff
@@ -103,43 +103,27 @@ func CloseWorkerNonce(k *keeper.Keeper, ctx sdk.Context, topicId keeper.TopicId,
 	return nil
 }
 
-// It is assumed `inferences` come from unique, registered, top inferers by EMA score descending
-// It is also assumed that the inferences are for the correct topic and nonce
-func insertInferencesFromTopInferers(
+func getAndSetQualifiedInferences(
```
xmariachi, line 106: name suggestion: `closeActiveInferencesSet`

Also worth adding a descriptive comment ;)

x/emissions/keeper/keeper.go
```diff
@@ -1004,38 +1018,50 @@ func (k *Keeper) AppendInference(
...
+		// Add new qualified inferer
+		err = k.AddQualifiedInferer(ctx, topic.Id, inference.Inferer)
+		if err != nil {
+			return errorsmod.Wrap(err, "error adding qualified inferer")
+		}
+		// Calculate new lowest score with updated infererAddresses
+		err = UpdateLowestScoreFromInfererAddresses(ctx, k, topic.Id, workerAddresses, inference.Inferer, lowestEmaScore.Address)
+		if err != nil {
+			return errorsmod.Wrap(err, "error getting low score from all inferences")
+		}
```
xmariachi, line 1060: Just to confirm, the strategy is to keep in `inferences` all of the actors that have ever been active in this block, and then perform a selection of them at closing time of the nonce (`closeWorkerNonce`).
This strategy needs both actions to be in sync and so I think it should be documented.

x/emissions/keeper/keeper.go
```diff
@@ -3410,3 +3439,125 @@ func (k *Keeper) GetPreviousForecasterScoreRatio(ctx context.Context, topicId To
 	}
 	return forecastTau, nil
 }
+
+// AddQualifiedInferer adds an inferer to the qualified inferers set for a topic
+func (k *Keeper) AddQualifiedInferer(ctx context.Context, topicId TopicId, inferer ActorId) error {
+	if err := types.ValidateTopicId(topicId); err != nil {
```
xmariachi, line 3445: This also should check the MaxINferersToReward

x/emissions/keeper/keeper.go
```diff
@@ -3410,3 +3439,125 @@ func (k *Keeper) GetPreviousForecasterScoreRatio(ctx context.Context, topicId To
...
+	})
+	if err != nil {
+		return nil, errorsmod.Wrap(err, "error walking qualified inferers")
+	}
+	return inferers, nil
+}
+
+// AddQualifiedForecaster adds a forecaster to the qualified forecasters set for a topic
+func (k *Keeper) AddQualifiedForecaster(ctx context.Context, topicId TopicId, forecaster ActorId) error {
+	if err := types.ValidateTopicId(topicId); err != nil {
```
xmariachi, line 3483: This also should check the MaxForecastersToReward

x/emissions/keeper/msgserver/msg_server_worker_payload.go
```diff
@@ -16,41 +16,40 @@ import (
...
 	err := ms.k.ValidateStringIsBech32(msg.Sender)
 	if err != nil {
 		return nil, err
 	}
-	err = checkInputLength(ctx, ms, msg)
-	if err != nil {
-		return nil, err
-	}
 
+	blockHeight := sdkCtx.BlockHeight()
```
xmariachi, line 24: good catch

xmariachi (changes requested):
Several changes requested, but agree on the direction of the improvements.

x/emissions/keeper/keeper.go
guilherme-brandao: This is already checked in the unique function that calls `AddQualifiedInferer `

x/emissions/keeper/keeper.go
guilherme-brandao: This is already checked in the unique function that calls `AddQualifiedForecaster `

x/emissions/keeper/keeper.go
```diff
@@ -1044,24 +1070,27 @@ func (k *Keeper) AppendInference(
...
-		// Update latests inferences for each worker
-		key := collections.Join(topicId, inference.Inferer)
-		if err := k.inferences.Set(ctx, key, *inference); err != nil {
-			return errorsmod.Wrap(err, "error setting inferences")
-		}
+	err := inference.Validate()
+	if err != nil {
+		return errorsmod.Wrap(err, "inference in list is invalid")
 	}
+	key := collections.Join(topicId, inference.Inferer)
```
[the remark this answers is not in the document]
guilherme-brandao: This has already been checked before

x/emissions/keeper/keeper.go
```diff
@@ -1044,24 +1070,27 @@ func (k *Keeper) AppendInference(
...
+	return k.inferences.Set(ctx, key, inference)
+}
+
+func (k *Keeper) InsertQualifiedInferences(
+	ctx context.Context,
+	topicId TopicId,
+	nonceBlockHeight BlockHeight,
+	inferences types.Inferences,
+) error {
```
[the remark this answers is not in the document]
guilherme-brandao: This has already been checked before

x/emissions/keeper/keeper.go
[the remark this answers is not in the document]
xmariachi: imho function should ensure this too as last line of defense. Because it could be called from swh else at some point.

x/emissions/keeper/keeper.go
```diff
@@ -1004,64 +1029,79 @@ func (k *Keeper) AppendInference(
...
+	topicId TopicId,
+	inference types.Inference,
+) error {
+	err := inference.Validate()
+	if err != nil {
+		return errorsmod.Wrap(err, "inference in list is invalid")
+	}
+	key := collections.Join(topicId, inference.Inferer)
+	return k.inferences.Set(ctx, key, inference)
+}
```
xmariachi, line 1095: I can't find where the `inferences` store is cleared. 
Seems like it should have all of its entries deleted before getting a new epoch in, right? 
(same for `forecasts`)

guilherme-brandao: added!

x/emissions/keeper/actor_utils/losses.go
```diff
@@ -22,17 +22,11 @@ import (
 func CloseReputerNonce(
 	k *keeper.Keeper,
 	ctx sdk.Context,
-	topicId keeper.TopicId,
+	topic types.Topic,
```
kpeluso, line 25: why is the whole topic being passed in? not just topicId like before? no other part of the topic is being used but the id...

guilherme-brandao: we use the `AlphaRegret`, so we were redundantly querying it before...

xmariachi (dismissed):
LGTM

x/emissions/proto/emissions/v4/query.proto
```diff
@@ -399,6 +399,21 @@ service QueryService {
     option (cosmos.query.v1.module_query_safe) = true;
     option (google.api.http).get = "/emissions/v4/next_churning_block_by_topic_id/{topic_id}";
   }
+
+  rpc GetActiveReputersForTopic(GetActiveReputersForTopicRequest) returns (GetActiveReputersForTopicResponse) {
```
relyt29, line 403: if we're indeed merging #615 this will want to be on `v5` not `v4`

relyt29 (changes requested):

x/emissions/keeper/keeper.go
```diff
@@ -203,6 +214,11 @@ type Keeper struct {
 
 	topicLastWorkerCommit  collections.Map[TopicId, types.TimestampedActorNonce]
 	topicLastReputerCommit collections.Map[TopicId, types.TimestampedActorNonce]
+
+	// active reputers for a topic
+	activeReputers collections.KeySet[collections.Pair[TopicId, ActorId]]
```
relyt29, line 219: all of these need export and import in `x/emissions/keeper/genesis.go` so we can dump database state in an emergency

guilherme-brandao: Updated!

x/emissions/keeper/actor_utils/losses.go
```diff
@@ -69,44 +63,22 @@ func CloseReputerNonce(
...
-
-		// Check that the reputer is registered in the topic
-		isReputerRegistered, err := k.IsReputerRegisteredInTopic(ctx, bundle.ValueBundle.TopicId, reputer)
-		if err != nil {
-			continue
-		}
-		// We'll keep what we can get from the payload, but we'll ignore the rest
-		if !isReputerRegistered {
-			continue
+			return types.ErrNoValidBundles
```
xmariachi, line 77: imho they are not that redundant.

ingress checks and closeReputerNonce happen at different times, and react (error) on different ways in each case(to user, or to validator logs); and for completeness, none of the function should allow that.

Also, it is possible that the reputer has deregistered during the time from the tx was made to the time when the reputer nonce is closed.

xmariachi (dismissed):
in general lgtm, but requested setting back a validation that I think is not redundant

x/emissions/proto/emissions/v3/genesis.proto
```diff
@@ -185,6 +185,21 @@ message GenesisState {
   repeated TopicIdAndDec previousTopicQuantileInfererScoreEma = 63;
   repeated TopicIdAndDec previousTopicQuantileForecasterScoreEma = 64;
   repeated TopicIdAndDec previousTopicQuantileReputerScoreEma = 65;
+
```
kpeluso, line 188: This for sure should be in `v3`?

x/emissions/proto/emissions/v4/query.proto
```diff
@@ -399,6 +399,21 @@ service QueryService {
...
+  rpc GetActiveForecastersForTopic(GetActiveForecastersForTopicRequest) returns (GetActiveForecastersForTopicResponse) {
+    option (cosmos.query.v1.module_query_safe) = true;
+    option (google.api.http).get = "/emissions/v4/active_forecasters/{topic_id}";
+  }
+
+  rpc GetActiveInferersForTopic(GetActiveInferersForTopicRequest) returns (GetActiveInferersForTopicResponse) {
+    option (cosmos.query.v1.module_query_safe) = true;
+    option (google.api.http).get = "/emissions/v4/active_inferers/{topic_id}";
+  }
 }
```
kpeluso, line 417: Missing corresponding queries for the other new maps:

* LowestInfererScoreEma
* LowestForecasterScoreEma
* LowestReputerScoreEma
* LossBundles

x/emissions/keeper/keeper.go
```diff
@@ -134,6 +146,9 @@ type Keeper struct {
 	// map of (topic, worker) -> forecast[]
 	forecasts collections.Map[collections.Pair[TopicId, ActorId], types.Forecast]
 
+	// map of (topic, reputer) -> reputer loss
+	lossBundles collections.Map[collections.Pair[TopicId, Reputer], types.ReputerValueBundle]
```
kpeluso, line 150: Would consider calling this `topicLastReputerPayload`

kpeluso, line 150: There may be a way to consolidate `lossBundles`, `topicLastReputerCommit`, and `allLossBundles`. May avoid loading unnecessary data at places by keeping them separate though, just a thought.

x/emissions/keeper/keeper.go
```diff
@@ -1135,59 +1177,82 @@ func (k *Keeper) AppendForecast(
...
-	if previousEmaScore.Score.Gt(lowScore.Score) {
+
+	if previousEmaScore.Score.Gt(lowestEmaScore.Score) {
 		// Update EMA score for the lowest score forecaster, who is not the current forecaster
 		err = k.CalcAndSaveForecasterScoreEmaWithLastSavedTopicQuantile(
 			ctx,
 			topic,
 			nonceBlockHeight,
-			forecasts.Forecasts[lowScoreIndex].Forecaster,
+			lowestEmaScore,
```
kpeluso, line 1200: good opt

x/emissions/keeper/queryserver/query_server_losses.go
```diff
@@ -61,3 +61,25 @@ func (qs queryServer) GetReputerLossBundlesAtBlock(ctx context.Context, req *typ
...
 }
+
+func (qs queryServer) GetActiveReputersForTopic(
+	ctx context.Context,
+	req *types.GetActiveReputersForTopicRequest,
+) (
+	*types.GetActiveReputersForTopicResponse,
+	error,
+) {
+	topicExists, err := qs.k.TopicExists(ctx, req.TopicId)
```
kpeluso, line 72: Must record metrics here

x/emissions/keeper/queryserver/query_server_inferences.go
```diff
@@ -55,6 +55,22 @@ func (qs queryServer) GetInferencesAtBlock(ctx context.Context, req *emissionsty
 	return &emissionstypes.GetInferencesAtBlockResponse{Inferences: inferences}, nil
 }
 
+func (qs queryServer) GetActiveInferersForTopic(ctx context.Context, req *emissionstypes.GetActiveInferersForTopicRequest) (*emissionstypes.GetActiveInferersForTopicResponse, error) {
+	topicExists, err := qs.k.TopicExists(ctx, req.TopicId)
```
kpeluso, line 59: Must record metrics here

x/emissions/keeper/queryserver/query_server_forecasts.go
```diff
@@ -28,3 +28,19 @@ func (qs queryServer) GetForecastsAtBlock(ctx context.Context, req *types.GetFor
 
 	return &types.GetForecastsAtBlockResponse{Forecasts: forecasts}, nil
 }
+
+func (qs queryServer) GetActiveForecastersForTopic(ctx context.Context, req *types.GetActiveForecastersForTopicRequest) (*types.GetActiveForecastersForTopicResponse, error) {
+	topicExists, err := qs.k.TopicExists(ctx, req.TopicId)
```
kpeluso, line 33: Must record metrics here

x/emissions/keeper/keeper.go
```diff
@@ -3482,3 +3582,263 @@ func (k *Keeper) GetPreviousForecasterScoreRatio(ctx context.Context, topicId To
...
+
+// SetLowestReputerScoreEma sets the lowest reputer score EMA for a topic
+func (k *Keeper) SetLowestReputerScoreEma(ctx context.Context, topicId TopicId, lowestScore types.Score) error {
+	if err := types.ValidateTopicId(topicId); err != nil {
+		return errorsmod.Wrap(err, "invalid topic id")
+	}
+	if err := types.ValidateBech32(lowestScore.Address); err != nil {
+		return errorsmod.Wrap(err, "invalid address")
+	}
+	return k.lowestReputerScoreEma.Set(ctx, topicId, lowestScore)
```
kpeluso, line 3821: Should fire off an event imo, and for lowest inferer and forecaster score EMA setting as well

x/emissions/keeper/keeper.go
```diff
@@ -3482,3 +3582,263 @@ func (k *Keeper) GetPreviousForecasterScoreRatio(ctx context.Context, topicId To
...
+func (k *Keeper) RemoveActiveReputer(ctx context.Context, topicId TopicId, reputer ActorId) error {
+	key := collections.Join(topicId, reputer)
+	return k.activeReputers.Remove(ctx, key)
+}
+
+// GetActiveReputersForTopic returns all active reputers for a specific topic
+func (k *Keeper) GetActiveReputersForTopic(ctx context.Context, topicId TopicId) ([]ActorId, error) {
+	var reputers []ActorId
+	rng := collections.NewPrefixedPairRange[TopicId, ActorId](topicId)
+	err := k.activeReputers.Walk(ctx, rng, func(key collections.Pair[TopicId, ActorId]) (bool, error) {
```
kpeluso, line 3756: Though not necessary due to the indirect effect of other limits, limiting the number of loops would make this supremely safe. Here and the other spots you're `.Walk(...)`-ing

x/emissions/keeper/actor_utils/worker.go
```diff
@@ -97,72 +87,72 @@ func CloseWorkerNonce(k *keeper.Keeper, ctx sdk.Context, topicId keeper.TopicId,
...
 
-	return acceptedInferers, nil
+	return activeInfererAddressesMap, nil
 }
 
 // insert forecasts from top forecasters
 // check forecast elements to ensure they are forecasts made about
 // the active list of inferers.
-func insertForecastsFromTopForecasters(
+func closeActiveForecastsSet(
```
kpeluso, line 142: better

x/emissions/keeper/actor_utils/losses.go
kpeluso: Not all functions need to check everything -- a function's "contract" it signs with the world includes its intention specified in its comments (by its declaration). That can include making assumptions about the arguments and viewed state. This is esp true and necessary in a compute-constrained environment.

We're also building `x/emissions` as nearly a monolith, so we get to make more "assumptions" like this (as opposed to a modular/microservice architecture).

The real pull to add these validations is if we expect / should prepare to ever edit the values in these KV stores at any point between ingress and their usage. This is minimized by adding these validations to within the scope of any setter. But then again such checks too can be averted by state-editing migrations or not using the designated setters.

So I'm a bit torn but reintroducing them seems unavoidable if we want to max safety.

Curious @spooktheducks thoughts...

x/emissions/keeper/score_utils.go
```diff
@@ -6,77 +6,161 @@ import (
...
 	if err != nil {
-		return types.Score{}, lowScoreIndex, err
+		return err
 	}
-	for index, extForecast := range forecasts.Forecasts {
-		extScore, err := k.GetForecasterScoreEma(ctx, topicId, extForecast.Forecaster)
+	for _, address := range forecasterAddresses {
+		score, err := k.GetForecasterScoreEma(ctx, topicId, address)
+		if err != nil {
+			return err
```
kpeluso, line 140: It's interesting how you `return err` here but `continue` in L159 i.e. `GetLowestScoreFromAllForecasters()`. Same for the pairs of analogous functions of other actors.

x/emissions/keeper/score_utils.go
```diff
@@ -6,77 +6,161 @@ import (
...
+	// remove forecaster from the list
+	for i, address := range forecasterAddresses {
+		if address == removedForecasterAddress {
+			forecasterAddresses = append(forecasterAddresses[:i], forecasterAddresses[i+1:]...)
+			break
+		}
+	}
+	// add new forecaster to the list
+	forecasterAddresses = append(forecasterAddresses, addedForecaster)
+	lowScore, err := k.GetForecasterScoreEma(ctx, topicId, forecasterAddresses[0])
```
kpeluso, line 133: This call and the loop below it repeat the functionality of `GetLowestScoreFromAllForecasters()`, so I think you should call it here. Same for the pairs of analogous functions of other actors.

kpeluso (changes requested):
A subset of my previous comments are merge blocking.

x/emissions/proto/emissions/v4/query.proto
guilherme-brandao: We already have queries for the first 3, please double-check. 

The latter, I don't think it is necessary because it's an ephemeral storage; it will be deleted when the reputer nonce is closed, but it is open to discussion. If we decide to not have it, I need to delete `GetWorkerLatestInferenceByTopicId`.

x/emissions/keeper/keeper.go
guilherme-brandao: I followed the pattern that we were using to name the KVStore that we are using to save individual submissions: `inferences`, `forecasts` then `lossBundles`.

guilherme-brandao: They have different purposes, but we can think about improving this in the future ofc.

x/emissions/keeper/queryserver/query_server_losses.go
guilherme-brandao: good catch

x/emissions/keeper/keeper.go
guilherme-brandao: as you said, the loop is already capped by `MaxTopReputersToReward` but it is probably not a biggie to add since we already have the module params in the Txs that are calling it. 

On the `.Clear()`, I don't think we need it because the intention is to really Clear everything. 

Lmk what's your final say on this.

x/emissions/keeper/keeper.go
guilherme-brandao: Should we? We already do on the reward distribution. We can make a different events for this, but I'm not sure if it will gonna have any use, at least for now. Wondering what @xmariachi thinks about this...

x/emissions/keeper/score_utils.go
guilherme-brandao: Followed the same pattern in this one

x/emissions/keeper/score_utils.go
guilherme-brandao: good catch!

x/emissions/keeper/keeper.go
kpeluso: My point is that we should check if `.Clear()` is `O(1)` or `O(n)` -- if latter it should be bounded. But it's indirectly bounded in this case as discussed => resolved

relyt29:
I would like to give this a review because I feel like I haven't had a thorough chance to look at it, but I want to wait until after it's rebased on-top of dev / takes the experience PR into account

x/emissions/keeper/keeper.go
xmariachi: Tough call.
I guess the intended usage is to have, while an epoch runs, a reading of the current lowest score, for the actors to have a more informed decision to send or not.

Once the window is closed that can be taken from the lowest score of the active inferers, however these are published when ground_truth_lag happens. (This includes the case for analytics - they can be taken from the indexer).

But this may be a very big number of events per topic/epoch (I'd say typically hundreds but can be thousands on testnet at least, although we should not optimize for testnet), and I'm not sold on the usage of this, so I'd lean towards not adding it for now, all things considered. We can emit this on-demand.

x/emissions/keeper/actor_utils/losses.go
relyt29: `d e f e n s e   i n   d e p t h`

x/emissions/proto/emissions/v5/query.proto
```diff
@@ -409,6 +409,21 @@ service QueryService {
     option (cosmos.query.v1.module_query_safe) = true;
     option (google.api.http).get = "/emissions/v5/count_forecaster_inclusions_in_topic/{topic_id}/{forecaster}";
   }
+
+  rpc GetActiveReputersForTopic(GetActiveReputersForTopicRequest) returns (GetActiveReputersForTopicResponse) {
+    option (cosmos.query.v1.module_query_safe) = true;
+    option (google.api.http).get = "/emissions/v4/active_reputers/{topic_id}";
```
relyt29, line 415: should these say v5 in the url link? should we perhaps set "vX" to a variable in the proto that's easily updateable so errors like this don't happen in the future?

relyt29 (changes requested):
directionally good, one minor thing here

x/emissions/proto/emissions/v5/query.proto
guilherme-brandao: good catch!

relyt29 (approved):
it's a big PR but to my limited eyesight it looks good

x/emissions/keeper/keeper.go
```diff
@@ -3537,3 +3682,263 @@ func (k *Keeper) GetPreviousForecasterScoreRatio(ctx context.Context, topicId To
...
+func (k *Keeper) IsActiveForecaster(ctx context.Context, topicId TopicId, forecaster ActorId) (bool, error) {
+	key := collections.Join(topicId, forecaster)
+	return k.activeForecasters.Has(ctx, key)
+}
+
+// RemoveActiveForecaster removes a forecaster from the active forecasters set for a topic
+func (k *Keeper) RemoveActiveForecaster(ctx context.Context, topicId TopicId, forecaster ActorId) error {
+	key := collections.Join(topicId, forecaster)
+	return k.activeForecasters.Remove(ctx, key)
+}
```
kpeluso, line 3746: Should this not be indexed by nonce as well? because active set changes across nonces... Not every worker is active for every epoch. An active participant may not even participate in all epochs. Also, notably, an epoch can be judged way later (ground truth lag blocks later) so if there is a dynamic active set (which there is) would the wrong actors be scored when an earlier nonce is ready to be scored?

kpeluso (changes requested):

x/emissions/keeper/keeper.go
kpeluso: oh I"m sorry this is wrong -- you don't query this map when scoring workers. These maps are reset between epochs per actor type (Worker | Reputer)

kpeluso (approved):

x/emissions/keeper/actor_utils/losses.go
kpeluso: There is no deregistration. The checks are made earlier. I'd like to talk more offline ab it -- can quickly add back in if we feel we want to. Will merge the PR now though as it looks good and it does so much more than touch this validation.

PR #625, opened 2024-10-12
WIP-Impl idea for wp-like rewards

xmariachi:

* New keeper field `totalSumPreviousTopicWeights` which keeps track of the sum of active previous topics weights (not per-block), to introduce as total weights in rewards calculation.
   * Notably includes keeping track of this on inactivation/reactivation.
* Even if rewards are still calculated for topics active in a block, it uses all "currently active" topics. 
* In case of error on reactivation, it explicitly inactivates, to substract from the `totalSumPreviousTopicWeights`, and trying to give certainty to that topic's status. 
* Maintains keeping invalidated topics' previousTopicWeights - only removes from the `totalSumPreviousTopicWeights` 
* Applies time-based correction on the rewards (ie adjusting per epochLength's values). Uses `BlocksPerMonth` to normalize as per research's suggestions around eq 53. This will lead to an initial apparent big reduction on the amounts given.

**ALTERNATIVE IMPLEMENTATIONS**

* **Keeping total weights:**  Calculate overall-active topics (not per block), calculate its new weigths, use them for doing the calculations on the rewards. 
  * this is simpler but it involves iteration over all of the topics. Current impl involves keeping track of a field in the keeper on activation/inactivation of topics. However this could be a safer, less complex alt. Looking for opinions here @relyt29 @kpeluso @guilherme-brandao 

* Rewards calculation and emission: Note also that, because of these changes, rewards are accumulated (each topic gets its weight-based, epoch-based share, but the rest remains there, accumulating for next blocks). An alternative, recommended by research, is to not generate rewards every block(as it's done in `mint`'s `BeginBlock`), but just keep the rewards until an epochEnd arrives and then, after calculation, that exact amount is sent from treasury then.

## Still Left Todo

* Needs a migration where the newly introduced `keeper.totalSumPreviousTopicWeights` is filled up with data from the active topics' previousTopicWeights.


kpeluso:
@xmariachi Absolutely cannot afford to iterate over every topic in the long run

relyt29:
can we use the [sushiswap rewardDebt technique](https://dev.to/heymarkkop/understanding-sushiswaps-masterchef-staking-rewards-1m6f) here, as we do with delegated validator rewards?

xmariachi:
> can we use the [sushiswap rewardDebt technique](https://dev.to/heymarkkop/understanding-sushiswaps-masterchef-staking-rewards-1m6f) here, as we do with delegated validator rewards?

Thanks a lot @relyt29 
I read through that link.
I understand here that `epochEnd` would play the role of the `harvest`.  
But I think it can be simplified, not really needing `rewardDebt`. 
There are several differences I see (eg stake not playing a part, but topic weights instead, see "NOTE 1") but the main one I see:
* `rewardsPerBlock` is variable in our case, not constant.

This would require some "per-activity-zones" calculation of rewards, which is pretty tricky. And because epochs are so variable and overlapping, keeping track of such activity zones would again incur in inefficient processing.

* Suggestion: we could **approximate rewards per block**: 
Rewards per block may not be exact, but we can keep an approximation (a slow EMA), then define: 
`rewardToGiveTopic = (topicWeight/totalWeights)* approximateRewardsPerBlock * EpochLength`
In this case we would not be needing to use a time horizon (eg `BlocksPerMonth`).
And no need to keep a different structure like `PoolStaker` I think (because we know `topic.EpochLength` is fixed)
And in any case, the rewards to be given are dependent on module params, so they wont change as often, and this would smooth its effect.


Problem is that rewards given may not be exact on param value changes, however it makes calculations are very light.
Another simplification here is that weights are taken at the block of topic reward, but this is an OK approximation (as per research and my opinion too)

----
NOTE: In practice, the current implementation's effect is pretty similar to the one provided; the main difference I'd say is that the rewards used are the accumulated ones, up to that point in time. This makes the time-dimension placement still a factor, even if tiny, _very_ dilluted because the rewards substracted by the previous topics is still a tiny fraction of the total - but this is not ideal. I prefer the solution above.

x/emissions/keeper/topic_activation.go
```diff
@@ -157,6 +157,32 @@ func (k *Keeper) inactivateTopicWithoutMinWeightReset(ctx context.Context, topic
 		return errors.Wrap(err, "failed to remove active topics")
 	}
 
+	// Remove previous weight from total sum of previous topic weights, but keep on list
```
relyt29, line 160: so this is good, but shouldn't there be an equivalent function happening when a topic is first activated? is that happening in the keeper and i'm not seeing it?

xmariachi: Good catch. 
This is taken care of in the setPreviousTopicWeight for new topics that get activated, but old topics with a weight should also have this protection or we would be double-adding to total weights.
I'm reviewing this now. 
One of the candidates for explicit unit testing: activate/inactivate/activate, and check total weights.

xmariachi: Fixed, tested with current tests, but pending adding the test defined above.
No test was breaking before or after this.

guilherme-brandao (changes requested):

x/emissions/keeper/topic_activation.go
```diff
@@ -262,6 +288,27 @@ func (k *Keeper) ActivateTopic(ctx context.Context, topicId TopicId) error {
 		return errors.Wrap(err, "failed to set active topics")
 	}
 
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
```
guilherme-brandao, line 291: Context is already unwrapped above, reuse it

x/emissions/module/rewards/rewards.go
```diff
@@ -45,26 +46,46 @@ func EmitRewards(
...
+	// Get the global total sum of previous topic weights
+	totalWeightOfActiveTopics, err := k.GetTotalSumPreviousTopicWeights(ctx)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get total sum of previous topic weights")
+	}
+	// Get epoch lengths for sorted rewardable topics
+	epochLengths := make(map[uint64]int64)
 	for _, topicId := range sortedRewardableTopics {
-		sumWeightOfRewardableTopics, err = sumWeightOfRewardableTopics.Add(*weights[topicId])
+		topic, err := k.GetTopic(ctx, topicId)
```
guilherme-brandao, line 58: During the `EndBlocker` we do `GetTopic` multiple times: 
- On `GetAndUpdateActiveTopicWeights`
- On `UpdateNoncesOfActiveTopics`
- On `EmitRewards` 
- On `getWorkersRewardFractions`
- On `pruneRecordsAfterRewards`

Probably we can optimize it... 

I know 80% of this wasn't introduced in your PR, so lmk if you think it's better to create a separate ticket to optimize this

xmariachi: yeah probably wiser to separate that , to separate responsibiitlies. This PR has enough to worry about lol

x/emissions/module/rewards/rewards.go
```diff
@@ -159,26 +180,52 @@ func getDistributionAndPayoutRewardsToTopicActors(
...
+			return nil, errors.Wrapf(err, "calcTopicRewards:topic reward multiplication with epoch length fraction error")
+		}
+		topicRewards[topicId] = &topicRewardPerEpoch
+		totalTopicRewardsSum, err = totalTopicRewardsSum.Add(topicRewardPerEpoch)
+		if err != nil {
+			return nil, errors.Wrapf(err, "calcTopicRewards: total topic rewards sum error")
+		}
+	}
+	if totalTopicRewardsSum.Gt(totalReward) {
+		return nil, errors.Wrapf(types.ErrInvalidReward, "total topic rewards sum %s is greater than total reward in treasury %s, cancelling rewards distribution", totalTopicRewardsSum.String(), totalReward.String())
```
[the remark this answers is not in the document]
xmariachi: if we return the error with a meaningful message, I think we don't want to log it. The caller that chooses to process the error (and not propagate it) is responsible to log it.
Otherwise we end up with errors logged at every step, while we only want to have the one on top, which also writes the whole stack trace so has the whole info.

x/emissions/module/rewards/rewards.go
```diff
@@ -45,26 +46,46 @@ func EmitRewards(
 	if uint64(len(sortedRewardableTopics)) > moduleParams.MaxActiveTopicsPerBlock {
 		sortedRewardableTopics = sortedRewardableTopics[:moduleParams.MaxActiveTopicsPerBlock]
 	}
-	// Get total weight of rewardable topics
-	sumWeightOfRewardableTopics := alloraMath.ZeroDec()
+
+	// Get the global total sum of previous topic weights
+	totalWeightOfActiveTopics, err := k.GetTotalSumPreviousTopicWeights(ctx)
```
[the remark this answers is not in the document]
xmariachi: done, ty

x/emissions/keeper/topic_activation.go
xmariachi: added test for such case

x/emissions/module/rewards/rewards.go
```diff
@@ -159,26 +180,52 @@ func getDistributionAndPayoutRewardsToTopicActors(
...
+	}
+	totalTopicRewardsSum := alloraMath.ZeroDec()
 	topicRewards := make(map[TopicId]*alloraMath.Dec)
 	for _, topicId := range sortedTopics {
-		topicRewardFraction, err := GetTopicRewardFraction(weights[topicId], sumWeight)
+		topicWeight := weights[topicId]
+		if topicWeight == nil {
+			zero := alloraMath.ZeroDec()
+			topicWeight = &zero
+		}
```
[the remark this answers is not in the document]
xmariachi: done, ty

xmariachi:
Comments so far have been addressed.

x/emissions/module/rewards/rewards.go
```diff
@@ -159,26 +176,49 @@ func getDistributionAndPayoutRewardsToTopicActors(
...
 		if err != nil {
 			return nil, errors.Wrapf(err, "topic reward error")
 		}
-		topicRewards[topicId] = &topicReward
+		epochLength := epochLengths[topicId]
+		if epochLength <= 0 {
+			return nil, errors.Wrapf(types.ErrInvalidLengthTopic, "epoch length is nil or zero for topic %d", topicId)
+		}
+		epochLengthDec := alloraMath.NewDecFromInt64(epochLength)
+		topicRewardPerEpoch, err := topicRewardPerBlock.Mul(epochLengthDec)
```
kpeluso, line 210: I'm not sure this is right / what @jmdkastro meant. I think this is nearly always going to trip the `totalTopicRewardsSum.Gt(totalReward)` condition below (henceforth: "the condition below").

Per [his comment](https://linear.app/alloralabs/issue/RES-424/asynchronous-topic-weightpayout-fix#comment-fc74dda0), I believe we have to normalize by the unit time "over which we want this [pot] to be depleted". This is the `[time]` term he mentions when he says `dt/[time]`.

Even with this division, I'm not convinced that we're going to reliably *not* trip the condition below. I'm wondering if we have to rebalance/renormalize rewards per topic upon satisfaction of that condition, and/or if we should apply some heuristic that guarantees (perhaps with some v high probability) the total rewards will be less than whatever is left in the rewards pot.

Maybe we need to define a "global normalization divisor" that gets updated upon every topic inactivation and activation.

kpeluso (changes requested):

x/emissions/module/rewards/rewards.go
kpeluso: I did all the algebra and actually I see now: you're right, I'm wrong! The terms cancel out and yeah you only need to multiply by `[time]`. Very good.

kpeluso (approved):
I had previously requested changes because some math seemed off but I did the algebra and it makes sense now. You can see my resolved comment above for more details. Every other comment I had in my previous review is minor and shouldn't block progress. Good work!

xmariachi:
> I had previously requested changes because some math seemed off but I did the algebra and it makes sense now. You can see my resolved comment above for more details. Every other comment I had in my previous review is minor and shouldn't block progress. Good work!

Thanks a lot, also for double-checking the algebra :) 

A comment on one decision taken (which only affects testnet - minor thing but affecting "life after migration", though for one epoch only at worst case):
In a currently running network, with potentially zero rewards in pool and with several active topics (ie current testnet), there is an unlikely but possible chance that there are not enough rewards yet in the pool by the time the reward payment is due. Imagine that eg next block after migration, an endBlock ends for a topic with a big weight.

The question here is, if there are not enough rewards in pool, what to do: 
1) fail and accumulate, looking for a stable rewards pool which will most likely be OK in a few blocks(missing at worst one reward round for _some_ topics); 
2) "give what you have", splitting the current pool among the active topics in the block.

I think it's more appropriate to fail and accumulate, and ensure all rewards are given with same formula. 

Due to the current algebra, this won't affect mainnet, since rewards are accrued from the entire lifetime of the topics.

--- 

Another note: just highlighting, that one of the gotchas of current impl, is that the `totalSumPreviousTopicWeight` gets updated on these situations: 
* when a topic is active: on its epochEnd - gets "updated" (remove previous weight, add new one)
* transition active -> inactive: remove previous weight 
* transition inactive -> active: add previous weight 

This goes to make clear that this "link" between the two process (in/activation of topic and correctness of rewards calculation) are linked through this.

x/emissions/module/rewards/rewards.go
```diff
@@ -159,26 +176,49 @@ func getDistributionAndPayoutRewardsToTopicActors(
...
 	ctx sdk.Context,
-	k keeper.Keeper,
 	weights map[uint64]*alloraMath.Dec,
 	sortedTopics []uint64,
-	sumWeight alloraMath.Dec,
-	totalReward alloraMath.Dec,
+	sumWeight alloraMath.Dec, // sum of all topic weights
+	totalReward alloraMath.Dec, // maxTotalReward in treasury
+	epochLengths map[uint64]int64, // epoch lengths for each topic
+	currentBlockEmissionDec alloraMath.Dec, // perBlockEmission
```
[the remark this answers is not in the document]
xmariachi: Lol I had added explicit comments "maxTotalReward in treasury" and "perBlockEmission" bc I had had the same feeling when self-reviewing!

Will improve this and actually add a comment to the function.

kpeluso:
@xmariachi Yes I noticed the "fail and accumulate" characteristic. I was weighing this against rebalancing/renormalizing the rewards to what is possible to allot, but thought this could be too compute-heavy. wyt?

I like the "due" vs "active" distinction, that's a good call-out.

@kush-alloralabs take note on all the above so we could update docs and unify language.

xmariachi:
> @xmariachi Yes I noticed the "fail and accumulate" characteristic. I was weighing this against rebalancing/renormalizing the rewards to what is possible to allot, but thought this could be too compute-heavy. wyt?

This will only affect those cases when bootstrapping rewards(testnet), or if there is a dramatic increase on the rewards emitted per block (currently changing monthly, or when a relevant parameter is changed). So compute-heavy is not that relevant.

As an alternative in this case, when the treasury doesn't cover the full amount, I suggest distributing the available treasury rewards by proportionally weighing the participants' entitlements due for that block, using a formula based on `topicWeight * epochLength`. This approach follows a "distribute something, even if it's not the full expected amount" policy. The algebra should naturally allow for a full recovery eventually, and in most cases, quite quickly, so we are good there.

Maybe this is nicer to actors that otherwise may complain ab a missing reward?

CHANGELOG.md
```diff
@@ -66,6 +66,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 * [#538](https://github.com/allora-network/allora-chain/pull/538) Refactor Inference Synthesis to use Functions instead of "Builder Pattern"
+* [#625](https://github.com/allora-network/allora-chain/pull/625) Rewards system revamped, closer to the whitepaper spec.
```
relyt29, line 69: This description does not actually describe what changed

```suggestion
* [#625](https://github.com/allora-network/allora-chain/pull/625) Determine what percentage of rewards to pay to each topic based on time-based accumulation rather than instantaneously.
```

x/emissions/proto/emissions/v5/query.proto
```diff
@@ -250,6 +250,11 @@ service QueryService {
     option (google.api.http).get = "/emissions/v5/previous_topic_weight/{topic_id}";
   }
 
+  rpc GetTotalSumPreviousTopicWeights(GetTotalSumPreviousTopicWeightsRequest) returns (GetTotalSumPreviousTopicWeightsResponse) {
+    option (cosmos.query.v1.module_query_safe) = true;
+    option (google.api.http).get = "/emissions/v5/previous_total_topic_weight";
```
relyt29, line 255:
```suggestion
    option (google.api.http).get = "/emissions/v5/sum_previous_total_topic_weight";
```

x/emissions/migrations/v5/migrate_test.go
```diff
@@ -257,6 +257,155 @@ func (s *EmissionsV5MigrationTestSuite) TestMigratedTopicWithNaNInitialRegret()
...
+	// Verify the sumPreviousTopicWeights store has been updated correctly
+	sumPreviousTopicWeightsStore := prefix.NewStore(store, emissionstypes.TotalSumPreviousTopicWeightsKey)
+	weightsIterator := sumPreviousTopicWeightsStore.Iterator(nil, nil)
+	s.Require().True(weightsIterator.Valid())
+	defer weightsIterator.Close()
+
+	var weightsMsg alloraMath.Dec // collcodec.ValueCodec[alloraMath.Dec]
+	err = weightsMsg.Unmarshal(weightsIterator.Value())
+	s.Require().NoError(err)
+	s.Require().True(topic1WeightDec.Equal(weightsMsg))
```
relyt29, line 347: No issue with this test in terms of correctness, it's good

But philosophically, do we need to do all this checking of every other value in this function? In general can we write many small tests that just test one thing, rather than one big test?

x/mint/testutil/emissions_keeper_mocks.go
```diff
@@ -110,3 +110,15 @@ func (m *MockEmissionsKeeper) IsWhitelistAdmin(ctx context.Context, admin string
 	ret1, _ := ret[1].(error)
 	return ret0, ret1
 }
+
+func (mr *MockEmissionsKeeperMockRecorder) SetRewardCurrentBlockEmission(ctx interface{}, emission interface{}) *gomock.Call {
```
relyt29, line 114: @fernandofcampos this is one of the places where we traditionally have had to write mocks ourselves even though the files are supposed to be autogenerated

relyt29 (dismissed):
minor nitpicks, overall as far as I can tell, should be correct

x/emissions/migrations/v5/migrate_test.go
xmariachi: They are complementary imo.
Re. the verbose checks on the topics in this function too, which are similar to the ones in the other test ab topics: I left them in knowingly because there is one consideration: ensuring that the topic also change in case of active topics (the other test for the topic correctness doesn't have active topics to work with).

x/emissions/proto/emissions/v5/query.proto
xmariachi: applied and generated protos

x/mint/keeper/keeper.go
```diff
@@ -159,12 +159,17 @@ func (k Keeper) PayAlloraRewardsFromEcosystem(ctx context.Context, rewards sdk.C
...
 		ctx,
 		types.EcosystemModuleName,
 		emissionstypes.AlloraRewardsAccountName,
 		rewards,
 	)
+	if err != nil {
+		return err
+	}
+
+	return k.emissionsKeeper.SetRewardCurrentBlockEmission(ctx, rewards.AmountOf(params.BaseCoinUnit))
```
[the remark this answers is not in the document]
xmariachi: on it

xmariachi:
> ...
>  This approach follows a "distribute something, even if it's not the full expected amount" policy. The algebra should naturally allow for a full recovery eventually, and in most cases, quite quickly, so we are good there.
> 
> Maybe this is nicer to actors that otherwise may complain ab a missing reward?

added this impl, also refactored code for clarity.

x/emissions/module/rewards/rewards.go
```diff
@@ -157,29 +219,115 @@ func getDistributionAndPayoutRewardsToTopicActors(
...
+	topicRewards, totalTopicRewardsSum, err := calculateRewardsForCurrentTopics(args)
+	if err != nil {
+		return nil, errors.Wrapf(err, "calcTopicRewards: calculate rewards for current topics error")
+	}
+	if totalTopicRewardsSum.Gt(args.TotalAvailableInRewardsTreasury) {
+		// Note this will happen when the treasury is not enough to cover the rewards for all topics
+		// This will happen very rarely (cases where rewards emission changes dramatically), so optimization is done for the main case.
+		Logger(args.Ctx).Warn("Treasury lower than calculated rewards. Distributing treasury equally among current topics.")
+		if len(args.SortedTopics) == 1 {
+			// most likely case, only one topic - so it gets the entire treasury directly
```
kpeluso, line 244: > most likely case

we should hold this is only true in the short term!

kpeluso (approved):

kpeluso (approved):

guilherme-brandao (approved):

x/mint/keeper/keeper.go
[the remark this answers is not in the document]
kpeluso: `EmitNewRewardCurrentBlockEmissionEvent` added to `x/mint/module/abci.go`

PR #626, opened 2024-10-13
rm excessive tx length check

kpeluso:

I think `checkInputLength()`, a check of the total msg size on each data ingress, is redundant and partially causing slowdowns at scale, shown here edited in [Guilherme’s recent PR](https://github.com/allora-network/allora-chain/pull/623)

Our profiling shows a lot of marshalling taking up precious memory, and this check is on every singple data ingress => easily could be one of the culprits. This check marshalls the entire proto msg afterall.

Meanwhile, there are configs to handle this already in `config.toml` and `app.toml` e.g. `max_tx_bytes` and gas-related params and market action. We obviosuly didn’t know this when we first added these checks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@ Summary TBD
 ### Refactor
 
 * [#538](https://github.com/allora-network/allora-chain/pull/538) Refactor Inference Synthesis to use Functions instead of "Builder Pattern"
+* [#623](https://github.com/allora-network/allora-chain/pull/623) Optimize worker and reputer payload ingress
 
 ### Monitoring
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,10 @@ Summary TBD
 * [#620](https://github.com/allora-network/allora-chain/pull/620) Add a static analyzer to detect non-deferred `.Close()` calls, improve migration error handling
 * [#622](https://github.com/allora-network/allora-chain/pull/622) Add telemetry metrics on queries/txs
 
+### Removed
+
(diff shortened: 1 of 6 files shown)


relyt29:
It looks like `TestMsgServerTestSuite -> TestInsertingHugeBundleWorkerPayloadFails` is not passing. It would be nice if we could somehow capture the `max_tx_bytes` enforcement for that test in the unit test environment, but I am not sure how to do so / if that is possible.

kpeluso:
We could look at the length of strings in the protos and enforce a stateful validation that uses a new `max_string_length` param @relyt29

xmariachi (changes requested):
We can keep the sanity size check, without having to marshall the proto, using `proto.Size`

[Size-limiting program output](https://protobuf.dev/reference/go/size/#size-limiting-program-output)

[code](https://go.googlesource.com/protobuf/+/refs/heads/master/proto/size.go)

From ChatGPT:
> proto.Size() is generally faster than marshalling because it avoids the overhead of constructing the entire byte slice representing the serialized message. Instead, it simply traverses the message fields and accumulates their sizes without performing actual encoding operations.

It can be tested experimentally by getting the times of calculating sizes using each operation over a large quantity of protos.

relyt29:
I vote for keeping proto.Size then, unless we can get a unit test that confirms `max_tx_bytes` is enforced

even if it isn't I think `max_tx_bytes` is up to each validator separately, don't want beefier validators causing issues for less beefy validators

spooktheducks (commented):
Just making a note re: our convo -- instead of losing the check entirely, we can call `theProtobuf.Size()` which only does extremely simple arithmetic

kpeluso:
Cancelling in favor of https://github.com/allora-network/allora-chain/pull/630 , which is aligned with all the suggestions above!

PR #627, opened 2024-10-14
Add fee market module

guilherme-brandao:
This PR introduces the fee market module from the [Skip Fee Market repository](https://github.com/skip-mev/feemarket) into our blockchain. The integration of this module aims to enhance transaction prioritization and fee handling by implementing a more efficient and fair mechanism for processing transactions. This addition will allow validators to better manage network congestion and optimize resource allocation.

https://linear.app/alloralabs/issue/PROTO-2564/add-xfeemarket

Yes


kpeluso (dismissed):

app/upgrades.go
```diff
@@ -23,5 +26,21 @@ func (app *AlloraApp) setupUpgradeHandlers() {
...
 			handler.CreateUpgradeHandler(app.ModuleManager, app.Configurator()))
+
+		upgradeInfo, err := app.UpgradeKeeper.ReadUpgradeInfoFromDisk()
+		if err != nil {
+			panic(err)
+		}
+
+		if upgradeInfo.Name == v0_7_0.Upgrade.UpgradeName && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) {
+			storeUpgrades := storetypes.StoreUpgrades{
+				Added:   []string{feemarkettypes.ModuleName},
```
amimart: By adding the `feegrant` module we need to mention it as well, I think here we need to specify the `StoreKey` but it's only a presentation matter as it is the same value (both ways seems used in the ecosystem..), see: https://github.com/skip-mev/feemarket/blob/f1f216e15aa6bcc05075a7079496764820a0f545/x/feemarket/types/keys.go#L7

```suggestion
				Added:   []string{feegrant.StoreKey, feemarkettypes.StoreKey},
```

app/feemarket.go
```diff
@@ -0,0 +1,35 @@
...
+	storetypes "cosmossdk.io/store/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+	feemarket "github.com/skip-mev/feemarket/x/feemarket"
+	feemarketkeeper "github.com/skip-mev/feemarket/x/feemarket/keeper"
+	feemarkettypes "github.com/skip-mev/feemarket/x/feemarket/types"
+)
+
+// registerFeeMarketModule registers the feemarket module.
+func (app *AlloraApp) registerFeeMarketModule() {
```
amimart: It's not mentioned in the `feemarket` documentation but it supports auto-wiring using `depinject`: https://github.com/skip-mev/feemarket/blob/f1f216e15aa6bcc05075a7079496764820a0f545/x/feemarket/module.go#L178

I'd rather register it in the `app.yaml` as I consider it the canonical way to register a module, and only fallback to manual wiring when not possible otherwise (e.g. IBC modules) if that make sense to you

CHANGELOG.md
```diff
@@ -90,6 +90,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* [#627](https://github.com/allora-network/allora-chain/pull/627) Add fee market module
```
amimart: Would also be worth mentioning the addition of the `feegrant` module

amimart (changes requested):
Nice one!

I've noticed the addition of the `feegrant` module besides the `feemarket`, it is very often present in cosmos based chain and allows to provide fee payer accounts. I'm in total favour of adding it but it's worth to mention that the `feemarket` doesn't require it and can take `nil` as `FeegrantKeeper`.

I've let some remarks, but the main concern I have is regarding the upgrade, I think we should store the initial params & state we want for the `feemarket` module otherwise we'll have zero values. Can help on that if you want, that may also be addressed in a different PR.

app/upgrades.go
guilherme-brandao: Good catch! I think the fee grant can be a nice addition; it could be an extra feature for topic creators to drive some demand. @kpeluso wdyt?

app/feemarket.go
guilherme-brandao: Nice find! I will consider this!

app/upgrades.go
spooktheducks: I think it's a great idea, let's do it

app/app.yaml
```diff
@@ -17,16 +17,20 @@ modules:
         - slashing
         # staking module is required if HistoricalEntries param > 0
         - staking
+        - feegrant
```
amimart: The `feegrant` doesn't have a `BeginBlock` func, it's not a problem to mention it but it'll help keep it clear
```suggestion
```

app/app.go
```diff
@@ -248,6 +260,46 @@ func NewAlloraApp(
 		return app.App.InitChainer(ctx, req)
 	})
 
+	// set denom resolver to test variant.
+	app.FeeMarketKeeper.SetDenomResolver(&feemarkettypes.TestDenomResolver{})
```
amimart: I thing it could be dangerous to use the test resolver that allows conversion from any currencies, we could instead copy the resolver from the hub: https://github.com/cosmos/gaia/blob/6159a47e1eed631e614b93595ca23f8783707dc6/app/keepers/keepers.go#L596

This one will disallow denom conversions, that's what we want I guess :)

app/app.yaml
```diff
@@ -45,12 +49,14 @@ modules:
...
         - genutil
         - authz
+        - feegrant
         - transfer
         - interchainaccounts
         - feeibc
         - upgrade
         - consensus
         - circuit
+        - feemarket
```
amimart, line 59:
```suggestion
        # feemarket must be after genutil to allow genesis transactions with 0fees
        - feemarket
```

Could be worth to document :)

app/app.yaml
guilherme-brandao: I don’t know why I added it there; Thanks for the correction.

app/app.go
guilherme-brandao: I think if we don't want the conversion to other currencies, we don't need a resolver, so we can just delete this line.

amimart: Oh yes you're right let's remove it!

amimart (approved):
Excellent, seems all good to me! Thanks :)

app/upgrades/v0_7_0/upgrades.go
```diff
@@ -15,13 +22,57 @@ const (
...
+
+	feeMarketParams, err := keepers.FeeMarketKeeper.GetParams(sdkCtx)
+	if err != nil {
+		return err
 	}
+
+	feeMarketParams.Enabled = true
+	feeMarketParams.FeeDenom = params.BaseCoinUnit
+	feeMarketParams.DistributeFees = true
+	feeMarketParams.MinBaseGasPrice = cosmosmath.LegacyMustNewDecFromStr("10")
```
amimart, line 68: We could also set the default values for `Window` & `MaxBlockUtilization`: https://github.com/skip-mev/feemarket/blob/f1f216e15aa6bcc05075a7079496764820a0f545/x/feemarket/types/eip1559.go

xmariachi:
Testing against this branch @guilherme-brandao , `gas-price` works, but `gas-prices` ends up in `nil pointer dereference`:  
```
$ allorad q feemarket gas-price uallo 
price:
  amount: "1.000000000000000000"
  denom: uallo

$ allorad q feemarket gas-prices      
Error: rpc error: code = Unknown desc = runtime error: invalid memory address or nil pointer dereference: panic
Usage:
  allorad query feemarket gas-prices [flags]
```

test/ibc/ibc_test.go
```diff
@@ -48,6 +48,7 @@ func TestIBCTestSuite(t *testing.T) {
 }
 
 func InitAlloraApp() (ibctesting.TestingApp, map[string]json.RawMessage) {
+	app.UseFeeMarketDecorator = false
```
kpeluso, line 51: Does x/feemarket have a non-obvious effect on ibc?

app/upgrades/v0_3_0/upgrades.go
```diff
@@ -15,11 +17,13 @@ const (
 var Upgrade = upgrades.Upgrade{
 	UpgradeName:          UpgradeName,
 	CreateUpgradeHandler: CreateUpgradeHandler,
+	StoreUpgrades:        storetypes.StoreUpgrades{Added: nil, Renamed: nil, Deleted: nil},
 }
 
 func CreateUpgradeHandler(
 	moduleManager *module.Manager,
 	configurator module.Configurator,
+	keepers *keepers.AppKeepers,
```
kpeluso, line 26: not sure if I like editing previous migrations vs just adding "ignore lint" comment (here and above and in other old files of migrations)

app/upgrades/v0_7_0/upgrades.go
```diff
@@ -15,13 +22,57 @@ const (
...
-		return moduleManager.RunMigrations(ctx, configurator, vm)
+
+		vm, err := moduleManager.RunMigrations(ctx, configurator, vm)
+		if err != nil {
+			return vm, err
+		}
+
+		if err := ConfigureFeeMarketModule(ctx, keepers); err != nil {
+			return vm, err
+		}
```
kpeluso: be it within `RunMigrations` or anything within the scope of `CreateUpgradeHandler`, we should have way more logging at every single step of this process. We need to know exactly how our migration is progressing during the migration/upgrade.

kpeluso (changes requested):
just missing some logs

app/upgrades/v0_7_0/upgrades.go
guilherme-brandao: If we want to set these properties to the default values, we don't need to declare them here; they will automatically be set to default.

app/upgrades/v0_3_0/upgrades.go
guilherme-brandao: What's your concern? Maybe I'm missing something, but I can't see any potential risks associated with this specific change.

app/upgrades/v0_7_0/upgrades.go
guilherme-brandao: Agreed! Adding more logs...

guilherme-brandao:
> Testing against this branch @guilherme-brandao , `gas-price` works, but `gas-prices` ends up in `nil pointer dereference`:
> 
> ```
> $ allorad q feemarket gas-price uallo 
> price:
>   amount: "1.000000000000000000"
>   denom: uallo
> 
> $ allorad q feemarket gas-prices      
> Error: rpc error: code = Unknown desc = runtime error: invalid memory address or nil pointer dereference: panic
> Usage:
>   allorad query feemarket gas-prices [flags]
> ```

@xmariachi This query is not available because we didn't enable the possibility to pay fees with other Denoms just `uallo`.

amimart:
> @xmariachi This query is not available because we didn't enable the possibility to pay fees with other Denoms just `uallo`.

I think the dereference error come from the missing `DenomResolver`, the query is expected to return the base fee denom & `DenomResolver.ExtraDenoms`, here is the result running on the hub:

```bash
> gaiad --node "https://cosmos-rpc.polkachu.com/" query feemarket gas-prices
prices:
- amount: "0.005000000000000000"
  denom: uatom
```

I think we should set a custom `DenomResolver` that returns no extra denoms

test/ibc/ibc_test.go
guilherme-brandao: There's nothing related to IBC in the fee market module docs. Also, I've just run some tests, and nothing caught my attention; everything is working as expected. In this specific test, if I enable the fee market, it will just require me to set a fee for the IBC transaction. I left the fee market disabled in this test for simplicity.

guilherme-brandao:
> > @xmariachi This query is not available because we didn't enable the possibility to pay fees with other Denoms just `uallo`.
> 
> I think the dereference error come from the missing `DenomResolver`, the query is expected to return the base fee denom & `DenomResolver.ExtraDenoms`, here is the result running on the hub:
> 
> ```shell
> > gaiad --node "https://cosmos-rpc.polkachu.com/" query feemarket gas-prices
> prices:
> - amount: "0.005000000000000000"
>   denom: uatom
> ```
> 
> I think we should set a custom `DenomResolver` that returns no extra denoms

@amimart Yeah, I've tried adding the default resolver that you suggested, but I'm having a problem setting it up. I think it's because of the dependency injection, which doesn't return a pointer to the keeper instance. As a result, the keeper instance isn't updated with the new resolver during initialization. To avoid this panic and provide a meaningful error message, we would need to initialize the app without dependency injection or find another workaround, which might involve replicating some of the code from the default module. I'm not sure if it's worth spending more time on this query, especially since it shouldn't be enabled. Lmk wyt/ if you have another strategy that we can use for this.

amimart:
> @amimart Yeah, I've tried adding the default resolver that you suggested, but I'm having a problem setting it up. I think it's because of the dependency injection, which doesn't return a pointer to the keeper instance.

Arf yes I see.. This is a copy of the keeper provided by the dependency injection

> To avoid this panic and provide a meaningful error message, we would need to initialize the app without dependency injection or find another workaround, which might involve replicating some of the code from the default module. I'm not sure if it's worth spending more time on this query, especially since it shouldn't be enabled. Lmk wyt/ if you have another strategy that we can use for this.

I think this is unfortunately used by the [Hermes IBC relayer](https://github.com/informalsystems/hermes), I'm afraid we need to manually wire the module.. I can work on that if you prefer

app/feemarket.go
```diff
@@ -0,0 +1,50 @@
...
+	if err := app.RegisterStores(
```
amimart: This logic could have been added to `legacy.go` with ibc wiring, no problem having it separated though :)

amimart (changes requested):
The feemarket module should also added here: https://github.com/allora-network/allora-chain/blob/060cd2acad7b3c23bc59bdd374ae5e69121caf07/app/legacy.go#L178

This is to register populate the related CLI commands

app/feemarket.go
guilherme-brandao: oh yeah, I agree!

guilherme-brandao: We don't need to add this interface registration that you mentioned; we already register interfaces during the `RegisterModules` function.

amimart (approved):
Looks all good to me thanks! 😉

fernandofcampos (approved):

kpeluso (approved):

PR #630, opened 2024-10-15
Use a pool to reduce allocations and GC operations on marshal calls

fernandofcampos:

Reduce memory allocations and GC operations by reusing buffers from a pool when we call marshal.

Inspired by https://github.com/cosmos/cosmos-sdk/blob/ee3d320eaa55e5aab1bf216f03a8d8aa5833d549/store/v2/internal/encoding/encoding.go#L140-L141

https://linear.app/alloralabs/issue/cosm/efficient-protobuf-serialization

## Still Left Todo

### DIFF
--- a/test/testutil/testdata.go
+++ b/test/testutil/testdata.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"strings"
+	"sync"
 
 	alloraMath "github.com/allora-network/allora-chain/math"
 	"github.com/allora-network/allora-chain/x/emissions/keeper"
@@ -13,6 +14,12 @@ import (
 	"github.com/cometbft/cometbft/crypto/secp256k1"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
+
+var valueBundleBufferPool = sync.Pool{
+	New: func() interface{} {
+		return make([]byte, 0, 1024)
+	},
+}
 
 func GenerateTestAccounts(count int) (
 	privKeys []secp256k1.PrivKey,
@@ -531,13 +538,15 @@ func GetNetworkLossFromCsv(
 }
 
 func signValueBundle(valueBundle *emissionstypes.ValueBundle, privateKey secp256k1.PrivKey) []byte {
-	src := make([]byte, 0)
-	src, err := valueBundle.XXX_Marshal(src, true)
+	buf := valueBundleBufferPool.Get().([]byte) // nolint:forcetypeassert
+	defer valueBundleBufferPool.Put(buf)        //nolint:staticcheck
+	buf = buf[:0]
+	marshaled, err := valueBundle.XXX_Marshal(buf, true)
 	if err != nil {
 		panic(err)
 	}
 
-	valueBundleSignature, err := privateKey.Sign(src)
+	valueBundleSignature, err := privateKey.Sign(marshaled)
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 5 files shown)


test/testutil/testdata.go
```diff
@@ -14,6 +15,12 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
+var valueBundleBufferPool = sync.Pool{
```
spooktheducks, line 18: We might want to use `&sync.Pool`, because the documentation states:

> A Pool must not be copied after first use.

I don't see anything in your code that would cause a copy, but if we use a reference, it's much less likely somebody will make a mistake in the future.

x/emissions/types/validations.go
```diff
@@ -11,6 +12,18 @@ import (
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
+var reputerValueBundleBufferPool = sync.Pool{
+	New: func() interface{} {
```
spooktheducks, line 16: We should encourage `any` instead of `interface{}` in our codebase

x/emissions/types/validations.go
```diff
@@ -212,9 +225,14 @@ func (bundle *WorkerDataBundle) Validate() error {
 	}
 
 	// Check signature from the bundle, throw if invalid!
-	src := make([]byte, 0)
-	src, _ = bundle.InferenceForecastsBundle.XXX_Marshal(src, true)
-	if !pubkey.VerifySignature(src, bundle.InferencesForecastsBundleSignature) {
+	buf := inferenceForecastsBundleBufferPool.Get().([]byte) // nolint:forcetypeassert
+	defer inferenceForecastsBundleBufferPool.Put(buf)        //nolint:staticcheck
+	buf = buf[:0]
```
spooktheducks, line 230: I wonder if we should make a wrapper type for `sync.Pool` that automatically zeroes the length of the buffers (and also does the type assertion).  That way nobody will forget to do this.

utils/syncpool_test.go
```diff
@@ -0,0 +1,90 @@
...
+		t.Fatal("Expected pool to be initialized")
+	}
+
+	x := pool.Get()
+	if x != 42 {
+		t.Errorf("Expected Get() to return 42, got %d", x)
+	}
+}
+
+func TestPool_GetPut(t *testing.T) {
```
spooktheducks, line 26: 5D chess, using an immutable type here

utils/syncpool_test.go
```diff
@@ -0,0 +1,90 @@
...
+	if s != "updated" {
+		t.Errorf("Expected 'updated', got '%s'", s)
+	}
+}
+
+func TestNewBytesPool(t *testing.T) {
+	size := 1024
+	bytesPool := NewBytesPool(size, 0)
+	if bytesPool == nil {
+		t.Fatal("Expected bytes pool to be initialized")
```
spooktheducks, line 49: can you use the https://github.com/stretchr/testify/require package to reduce the size of these assertion blocks?

spooktheducks (changes requested):
Really excited about this PR.  Love the `utils.Pool[T]` type, should prove to be very useful.  Only one tiny nitpick in the tests.

spooktheducks (approved):

utils/syncpool.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	// Reset
+	if p.reset != nil {
+		x = p.reset(x)
+	}
+
+	// Put it back in the pool
+	p.Pool.Put(x)
+}
```
kpeluso, line 30: Wouldn't expect a Put method to require client to enforce idempotency itself...

kpeluso (approved):

PR #631, opened 2024-10-16
Update proto builder version and other dependencies

fernandofcampos:

Current proto-builder version (0.14.0) is more than an year old. Other grpc and protobuf related dependencies are also outdated.

## Still Left Todo


relyt29 (approved):
I don't think it hugely matters, but maybe it would be better in the future to be able to regen the protos only for the latest proto version, so that the go files aren't regenerated or changed after an upgrade is done for that software version

kpeluso (approved):

spooktheducks (approved):

PR #636, opened 2024-10-21
set lossBlockHeight in GetNetworkInference

kpeluso:

Populates fields that was never filled in in any code path.

   - Adds missing functionality + unit test coverage
   - Not needed. Adds missing functionality to already-documented code.

## Still Left Todo

Ideally could add linter to check if a value is never filled in by any code path in every function.
Could be a follow-up ticket though.


xmariachi (approved):
Note: 
WIth this fix, in the `else` we populate with this value. However in the `if`, line 118, we are returning an uninitialised value for `lossBlockHeight`, while we return a valid (though limited) `networkInferences` object.
I understand we are returning an unititialised value on purpose, to signal that the `networkInference` has been calculated using the fallback, right?
Just want to make sure this is expected.

kpeluso:
yes @xmariachi as you said. if not present or 0, this value is not present or not involved in the calculation of network inference

PR #638, opened 2024-10-21
Fix forecast-implied inference handling in network inferences calculation

guilherme-brandao:
- Remove forecast-implied inference duplication in GetNetworkInference response
- Identify when the forecaster didn't make any forecasts for the actual inferers (0 forecast-implied inferences) -> avoid returning non-zero weights for these forecasters


x/emissions/keeper/inference_synthesis/network_inferences.go
```diff
@@ -314,15 +310,26 @@ func GetCalcNetworkInferenceArgs(
...
 	}
+
+	// If there are forecast-implied inferences, add forecasters info
+	// It will not have available forecast-implied inferences if the forecasters
+	// didn't make any forecasts for the existing inferers
+	if len(forecastImpliedInferencesByWorker) > 0 {
+		calcArgs.Forecasters = sortedForecasters
+		calcArgs.ForecasterToForecast = forecasterToForecast
+		calcArgs.ForecasterToRegret = forecasterToRegret
+		calcArgs.ForecasterToForecastImpliedInference = forecastImpliedInferencesByWorker
```
xmariachi, line 331: in case len > 0, is no-nil guaranteed in the other maps? (sortedForecasters, forecasterToForecast, ...)

x/emissions/keeper/queryserver/query_server_inferences.go
```diff
@@ -231,7 +226,7 @@ func (qs queryServer) GetConfidenceIntervalsForInferenceData(
 	for _, forecast := range networkInferences.ForecasterValues {
 		weight, exists := forecasterWeights[forecast.Worker]
 		if exists {
-			inferences = append(inferences, forecastImpliedInferenceByWorker[forecast.Worker].Value)
+			inferences = append(inferences, forecast.Value)
```
xmariachi, line 229: In the place where we used to have a nil `*Inference`, we now will not have that - but may have an empty `Dec`. Is that an OK solution?

xmariachi (commented):
lgtm in that removes a duplicity and should fix an NPE; but commenting on second considerations

x/emissions/keeper/inference_synthesis/network_inferences.go
guilherme-brandao: yes!

x/emissions/keeper/queryserver/query_server_inferences.go
guilherme-brandao: yeah, since `forecastImpliedInferenceByWorker` was exactly `networkInferences.ForecasterValues` converted to map[string]*Inference. I don't think we have any problems here.

xmariachi: What I mean here is that we sometimes had a `nil`  `*Inference`, right? 
Now, what would be the expected behaviour for the case that previously generated that nil `*Inference` ? 
How are we expecting it to behave in that case now?

Does that generating a no-entry on `ForecastValues` ? 
Or an entry with an empty `Dec`? 
And how is that handled downstream?

guilherme-brandao: I don't think we could have a nil inference here, actually. What is in ForecasterValues is exactly the existing forecast-implied inferences, we don't do any fallback for an empty value.

kpeluso (approved):
I'm a bit confused by some of the git commits -- re-reviewing some of Bryn's changes as part of this PR, which merges back to Bryn's PR -- but must be history-related.

Ofc conflicts needs to be tended to but signaling my approval now.

xmariachi (approved):

kpeluso (approved):

PR #646, opened 2024-10-22
Compile the linter with the same dependencies as golangci-lint

fernandofcampos:

Plugins and golangci-lint must be compiled with the exact same stack, it's related to how Go plugins work.

Create a different go.mod file for the linter so we can control it.

## Still Left Todo


guilherme-brandao (approved):

xmariachi (approved):

relyt29 (approved):

Makefile
```diff
@@ -88,7 +88,7 @@ lint:
 
 build-maprange-linter:
```
relyt29, line 89: above we add check-defer-close to `make lint`

why are we having a separate make command for maprange linter? shouldn't we just always run all the linters?

relyt29 (approved):

Makefile
fernandofcampos: This linter lists all map ranges, and we decide case by case if it's a correct usage or not. To always run it we would need to add a //nolint comment to the cases that we consider licit.

relyt29: I see, it's just a fancier `grep` then?

relyt29: I guess it's more sophisticated than grep because it understands whether the thing being ranged over is a map or (some other rangeable data type)

relyt29: is there a single instance where ranging over a map is appropriate? can we turn this on and use nolint? why don't you want to do that by default if not?

fernandofcampos: Yes, more sophisticated than a single grep. And yes, most of our ranges over maps are appropriate. 

Even when we want to access the elemets in order, for example, we usually range to get all keys, and then we sort the keys, and then access the values in order.

I think using //nolint and running this with the other linters is a good idea, but we should do it in a separated task in my opinion.

relyt29 (approved):
approved given conversation above

Makefile
fernandofcampos: Btw, Cosmos has a more sophisticated linter that we can use/port: 

https://github.com/cosmos/gosec/blob/master/rules/sdk/iterate_over_maps.go

// This pass enforces ONLY key retrieval from maps. It resolves a problem that was
// discovered in the Cosmos-SDK in which maps were being iterated on by key and value
// and that produced non-determinism in upgrades.

PR #650, opened 2024-10-23
Make fuzzer setup run through every state transition once before fuzzing starts

relyt29:

This PR does two things:

1. Relaxes an invariant check. Before we checked that the sum of reward debt for all actors is exactly equal to the amount of tokens stored in the `AlloraPendingRewardForDelegatorAccountName` module account. Now instead we only check that the value is _greater_ than the sum of all reward debt. Rounding errors from integer division make perfectly checking the invariant for equality difficult. The real concern is that we should always have enough money to pay every delegator their fair share, which is a greater than check rather than an equals.
2. Makes the fuzzer setup run through every state transition at least once before fuzzing actually starts. Before this PR, we do have a fuzzing setup, but the thought originally was to just make it so the fuzzer is in a suitable state to start. Now in addition, we do every transition at least once to prove that all the transitions are working, before the fuzzer goes to town.

PROTO-2753

## Are these changes tested and documented

These changes have been tested by manual interaction with the fuzzer. The fuzzer Readme.MD file has been updated for documentation purposes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,7 +61,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-[#652](https://github.com/allora-network/allora-chain/pull/652) Reduce code duplication, set local_testnet_upgrade_l1.sh using environment variables with local_testnet_l1.sh instead #652 
+* [#652](https://github.com/allora-network/allora-chain/pull/652) Reduce code duplication, set local_testnet_upgrade_l1.sh using environment variables with local_testnet_l1.sh instead
+* [#650](https://github.com/allora-network/allora-chain/pull/650) Make fuzzer setup run through every state transition once before fuzzing starts
+
+### Changed
 
 ### Deprecated
 
@@ -101,7 +104,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [#603](https://github.com/allora-network/allora-chain/pull/603) Validate all Values for Correctness Prior to Storing them in the Keeper
 * [#620](https://github.com/allora-network/allora-chain/pull/620) Add a static analyzer to detect non-deferred `.Close()` calls, improve migration error handling
 * [#622](https://github.com/allora-network/allora-chain/pull/622) Add telemetry metrics on queries/txs
-
 
 ## v0.5.0
 
--- a/test/invariant/actors_setup_test.go
+++ b/test/invariant/actors_setup_test.go
@@ -11,7 +11,78 @@ import (
 	sdktypes "github.com/cosmos/cosmos-sdk/types"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	"github.com/ignite/cli/v28/ignite/pkg/cosmosaccount"
+	"github.com/stretchr/testify/require"
 )
+
+// set up the common state for the simulator
+// prior to either doing random simulation
+// or manual simulation
+func simulateSetUp(
+	m *testcommon.TestConfig,
+	numActors int,
+	epochLength int,
+	mode SimulationMode,
+) (
+	faucet Actor,
+	simulationData *SimulationData,
+) {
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -57,6 +57,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## v0.7.0
```
xmariachi: This line with `v0.7.0` should not be here, it's just `[Unreleased]` because we don't know which next version will be.

test/invariant/actors_setup_test.go
```diff
@@ -153,3 +224,336 @@ func (a *Actor) GetBalance(m *testcommon.TestConfig) (cosmossdk_io_math.Int, err
...
+		for _, topicId := range listTopics {
+			registerWorker(m, worker, UnusedActor, nil, topicId, data, iterationCount)
+			iterationCount++
+		}
+	}
+	return iterationCount
+}
+
+// starteDelegateDelegators delegates a list of delegators to a list of reputers on a list of topics.
+func starteDelegateDelegators(
```
xmariachi: typo `starte`

test/invariant/transitions_test.go
```diff
@@ -294,19 +294,9 @@ func pickActorAndTopicIdForStateTransition(
...
-				}
-				reputer, err := data.pickRandomReputerRegisteredInTopic(m.Client.Rand, topicId)
-				if err != nil {
-					continue
-				}
-				return true, worker, reputer, nil, topicId
-			}
+			randIndex := m.Client.Rand.Intn(len(topics))
+			topicId := topics[randIndex].Id
+			return true, UnusedActor, UnusedActor, nil, topicId
```
xmariachi, line 299: just curious, why `UnusedActor` here?

xmariachi (commented):
lgtm in general, left some comments though

And one comment out of scope of this PR: 

In the fuzzer, we generate values eg in "happy" range. eg `invariants_test.go:142`
```
Value:   alloraMath.NewDecFromInt64(int64(m.Client.Rand.Intn(51) + 50))
``` 
Should we also allow more variability ( and especially generate zeroes) for a wider coverage?

relyt29:
> Should we also allow more variability ( and especially generate zeroes) for a wider coverage?

yes, but maybe in a future PR, there's a lot to do already :sob:

CHANGELOG.md
relyt29: oh ok, didn't realize that

test/invariant/transitions_test.go
relyt29: doInferenceAndReputation doesn't pick a particular actor to do the inference and reputation

it uses every registered actor in the topic, makes every single person submit a worker and reputer payload

yes this could be changed in a future release for better coverage, but at a minimum is how the fuzzer has worked historically

x/emissions/keeper/invariants.go
```diff
@@ -314,100 +314,112 @@ func StakingInvariantSumStakeFromStakeReputerAuthorityEqualTotalStakeAndTopicSta
...
+		// get the sum of all accumulated debts that happened to be staked upon a reputer
+		for ; delegatedStakesIter.Valid(); delegatedStakesIter.Next() {
+			keyValue, err := delegatedStakesIter.KeyValue()
 			if err != nil {
 				panic("failed to get key value from delegatedStakes iterator")
 			}
-			topicAndReputer := TopicAndReputer{topicId: keyValue.Key.K1(), reputer: keyValue.Key.K3()}
-			rewardDebtSum, has := rewardDebtSums[topicAndReputer]
+			topicId := keyValue.Key.K1()
+			// delegator := keyValue.Key.K2() // unused
```
guilherme-brandao: delete(?)

guilherme-brandao (dismissed):
looks good!

xmariachi (approved):

guilherme-brandao (approved):

PR #652, opened 2024-10-23
Reduce code duplication, set local_testnet_upgrade_l1.sh using environment variables with local_testnet_l1.sh instead

relyt29:

This is a refactor that removes the `local_testnet_l1_upgrade.sh` file and instead implements its functionality in `local_testnet_l1.sh` with environment variables that control whether the upgrade is set up or not. Seems better from a code duplication / reuse perspective. I used the diff between the two files to ensure that the upgrade code was copied into the original testnet script.

PROTO-2746 is what caused me to observe this, but it was actually already fixed, so a tweak for that issue around the stake removal delay window is also included in this PR

No actual code is changed so our existing tests should be adequate


xmariachi (dismissed):
nice

guilherme-brandao (approved):

xmariachi (approved):

PR #653, opened 2024-10-23
Fuzzer Bugfixes, Allow User to Set Fuzzer Transition Probability Distribution

relyt29:

This PR does three things:

1. Squashes bugs in the fuzzer that was causing it to crash (in several cases where the fuzzer thought that it had the wrong state as compared to the chain).
2. Changes the fuzzer to (in addition to environment variables) takes its configuration from a config.json file in the fuzz directory
3. Allows the user to specify a probability distribution to follow for the state transitions the fuzzer will attempt. By setting the `transitionWeights` configuration parameter, the fuzzer will now choose state transitions following the percentage likelihood for each transition. E.g. if you set the weight of createTopic to 20%, then the fuzzer will, on every iteration, have a 20% likelihood of picking a createTopic as the next state transition it will attempt. This allows the user to guide fuzzing to an extent to try to probe specific parts of the codebase more thoroughly.

PROTO-2289

These changes have been tested manually by running the fuzzer on my local computer. Updates to the fuzzer documentation have been put in the README.md in the fuzzer folder. The changelog has been updated.


relyt29:
using this you can do cool things like force testing of unstake and cancel operations over everything else:

```
   fuzz_test.go:197: State Transitions Summary: {                                                                     
        createTopic: 2,                                                                                                
        fundTopic: 2,                                                                                                  
        registerWorker: 10,                                                                                            
        registerReputer: 8,                                                                                            
        unregisterWorker: 2,                                                                                           
        unregisterReputer: 1,                                                                                          
        stakeAsReputer: 541                                                                                            
        delegateStake: 559                                                                                             
        unstakeAsReputer: 492 
        undelegateStake: 523
        cancelStakeRemoval: 256
        cancelDelegateStakeRemoval: 239
        collectDelegatorRewards: 4
        doInferenceAndReputation: 2
        }
```

test/fuzz/README.md
```diff
@@ -5,33 +5,35 @@ The fuzz tests in this repo send random transactions to the chain in order to tr
 Example invocation:
 
 ```bash
-FUZZ_TEST=TRUE SEED=1 RPC_MODE="SingleRpc" \
-    RPC_URLS="http://localhost:26657" \
-    MAX_ITERATIONS=100 MODE="alternate" \
-    NUM_ACTORS=12 EPOCH_LENGTH=14 \
-    /usr/bin/go test -timeout 15m -run ^TestFuzzTestSuite$ -v ./test/fuzz
+FUZZ_TEST=TRUE go test -timeout 15m -run ^TestFuzzTestSuite$ -v ./test/fuzz
```
xmariachi, line 9: If you just run this against a basic, `allorad start` network, it won't work (complains about not having a `faucet` acct, and I guess about others too?). 
Since these requirements are not specified, this should explain how to run it against the local network that this is configured to run against.

```suggestion
bash test/local_testnet_l1.sh
FUZZ_TEST=TRUE go test -timeout 15m -run ^TestFuzzTestSuite$ -v ./test/fuzz
```

xmariachi (dismissed):
One change suggested on README after trying it.
Preemptive approval

test/fuzz/README.md
relyt29: ah good point thank you

test/fuzz/registration_test.go
```diff
@@ -39,31 +38,33 @@ func registerWorker(
...
+		return false
 	}
 
 	_, err = m.Client.WaitForTx(ctx, txResp.TxHash)
 	requireNoError(m.T, data.failOnErr, err)
-	wasErr = orErr(wasErr, err)
+	if err != nil {
+		iterFailLog(m.T, iteration, "failed to register ", actor, "as worker in topic id ", topicId, "tx wait error", err)
+		return false
+	}
```
guilherme-brandao: why both `requireNoError` and `err != nil` are required here?

guilherme-brandao (dismissed):
Looks good! nice work!

test/fuzz/registration_test.go
relyt29: requireNoError is poorly named, I will rename it

the idea is if the data.failOnErr is true, end the whole test and crash because we're supposed to be in behaved mode and we don't expect errors in that mode. If we are not in beahaved mode, continue, and return false to let the caller know the attempt failed

xmariachi (approved):

guilherme-brandao (approved):

PR #654, opened 2024-10-24
Reorganize Linter Folder, add linter to check fuzzer state transition probabilities add to 100 percent

relyt29:

This PR does two things:

1. I observed that @spooktheducks's defer close linter and @fernandofcampos's map range linters were in two separate directories which makes no sense. Now all the linters are in the `linters` folder
2. Adds a linter that is necessary for the fuzzer, that makes sure that the probability distribution of likelihoods of picking any given state transition (e.g. create topic, fund topic, register as worker, yada yada) all adds up to 100%.

Relevant to PROTO-2289 but not directly a fix for it.

Tested by hand that the existing linters run correctly as intended. Tested by hand the new linter. Both the old and new linters are all also tested by our CI suite.


relyt29:
integrating into #653

PR #663, opened 2024-11-01
Reintroduce whitelists on global and topic levels

kpeluso:

Whitelists reintroduced so protocol vanguards can better nurture our budding network.

This file is the best entrypoint into understanding everything I did at the most raw level. Every other change is a consequence of the functions I created here:
```
x/emissions/keeper/whitelist.go
```

Roles (permissions) of existing whitelist has been expanded and new whitelists and related abstractions have been introduced:
* whitelist admin
   * can add others to whitelists
   * can change global params (as is already the case today)
   * moderated by super admin
* global whitelist
   * can participate in all roles across topics except super admin (cannot add other admins nor update global params)
   * moderated by super admin
* topic creator whitelist
   * can create topics
   * moderated by super admin
* per topic worker whitelist
   * can participate in given topic as worker
   * moderated by topic creator and super admin
* per topic reputer whitelist
   * can participate in given topic as reputer
   * moderated by topic creator and super admin
* feature flag via global params to turn on/off whitelist entirely per whitelist
   * global topic creator and global actor whitelist feature flag, toggled by super admins via new global params
   * per topic worker whitelist feature flag
   * per topic reputer whitelist feature flag
   * these 2^^ per-topic feature flags toggled by topic creator and super admin
   * feature flag = on => whitelist checked ;; off => anyone can attempt to participate
* minor organizations: whitelist methods moved to their own file, out of crowded `keeper.go`
* whitelist abstractions
   * `setters` - update state for some whitelist
   * `getters` - read what a whitelist's state
   * `permissions` - actions someone can take, if permitted
   * `qualifiers` - helper functions for `permissions` that wrap `getters`

This PR also creates `x/emissions` module `v6`, `allorad` chain `v0.7.0`, and upgrade/migration logic for `v0.6.0 -> v0.7.0`. This is used to ensure all topic whitelists are turned on, that all new txs and queries are properly versioned, and that new global parameters are properly initialized.

## Breaking Changes for Clients

Indexer, Producer, offchain-node, Labs' BE, and other clients need to update their endpoint URLs `.../emissions/v5/...` to `.../emissions/v6/...`

https://linear.app/alloralabs/issue/PROTO-2678/reintroduce-whitelists

   - Unit tests created
   - Documentation release should be coordinated with feature enablement on appropriate chain. Will coordinate with @kush-alloralabs 

## Still Left Todo

Done.


x/emissions/keeper/msgserver/msg_server_worker_payload_test.go
```diff
@@ -159,7 +165,11 @@ func (s *MsgServerTestSuite) TestMsgInsertWorkerPayloadNotFailsWithNilInference(
 	blockHeight := workerMsg.WorkerDataBundle.InferenceForecastsBundle.Forecast.BlockHeight
 	ctx = ctx.WithBlockHeight(blockHeight)
 
-	_, err := msgServer.InsertWorkerPayload(ctx, &workerMsg)
+	// Add worker to topic whitelist
+	err := s.emissionsKeeper.AddToTopicWorkerWhitelist(ctx, workerMsg.WorkerDataBundle.TopicId, workerMsg.WorkerDataBundle.Worker)
```
xmariachi: Do we have "negative" tests? 
Could be done just like 
```
	_, err := msgServer.InsertWorkerPayload(ctx, &workerMsg)
	require.Error(err, "InsertWorkerPayload should return an error before adding worker to whitelist")

	// Add worker to topic whitelist
	err = s.emissionsKeeper.AddToTopicWorkerWhitelist(ctx, workerMsg.WorkerDataBundle.TopicId, workerMsg.WorkerDataBundle.Worker)
	require.NoError(err, "InsertWorkerPayload should not return an error after adding worker to whitelist")
```

x/emissions/proto/emissions/v5/query.proto
```diff
@@ -429,6 +429,167 @@ service QueryService {
...
+
+message IsWhitelistedTopicCreatorRequest {
+  string address = 1;
+}
+
+message IsWhitelistedTopicCreatorResponse {
+  bool is_whitelisted_topic_creator = 1;
+}
+
+message IsWhitelistGlobalActorRequest {
```
xmariachi: Inconsistency: "whitelist" vs "whitelisted"
isWhitelist`ed`TopicCr...
isWhitelistGlob...

x/emissions/keeper/msgserver/msg_server_topics.go
```diff
@@ -66,6 +77,16 @@ func (ms msgServer) CreateNewTopic(ctx context.Context, msg *types.CreateNewTopi
 		return nil, err
 	}
 
+	// Turn topic whitelist on by default so no one can squeeze in payloads before an admin notices or can act
```
xmariachi, line 80: :+1:

xmariachi (dismissed):
Good work. Current state lgtm, so I approve. 
However, find some comments in the code, plus these general comments: 

* Migration: This is a new minor version, `v0.7.0`, shouldn't the changes be in v6? Keeping them in v5 (which has already been migrated to) could be confusing going fw.
* 
* Just a comment on an alternative way, maybe more canonical/universal, to impl roles/permissions: via defining a unified permissions list per `address` and  `(address, topic)`, defining roles separately, so each entry can have a set of roles.
However because it is not expected this set of roles to grow (right?), the current impl is simpler and should also be slightly more performant at the expense of a bit more of storage space.

kpeluso:
Thank you for the review @xmariachi . Some responses to your overall comments:

> Migration: This is a new minor version, v0.7.0, shouldn't the changes be in v6? Keeping them in v5 (which has already been migrated to) could be confusing going fw.

My thinking was that these were strictly additions and we only technically _have to_ bother with a migration (more complexity) whenever we touch existing protos and properties.

> Just a comment on an alternative way, maybe more canonical/universal, to impl roles/permissions: via defining a unified permissions list per address and (address, topic), defining roles separately, so each entry can have a set of roles.
However because it is not expected this set of roles to grow (right?), the current impl is simpler and should also be slightly more performant at the expense of a bit more of storage space.

Exactly. It is simpler, the storage implied by these lists are bounded by protocol vanguard activity, and no, we don't expect this set of roles to grow -- this already covers everything except delegation and not-unique-to-allora actions e.g. sending ALLO.

x/emissions/types/errors.go
```diff
@@ -32,7 +32,7 @@ var (
 	ErrInsufficientStakeAfterRemoval            = errors.Register(ModuleName, 25, "insufficient stake after removal")
 	ErrFundAmountTooLow                         = errors.Register(ModuleName, 26, "topic fund amount too low")
 	ErrIntegerUnderflowUnmetDemand              = errors.Register(ModuleName, 27, "integer underflow for unmet demand")
-	ErrNotWhitelistAdmin                        = errors.Register(ModuleName, 28, "not whitelist admin")
+	ErrNotPermittedToUpdateWhitelistAdmins      = errors.Register(ModuleName, 28, "not permitted to update whitelist admins")
```
relyt29: If i remember correctly this is a breaking change for (not doing a) migration, errors are stored in the state machine

I think safer / better to remove references to the error registered with value 28 and instead create a new error that matches your needs

relyt29:
> My thinking was that these were strictly additions and we only technically have to bother with a migration (more complexity) whenever we touch existing protos and properties.

I think we should do this as a full on migration. It's adding a non-trivial amount of changes to protos and errors and quite a few other things. I think it's more appropriate to do non-x/upgrade migrations for hotfixes and small changes, this is not, changes the security model quite a bit

x/emissions/keeper/genesis.go
```diff
@@ -689,17 +689,20 @@ func (k *Keeper) InitGenesis(ctx context.Context, data *types.GenesisState) erro
 			}
 		}
 	}
-	//CoreTeamAddresses []string
-	if len(data.CoreTeamAddresses) != 0 {
+	// CoreTeamAddresses, WhitelistAdmins []string
+	// This allows us to add core team addresses to the whitelist during a genesis import
+	// while still keeping the original core team addresses in the genesis file
+	if len(data.CoreTeamAddresses) != 0 || len(data.WhitelistAdmins) != 0 {
 		// make sure what we are storage isn't garbage
```
relyt29:
```suggestion
		// make sure what we are storing isn't garbage
```

x/emissions/proto/emissions/v5/params.proto
```diff
@@ -152,4 +152,7 @@ message Params {
     (gogoproto.customtype) = "github.com/allora-network/allora-chain/math.Dec",
     (gogoproto.nullable) = false
   ];
+  bool global_whitelist_enabled = 50; // global whitelist enabled => all global whitelisted actors can create topics
```
relyt29: yeah we're modifying the params type, I think it is inappropriate not to do a migration even though we could (probably) get away without it by abusing the way golang unmarshals data

It just feels more risky and irresponsible when migrations of just the params object is easy anyways

x/emissions/proto/emissions/v5/params.proto
```diff
@@ -152,4 +152,7 @@ message Params {
...
+  // and participate in all topics as workers and reputers
+  bool topic_creator_whitelist_enabled = 51; // topic creator whitelist enabled => only topic creator whitelisted actors can create topics
```
relyt29: isn't this missing a flag for the per topic reputer and worker whitelists being enabled or not?

xmariachi:
> > My thinking was that these were strictly additions and we only technically have to bother with a migration (more complexity) whenever we touch existing protos and properties.
> 
> I think we should do this as a full on migration. It's adding a non-trivial amount of changes to protos and errors and quite a few other things. I think it's more appropriate to do non-x/upgrade migrations for hotfixes and small changes, this is not, changes the security model quite a bit

I agree with @relyt29 .
It also adds clarity on knowing what is being added on each migration / compat-breaking version
Also, modifying this migration would make a migration from 0.5.0 to 0.6.0 inconsistent (only 0.5.0 to 0.7.0 would be consistent) so there would be no way to get 0.6.0 state by migrations. imho no need to lose this traceability.

kpeluso:
@xmariachi @relyt29 I don't agree. This is testnet and we should expect to wipe previous migrations for mainnet, as well as module versions.

Our modules are not in sync already as `x/emissions` is v5, `x/mint` is less, and the chain is `v0.6.0`, and that's fine.

Traceability isn't a concern -- git history isn't going anywhere, and that's used for traceability more than testnet migrations that we'll probably wipe anyway. I think, however, you mean traceability in a way too closely resembling how databases migrate -- it's a clear analogy but at the same time, we'll never downgrade, so we don't need "reversabilitiy."

Strict additions to protos that clients don't touch need not be viewed as breaking, as it's breaking nothing about existing functionality for clients the chain cares to support atm. In fact, if we bump `x/emissions`, that would break all API clients regardless if they care about strictly typed global parameters or whitelists as they'd need to bump `v5->v6` in their urls (which is yet another reason to use versioned protos instead of versioned modules imo).

x/emissions/keeper/msgserver/msg_server_reputer_payload.go
```diff
@@ -20,6 +20,12 @@ func (ms msgServer) InsertReputerPayload(ctx context.Context, msg *types.InsertR
 	if err != nil {
 		return nil, err
 	}
+	canSubmit, err := ms.k.CanSubmitReputerPayload(ctx, msg.ReputerValueBundle.ValueBundle.TopicId, msg.Sender)
+	if err != nil {
+		return nil, err
+	} else if !canSubmit {
+		return nil, types.ErrNotPermittedToSubmitReputerPayload
+	}
```
guilherme-brandao: Stateless validations always come first... I'd move this below `msg.ReputerValueBundle.Validate()`

x/emissions/keeper/msgserver/msg_server_topics.go
```diff
@@ -13,6 +13,17 @@ import (
...
+	// Validate the address
+	if err := ms.k.ValidateStringIsBech32(msg.Creator); err != nil {
+		return nil, err
+	}
+	canCreate, err := ms.k.CanCreateTopic(ctx, msg.Creator)
+	if err != nil {
+		return nil, err
+	} else if !canCreate {
+		return nil, types.ErrNotPermittedToCreateTopic
+	}
```
guilherme-brandao, line 25: Same here... stateless validations first!

x/emissions/keeper/msgserver/msg_server_topics.go
```diff
@@ -66,6 +77,16 @@ func (ms msgServer) CreateNewTopic(ctx context.Context, msg *types.CreateNewTopi
...
 
+	// Turn topic whitelist on by default so no one can squeeze in payloads before an admin notices or can act
+	err = ms.k.EnableTopicWorkerWhitelist(ctx, topicId)
+	if err != nil {
+		return nil, err
+	}
+	err = ms.k.EnableTopicReputerWhitelist(ctx, topicId)
+	if err != nil {
+		return nil, err
+	}
```
guilherme-brandao: Why not add this as a message property instead of enabling it by default? I don't think it's a big deal, just curious to know the reasoning

x/emissions/keeper/keeper.go
```diff
@@ -223,7 +223,17 @@ type Keeper struct {
...
+	globalWhitelist       collections.KeySet[ActorId] // actors who can create topics and participate in all topics
+	topicCreatorWhitelist collections.KeySet[ActorId] // whitelist of actors who can create topics
+	// map from topic id to whitelist of workers
+	topicWorkerWhitelist collections.KeySet[collections.Pair[TopicId, ActorId]]
+	// map from topic id to whitelist of reputers
+	topicReputerWhitelist collections.KeySet[collections.Pair[TopicId, ActorId]]
+	// map from topic id to whether the whitelist is enabled for topic workers
+	topicWorkerWhitelistEnabled collections.KeySet[TopicId]
+	// map from topic id to whether the whitelist is enabled for topic reputers
+	topicReputerWhitelistEnabled collections.KeySet[TopicId]
```
guilherme-brandao, line 236: Why this is not a topic property?

x/emissions/keeper/keeper.go
```diff
@@ -223,7 +223,17 @@ type Keeper struct {
 
 	/// WHITELISTS
 
-	whitelistAdmins collections.KeySet[ActorId]
+	whitelistAdmins       collections.KeySet[ActorId] // can change parameters and decide who can be included in any whitelist
+	globalWhitelist       collections.KeySet[ActorId] // actors who can create topics and participate in all topics
+	topicCreatorWhitelist collections.KeySet[ActorId] // whitelist of actors who can create topics
```
guilherme-brandao, line 228: These could be just one KV

x/emissions/keeper/msgserver/msg_server_whitelist.go
```diff
@@ -40,18 +43,334 @@ func (ms msgServer) RemoveFromWhitelistAdmin(ctx context.Context, msg *types.Rem
...
+		return nil, err
+	}
+
+	// Check that sender is permitted to update global whitelists
+	canUpdate, err := ms.k.CanUpdateGlobalWhitelists(ctx, msg.Sender)
+	if err != nil {
+		return nil, err
+	} else if !canUpdate {
+		return nil, types.ErrNotPermittedToUpdateGlobalWhitelist
+	}
```
guilherme-brandao: here in the whitelist msgs, sometimes you're doing stateless first, then stateful, and others the opposite, I'd standardize to the former.

x/emissions/keeper/msgserver/msg_server_whitelist.go
```diff
@@ -40,18 +43,334 @@ func (ms msgServer) RemoveFromWhitelistAdmin(ctx context.Context, msg *types.Rem
...
+	// Validate the address
+	if err := ms.k.ValidateStringIsBech32(msg.Sender); err != nil {
+		return nil, err
+	}
+	canUpdate, err := ms.k.CanUpdateTopicWhitelist(ctx, msg.TopicId, msg.Sender)
+	if err != nil {
+		return nil, err
+	} else if !canUpdate {
+		return nil, types.ErrNotPermittedToUpdateTopicWhitelist
+	}
```
guilherme-brandao: Here, you're checking if the topic exists and then querying it again inside `CanUpdateTopicWhitelist.` You can just do one query to check both topic's existence, then topic creator. Please do the same thing in the others.

x/emissions/keeper/whitelist.go
```diff
@@ -0,0 +1,308 @@
...
+func (k *Keeper) CanSubmitWorkerPayload(ctx context.Context, topicId TopicId, actor ActorId) (bool, error) {
+	has, err := k.IsEnabledTopicWorker(ctx, topicId, actor)
+	if err != nil {
+		return false, err
+	}
+	if has {
+		return true, nil
+	}
+	return k.IsEnabledGlobalActor(ctx, actor)
+}
```
guilherme-brandao: `isTopicCreator` -> `has`

guilherme-brandao (changes requested):

relyt29 (dismissed):
not done but will leave this review for now (from 2 days ago)

x/emissions/proto/emissions/v5/params.proto
kpeluso: no that's handled per topic

kpeluso: We could consider adding something to globally turn on all whitelists, or an emergency switch to shut off all activity entirely. @spooktheducks @relyt29 thoughts? imo not the biggest fan of this as it could make a chilling effect but could be fine for a young network.

x/emissions/keeper/whitelist.go
kpeluso: will change the latter to match the former, not the former to match the latter

x/emissions/keeper/msgserver/msg_server_whitelist.go
kpeluso: Just have to be sure to catch a particular "not found" error first but then should avoid an extra query, yes.

x/emissions/keeper/keeper.go
kpeluso: If we map from address to all the roles they could possibly hold at the global level, yes could be 1 KV. I think that's harder to manage though, and also means more has to be marshaled at each query. So I don't agree with doing this as I understand it.

x/emissions/keeper/keeper.go
kpeluso: Topic bloated as is imo and we should try query only what we need per suggestion from Marko => less to unmarshall for each query.

x/emissions/keeper/msgserver/msg_server_topics.go
kpeluso: Good call out. Was just simpler this way without a tangible UX loss. I can add if people feel strongly though.

x/emissions/types/errors.go
kpeluso: yeah fair I'll do it like that

kpeluso: yeah fair I'll do it like that jus tot be safe

x/emissions/proto/emissions/v5/params.proto
kpeluso: imo it's: "feeling ok with ourselves as devs by doing the more dev correct thing" vs. "breaking everyone's client for something 99% of them don't care about". That's how I'd summarize my longer comment above.

x/emissions/keeper/msgserver/msg_server_worker_payload_test.go
kpeluso: great call. just added

x/emissions/keeper/msgserver/msg_server_topics.go
kpeluso: The only stateless check in this function is validation of `msg.Creator` string, and that precedes this highlighted code.

x/emissions/keeper/msgserver/msg_server_topics.go
kpeluso: added params to topic creation

x/emissions/proto/emissions/v5/params.proto
kpeluso: made migration

x/emissions/keeper/msgserver/msg_server_worker_payload_test.go
```diff
@@ -137,7 +137,13 @@ func (s *MsgServerTestSuite) TestMsgInsertWorkerPayload() {
 	blockHeight := workerMsg.WorkerDataBundle.InferenceForecastsBundle.Forecast.BlockHeight
 
 	ctx = ctx.WithBlockHeight(blockHeight)
-	_, err := msgServer.InsertWorkerPayload(ctx, &workerMsg)
+
+	err := s.emissionsKeeper.AddToTopicWorkerWhitelist(ctx, topicId, workerMsg.WorkerDataBundle.Worker)
+	require.NoError(err)
```
xmariachi, line 142:
```suggestion
	require.NoError(err)
```

x/emissions/keeper/msgserver/msg_server_worker_payload_test.go
```diff
@@ -137,7 +137,13 @@ func (s *MsgServerTestSuite) TestMsgInsertWorkerPayload() {
...
+	err = s.emissionsKeeper.AddToTopicWorkerWhitelist(ctx, workerMsg.WorkerDataBundle.TopicId, workerMsg.WorkerDataBundle.Worker)
```
xmariachi: These two lines seem redundant.
`workerMsg.WorkerDataBundle.TopicId` should be same as `TopicId` . If we want to check this then it could have its own assertion.

x/emissions/keeper/msgserver/msg_server_worker_payload_test.go
```diff
@@ -220,27 +241,38 @@ func (s *MsgServerTestSuite) TestMsgInsertWorkerPayloadFailsWithoutSignature() {
...
+	topicId := uint64(123)
 
 	workerPrivateKey := secp256k1.GenPrivKey()
 
 	workerMsg, _ := s.setUpMsgInsertWorkerPayload(workerPrivateKey)
 
+	// Add worker to topic whitelist
+	err := s.emissionsKeeper.AddToTopicWorkerWhitelist(ctx, topicId, workerMsg.WorkerDataBundle.Worker)
+	require.NoError(err)
+	err = s.emissionsKeeper.AddToTopicWorkerWhitelist(ctx, workerMsg.WorkerDataBundle.TopicId, workerMsg.WorkerDataBundle.Worker)
```
xmariachi, line 264: Duplicated two lines too?
If the purpose is to ensure that topicId is the same as workerMsg.WorkerDataBundle.TopicId, that should merit its own assertion.

x/emissions/keeper/msgserver/msg_server_worker_payload_test.go
```diff
@@ -404,15 +447,21 @@ func (s *MsgServerTestSuite) TestMsgInsertWorkerPayloadFailsWithMismatchedForeca
 	forecastsCount0 := s.getCountForecastsAtBlock(originalTopicId, blockHeight)
 	require.Equal(forecastsCount0, 0)
 
+	// Enable topic whitelists
+	err := s.emissionsKeeper.AddToTopicWorkerWhitelist(ctx, originalTopicId, workerMsg.WorkerDataBundle.Worker)
+	require.NoError(err)
+	err = s.emissionsKeeper.AddToTopicWorkerWhitelist(ctx, newTopicId, workerMsg.WorkerDataBundle.Worker)
```
xmariachi, line 453: same as above here

x/emissions/migrations/v6/migrate.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Iterate all topics to migrate using collections.go api
+	// Turn on worker and reputer whitelists for each topic
+	for id := uint64(1); id < nextTopicId; id++ {
+		idByte := make([]byte, 8)
+		binary.BigEndian.PutUint64(idByte, id)
+		ctx.Logger().Info(fmt.Sprintf("MIGRATION V5: Updating topic:%d", id))
```
xmariachi: Isn't this message "MIGRATION v6"?

x/emissions/module/autocli.go
```diff
@@ -707,13 +707,114 @@ func (am AppModule) AutoCLIOptions() *autocliv1.ModuleOptions {
...
+					Use:       "is-whitelisted-topic-creator [address]",
+					Short:     "Check if address is whitelisted as topic creator",
+					PositionalArgs: []*autocliv1.PositionalArgDescriptor{
+						{ProtoField: "address"},
+					},
+				},
+				{
+					RpcMethod: "IsWhitelistedGlobalActor",
+					Use:       "is-whitelist-global-actor [address]",
+					Short:     "Check if address is a global whitelist actor",
```
xmariachi: "globally whitelisted" instead of "global whitelist"

xmariachi (dismissed):
Re-reviewed, lgtm. Added some suggestions

x/emissions/keeper/msgserver/msg_server_whitelist.go
```diff
@@ -16,20 +16,23 @@ func (ms msgServer) AddToWhitelistAdmin(ctx context.Context, msg *types.AddToWhi
 	if err != nil {
 		return nil, err
 	}
-	// Check that sender is also a whitelist admin
-	isAdmin, err := ms.k.IsWhitelistAdmin(ctx, msg.Sender)
+
+	// Validate the address
+	err = ms.k.ValidateStringIsBech32(msg.Address)
```
amimart, line 21: This seems to be checked twice, here and in the `Keeper.AddWhitelistAdmin`, could remove one of them.

(Saw this pattern multiple times)

x/emissions/keeper/whitelist.go
```diff
@@ -0,0 +1,321 @@
...
+	if err := types.ValidateBech32(admin); err != nil {
+		return errorsmod.Wrap(err, "error validating admin id")
+	}
+	has, err := k.whitelistAdmins.Has(ctx, admin)
+	if err != nil {
+		return err
+	} else if !has {
+		return nil
+	}
+	return k.whitelistAdmins.Remove(ctx, admin)
```
amimart:
```suggestion
	return k.whitelistAdmins.Remove(ctx, admin)
```
The call to `KeySet.Remove` is already a no-op if the key is not present, no need to check the its existence before, we can reduce state access 😸

(This is also present in other places)

amimart (dismissed):
Seems pretty good to me, I just noted some slight suggestions

x/emissions/keeper/msgserver/msg_server_worker_payload_test.go
kpeluso: yup redundant. think i was testing the test!

x/emissions/keeper/whitelist.go
kpeluso: great catch! updated everything in `whitelist.go` with this in mind

x/emissions/keeper/msgserver/msg_server_whitelist.go
kpeluso: We started this rough "defense-in-depth" practice where we purposefully do excessive data validations. Not the biggest fan of it because of the excess, however not doing this did cost us previously as we ended up storing "bad values" in the past. The theory is that functions could be reused in contexts without these checks, which could risk "poisoning" state as undesirable values are written to state.

If I were to remove the excessive validations, I would remove it from `whitelist.go`.

imo the best way to handle it is to wrap all primitive types in a domain language e.g. `type Address ...` instead of a `string` that is an address. Then when you construct the domain object, you do the validation once, and never have to again.

cc @relyt29 @xmariachi @spooktheducks @guilherme-brandao @fernandofcampos if they wanted to share thoughts

amimart: Yes I see, I completely agree the type should ensure a _safety by design_, we can use `sdk.AccAddress` in the related `Keeper` method:
```go
func (k *Keeper) AddWhitelistAdmin(ctx context.Context, admin sdk.AccAddress)
```
Or even directly through the `ActorId` type:
```go
type ActorId = sdk.AccAddress
```
The last one provides a better expressiveness

This may bring a lot of changes, maybe better to address in a dedicated PR if we chose to go that way

x/emissions/proto/emissions/v6/tx.proto
```diff
@@ -0,0 +1,547 @@
...
+    (gogoproto.customtype) = "github.com/allora-network/allora-chain/math.Dec",
+    (gogoproto.nullable) = false
+  ];
+  repeated bool global_whitelist_enabled = 50;
+  repeated bool topic_creator_whitelist_enabled = 51;
+}
+
+message UpdateParamsRequest {
+  option (cosmos.msg.v1.signer) = "sender";
+  string sender = 1;
```
amimart, line 215: Also related to types checks mentioned in other comments, adding [proto scalars](https://docs.cosmos.network/v0.50/build/building-modules/protobuf-annotations#scalar) can give a hint to clients to enforce the type, but it's only on the client side..

```suggestion
  string sender = 1 string [(cosmos_proto.scalar) = "cosmos.AddressString"];
```

guilherme-brandao: Nice suggestion! And it won't require any migration.

x/emissions/keeper/keeper.go
guilherme-brandao: Not necessarily, it would just need to use a KeyPair - (ActorId, Permission). But yeah, it's just nice to have for now to make the code leaner.

x/emissions/keeper/keeper.go
guilherme-brandao: Good point! Makes sense

guilherme-brandao (approved):
Lgtm!

x/emissions/proto/emissions/v6/tx.proto
guilherme-brandao: I think we can do this in a separate PR

xmariachi (approved):

x/emissions/proto/emissions/v6/tx.proto
kpeluso: https://linear.app/alloralabs/issue/PROTO-2953/add-proto-scalars-to-all-proto-props

kpeluso: Great points! Made it low prio for now

amimart (approved):

x/emissions/keeper/msgserver/msg_server_whitelist.go
kpeluso: https://linear.app/alloralabs/issue/PROTO-2954/change-all-uses-of-primitive-types-to-use-classes

Great points. Saved for a follow-up ticket.

PR #670, opened 2024-11-13
Adjust topic initial regret initialization

guilherme-brandao:
#### PROBLEM
- To calculate the initial topic regret, we need to have at least one participant with enough experience to be used in the `StdDev`.

- workerRegrets are selected based on the worker experience: 
`experienceCount > 1/alphaRegret`

- After the last migration, we deleted all regrets, set the initial regret of topics to 0, and also set all scores to 0. All participants were reinitialized.

- A high number of workers with equal scores means high volatility in the active set and, consequently, a low chance of experienced participants in the active set. The problem gets a lot worse for topics that have long epochs.

#### SOLUTION
- If there are fewer than 10 workers (or not enough to calculate a reliable standard deviation), we include regrets with minimal experience.
- In cases where regrets are used with insufficient experience, set the dummy regret to the 25th percentile of existing regrets, without adjusting by a standard deviation multiple.
- Define "not enough workers" as fewer than 10 workers.
- This approach activates the intended functionality once there are at least 10 workers, each with at least 
`1/alphaRegret`  epochs of experience. If these conditions aren’t met, we’re in a cold start, prioritizing rapid inclusion to avoid gaps.

The essence of the change is to maintain the purpose of regret equilibrium before calculating standard deviation, aligning with the design intention of regret handling.

https://linear.app/alloralabs/issue/PROTO-2806/investigate-potential-inference-synthesis-issues-found-by-research

- [OK] If tested, please describe how. If not, why tests are not needed.
- [OK] If documented, please describe where. If not, describe why docs are not needed.
- [OK ] Added to `Unreleased` section of `CHANGELOG.md`?


x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -394,40 +407,53 @@ func GetCalcSetNetworkRegrets(args GetCalcSetNetworkRegretsArgs) error {
...
-// offset = cnorm - 8.25 / pnorm
-// dummy_regret[i] = np.percentile(regrets[i-1, :], 25.) + offset * denominator
-// It is assumed that the regrets are filtered by experience for each actor
-// i.e. if they have not been included in the topic for enough epochs, their regret is ignored.
+// When using experienced workers' regrets:
+//
+//	denominator = std(regrets) + epsilon
+//	offset = cnorm - 8.25 / pnorm
+//	initialRegret = percentile(regrets, 25) + offset * denominator
+//
```
relyt29:
```suggestion
// for cold starts
```

x/emissions/keeper/inference_synthesis/network_regrets_test.go
```diff
@@ -860,93 +862,170 @@ func (s *InferenceSynthesisTestSuite) TestUpdateTopicInitialRegret() {
...
+// the initial regret should be the 25th percentile of all workers' regrets
+// without the offset calculation
+// 3. The new initial regret should match the expected value based on the fixed regrets
 //
-// This test ensures that the initial regret of a topic is not modified when
-// there are no experienced workers, maintaining the system's integrity by
-// preventing premature updates to the topic's regret baseline.
-func (s *InferenceSynthesisTestSuite) TestNotUpdateTopicInitialRegret() {
+// This test ensures that the fallback regret calculation works correctly when there
+// are no experienced workers but enough total workers to use the fallback mechanism.
```
relyt29, line 930: do we have a corresponding test for when the system is not at a cold start and there are >10 experienced workers?

relyt29 (dismissed):
Conceptually looks good. Minor comments

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -377,15 +380,25 @@ func GetCalcSetNetworkRegrets(args GetCalcSetNetworkRegretsArgs) error {
...
 	}
 
+	// Get the initial topic std dev regrets
+	// If we don't have enough workers with enough experience, use the fallback regrets
+	initialTopicStdDevRegrets := make([]alloraMath.Dec, 0)
+	if len(workersRegrets) >= 10 {
+		initialTopicStdDevRegrets = workersRegrets
+	} else if len(fallbackRegrets) > 0 {
+		initialTopicStdDevRegrets = fallbackRegrets
+	}
```
amimart: This could be simplified
```suggestion
	var initialTopicStdDevRegrets []alloraMath.Dec
	if len(workersRegrets) >= 10 {
		initialTopicStdDevRegrets = workersRegrets
	} else {
		initialTopicStdDevRegrets = fallbackRegrets
	}
```

amimart (commented):
Nice, just noted some simplification

x/emissions/keeper/inference_synthesis/network_regrets_test.go
guilherme-brandao: Yeah, in the same file, we have a couple of tests that cover the non-cold start part.

relyt29: what are those tests, I was struggling to figure them out

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -394,40 +406,53 @@ func GetCalcSetNetworkRegrets(args GetCalcSetNetworkRegretsArgs) error {
...
-	// Calculate the Denominator
+	quantile, err := alloraMath.GetQuantileOfDecs(regrets, quantileRegret)
+	if err != nil {
+		return alloraMath.ZeroDec(), err
+	}
+
+	// If using fallback regrets, just return the quantile
+	if usingFallbackRegrets {
+		return quantile, nil
+	}
```
kpeluso: Don't like the idea of splitting up control flow logic between scopes. Would be cleaner if you were to
1. Decide on which scope to do the control flow. In this case, I say the high-level scope (which calls `CalcTopicInitialRegret()`)
2. Pass in just `quantile` instead of `quantileRegret` and `usingFallbackRegrets`. This has impacts in L393 as of this commit.

kpeluso (changes requested):

x/emissions/keeper/inference_synthesis/network_regrets_test.go
guilherme-brandao: `TestUpdateTopicInitialRegret`

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -377,9 +380,12 @@ func GetCalcSetNetworkRegrets(args GetCalcSetNetworkRegretsArgs) error {
...
 
-	// Recalculate topic initial regret
-	if len(workersRegrets) > 0 {
-		updatedTopicInitialRegret, err := CalcTopicInitialRegret(
+	// Check if we have enough workers with enough experience to calculate the topic initial regret
+	// If we don't have enough workers with enough experience, use the fallback regrets
+	var updatedTopicInitialRegret alloraMath.Dec
+	var err error
+	if len(workersRegrets) >= 10 {
+		updatedTopicInitialRegret, err = CalcTopicInitialRegret(
```
kpeluso: I think you need to pass in the regret at the `InitialRegretQuantile` not `InitialRegretQuantile` itself.

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -377,9 +380,12 @@ func GetCalcSetNetworkRegrets(args GetCalcSetNetworkRegretsArgs) error {
 		}
 	}
 
-	// Recalculate topic initial regret
-	if len(workersRegrets) > 0 {
-		updatedTopicInitialRegret, err := CalcTopicInitialRegret(
+	// Check if we have enough workers with enough experience to calculate the topic initial regret
```
kpeluso: You changed up the logic a lot relative to what I last reviewed, is this really right this time? You were doing something if some array had >0 length at some point, I don't see that happening anymore...

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -377,9 +380,12 @@ func GetCalcSetNetworkRegrets(args GetCalcSetNetworkRegretsArgs) error {
...
+	// If we don't have enough workers with enough experience, use the fallback regrets
+	var updatedTopicInitialRegret alloraMath.Dec
+	var err error
+	if len(workersRegrets) >= 10 {
```
kpeluso: Must we always be limited to those regrets passed in this epoch?

Wondering if it makes more sense to keep a running, recent history of regrets like we do for scores @jmdkastro ?

For reference, see how `MaxSamplesToScaleScores` is used to see how we maintain a running, recent history of numbers.

kpeluso (changes requested):

x/emissions/keeper/inference_synthesis/network_regrets.go
jmdkastro: Regrets are already an EMA, so there it isn't necessary. Scores are instantaneous, so there it makes more sense to have some running history.

kpeluso:
The PR description should be updated to reflect the latest intentions implemented by this PR / description of what's actually here @guilherme-brandao

kpeluso:
@guilherme-brandao the PR template was not fully filled out

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -377,57 +380,77 @@ func GetCalcSetNetworkRegrets(args GetCalcSetNetworkRegretsArgs) error {
...
-			args.PNorm,
-			args.CNorm,
-			args.InitialRegretQuantile,
-			args.PNormSafeDiv,
-		)
+	// Select which regrets to use for calculating the topic initial regret:
+	// - If we have 10 or more experienced workers, use their regrets
+	// - Otherwise, use fallback regrets which include all workers
+	regrets := fallbackRegrets
+	if len(workersRegrets) >= 10 {
```
kpeluso: should be a global parameter

x/emissions/keeper/inference_synthesis/network_regrets.go
```diff
@@ -377,57 +380,77 @@ func GetCalcSetNetworkRegrets(args GetCalcSetNetworkRegretsArgs) error {
...
+	// Only proceed if we have any regrets to process
+	if len(regrets) > 0 {
+		quantile, err := alloraMath.GetQuantileOfDecs(regrets, args.InitialRegretQuantile)
 		if err != nil {
 			return errorsmod.Wrapf(err, "Error calculating topic initial regret")
 		}
+
+		// Set initial value to the quantile (used when we don't have enough experienced workers)
+		updatedTopicInitialRegret := quantile
+		if len(workersRegrets) >= 10 {
```
kpeluso: here is that same should-be-global parameter again

kpeluso (changes requested):
Just 1 missing param. Otherwise, matches [this description](https://linear.app/alloralabs/issue/PROTO-2806/investigate-potential-inference-synthesis-issues-found-by-research#comment-35cc1352)

x/emissions/types/params.go
```diff
@@ -57,6 +57,7 @@ func DefaultParams() Params {
 		PNormSafeDiv:                        alloraMath.MustNewDecFromString("8.25"),      // pnorm divide value to calculate offset with cnorm
 		GlobalWhitelistEnabled:              true,                                         // global whitelist enabled => all global whitelisted actors can create topics and participate in all topics as workers and reputers
 		TopicCreatorWhitelistEnabled:        true,                                         // topic creator whitelist enabled => all topic creator whitelisted actors can create topics
+		MinExperiencedWorkerRegrets:         int64(10),                                    // minimum number of experienced workers required to use their regrets for calculating the topic initial regret
```
kpeluso: why would this ever be negative / an int??

kpeluso (approved):

amimart (approved):

PR #673, opened 2024-11-15
Fix bug in single reputer score calculation

guilherme-brandao:
#### PROBLEM
A bug was found in the calculation of the scores for the reputers, which occurs when there is only one Reputer participating in the epoch, and this reputer had a listening coefficient of 0 in the previous epoch.

#### SOLUTION
When this scenario occurs, we should set the listening coefficient to 0.5. This prevents division by 0 errors during the score calculation.

https://linear.app/alloralabs/issue/PROTO-2837/generate-rewards-stakefraction-division-undefined


CHANGELOG.md
```diff
@@ -63,7 +63,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [#634](https://github.com/allora-network/allora-chain/pull/634) Add events for listening coefficients and the computationally cheapest regrets
 
 ### Changed
-
+* [#673](https://github.com/allora-network/allora-chain/pull/673) Fix bug in single reputer score calculation
```
relyt29: nitpick but I would think this would go under "fixed" not "changed" category

x/emissions/module/rewards/rewards_internal.go
```diff
@@ -445,6 +445,13 @@ func GetAllReputersOutput(
 	coefficients := make([]alloraMath.Dec, len(initialCoefficients))
 	copy(coefficients, initialCoefficients)
 
+	// Check if there's only one reputer and the listening coefficient is zero
+	// If so, set the listening coefficient to 0.5
+	// This is to ensure there's no bugs in the scores calculation
+	if len(coefficients) == 1 && coefficients[0].IsZero() {
+		coefficients[0] = alloraMath.MustNewDecFromString("0.5")
```
relyt29: mrrr hardcoded constants mrrrr

do we not have a global param for this

amimart: AFAIU the idea is to force the single reputer listening coefficient to the nearest positive value of zero. 
Instead of `0.5` we could use the decimal128 min, which having 34 precision digits would be `alloraMath.MustNewDecFromString("0.000000000000000000000000000000001")` I guess.

amimart (commented):
Just noted a suggestion I'll let you judge, I also agree with @relyt29 remarks

x/emissions/module/rewards/rewards_internal.go
guilherme-brandao: According to Diederik, the other alternative would be to find the value closest to zero from the previous epoch or 0.5. I chose the latter option because of the simplicity, the other option would be necessary to keep track of all the previous epoch active reputers.

amimart: Oh yes makes sense

xmariachi: This covers in the 1-reputer case, but is it possible that we have all coefficients = 0 in a multi-reputer scenario?

guilherme-brandao: Good point, I will double check with diederik if we can fallback all the reputers to 0.5 in this specific case that you mentioned.

xmariachi (dismissed):
lgtm

amimart (approved):

x/emissions/module/rewards/rewards_internal.go
```diff
@@ -435,23 +435,27 @@ func GetAllReputersOutput(
...
 	copy(coefficients, initialCoefficients)
 
+	// Check if all coefficients are zero
+	// If so, set all coefficients to the fallback listening coefficient
+	if allCoefficientsZero {
+		for i := range coefficients {
+			coefficients[i] = params.FallbackListeningCoefficient
+		}
+	}
+
```
kpeluso: You're splitting up control flow between scopes here and the scope that calls `GetAllReputersOutput()`.

In the calling scope, you check if `allCoefficientsZero`, yet in here you recheck that boolean and actually apply that condition.

What you should instead do is just pass in a conditionally-altered `initialCoefficients`, one that uses fallbacks or not depending on the control flow. This makes the code way more readable.

kpeluso: This also affects L79 as of this current commits.

kpeluso (changes requested):

amimart (changes requested):
I think the `FallbackListeningCoefficient` new parameter should be set at its default value in the migration to 'v6'

amimart (approved):
LGTM 👍

xmariachi (approved):
lgtm

kpeluso (approved):

PR #687, opened 2024-11-27
Fix submission epoch boundaries

xmariachi:
* worker window: boundary block is exclusive of end block. Added comments where relevant.

* reputer window:
  * the upper boundary should be `gt_lag + (gt_lag%epochLength) + epochLength`, instead of `gt_lag*2`. The previous value would be inconsistent, especially in cases where `gt_lag is >> epochLength`.
  * added comments where relevant

- [X] If tested, please describe how. If not, why tests are not needed. -- covered by unit tests and integration tests, also tested locally.
- [x] If documented, please describe where. If not, describe why docs are not needed. -- windows are documented in README.


kpeluso (commented):
Correct, great catch

kpeluso (approved):

amimart (approved):
LGTM

PR #688, opened 2024-11-28
Add the evidence module

amimart:

Add the evidence module in order to manage `DuplicateVoteEvidence` evidences provided by CometBFT and effectively slash concerned validators.

I've took the opportunity to add the `StoreUpgrades` notion to `Upgrade` to make reusable the logic attach to the addition/modification of stores.

Module wiring seems self-documented through the `app.yaml`, tested locally.

https://linear.app/alloralabs/issue/PROTO-2964/add-the-evidence-module

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * [612](https://github.com/allora-network/allora-chain/pull/612) Load testnet state into local fork via in-place-testnet command 
+* [688](https://github.com/allora-network/allora-chain/pull/688) Add the `evidence` module.
 
 ### Security
 
--- a/app/app.go
+++ b/app/app.go
@@ -46,21 +46,22 @@ import (
 	ibckeeper "github.com/cosmos/ibc-go/v8/modules/core/keeper"
 	metrics "github.com/hashicorp/go-metrics"
 
-	_ "cosmossdk.io/api/cosmos/tx/config/v1" // import for side-effects
-	_ "cosmossdk.io/x/circuit"               // import for side-effects
-	_ "cosmossdk.io/x/feegrant/module"       // import for side-effects
-	_ "cosmossdk.io/x/upgrade"
-	_ "github.com/allora-network/allora-chain/x/emissions/module"
-	_ "github.com/allora-network/allora-chain/x/mint/module" // import for side-effects
-	_ "github.com/cosmos/cosmos-sdk/x/auth"                  // import for side-effects
-	_ "github.com/cosmos/cosmos-sdk/x/auth/tx/config"        // import for side-effects
-	_ "github.com/cosmos/cosmos-sdk/x/authz/module"          // import for side-effects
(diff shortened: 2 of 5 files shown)


fernandofcampos (dismissed):

xmariachi (dismissed):
Approving, however a question: 

Per [docs](https://docs.cosmos.network/main/build/modules/evidence#evidence-handling): 
> If valid Equivocation evidence is included in a block, the validator's stake is reduced (slashed) by SlashFractionDoubleSign as defined by the x/slashing module of what their stake was when the infraction occurred, rather than when the evidence was discovered. We want to "follow the stake", i.e., the stake that contributed to the infraction should be slashed, even if it has since been redelegated or started unbonding.
> In addition, the validator is permanently jailed and tombstoned to make it impossible for that validator to ever re-enter the validator set.

Now that we set `evidence` module, do we expect this to happen now?
In that case, should we configure the `SlashFractionDoubleSign` and/or other parts of the `slashing` module?

amimart:
> Now that we set `evidence` module, do we expect this to happen now? In that case, should we configure the `SlashFractionDoubleSign` and/or other parts of the `slashing` module?

Exactly yes that's what will happen now, concerning the parameters these are already set in testnet, [here](https://github.com/allora-network/networks/blob/e08ad4d3dc35daed389d4ee830d074e38ff29b95/allora-testnet-1/genesis.json#L570) for slashing and [here](https://github.com/allora-network/networks/blob/e08ad4d3dc35daed389d4ee830d074e38ff29b95/allora-testnet-1/genesis.json#L614) for the evidence.

guilherme-brandao (dismissed):

amimart:
Will merge it after #627, conflicts should be easier to manage on this side

app/upgrades/v0_3_0/upgrades.go
```diff
@@ -15,6 +16,7 @@ const (
 var Upgrade = upgrades.Upgrade{
 	UpgradeName:          UpgradeName,
 	CreateUpgradeHandler: CreateUpgradeHandler,
+	StoreUpgrades:        storetypes.StoreUpgrades{Added: nil, Renamed: nil, Deleted: nil},
```
kpeluso: not sure if I like editing previous migrations vs just adding "ignore lint" comment (here and above and in other old files of migrations)

kpeluso (commented):
Is there no keeper to be configured here, as in https://github.com/allora-network/allora-chain/pull/627 ? If not, lgtm and I'll approve. [This suggests](https://docs.cosmos.network/main/build/modules/evidence#state) that there is some state to store, though maybe this is native to all cosmos chains given the prefix of `0x00`

amimart:
> Is there no keeper to be configured here, as in #627 ? If not, lgtm and I'll approve. [This suggests](https://docs.cosmos.network/main/build/modules/evidence#state) that there is some state to store, though maybe this is native to all cosmos chains given the prefix of `0x00`

There is a Keeper which is automatically instantiated, but you're right I'll inject it in the app that's a good practice

kpeluso (approved):

guilherme-brandao (approved):

PR #690, opened 2024-12-03
[PROTO-2787] Make investor token unlock amounts strictly monotonically increasing

relyt29:

A bug was discovered during conversation with @jmdkastro around tokenomics. We take a module parameter called `blocksPerMonth` to make the chain aware of the expected number of blocks per month. The actual blocks for every month varies from month to month and also due to network conditions. On testnet it has been not very stable. The expectation is that maintainers of the network would periodically update the blocksPerMonth parameter to keep it up to date, however, depending on the change of blocksPerMonth, that could actually compute different values for the tokenomics, potentially giving away too many tokens, leading to a higher than expected inflation rate.

The inflation rate is controlled by the target emission rate (e_i in the whitepaper). The target emission rate is inversely dependent upon the circulating supply of the token. The circulating supply of the token is dependent upon the investor token unlock schedule, which is time dependent - we expect a 3 year vesting with 1 year cliff unlock cycle (no tokens unlock until the first year). This unlock logic is dependent on the `blocksPerMonth` and if the `blocksPerMonth` changes in strange ways, it could lead to a too-high inflation rate.

For example, lets say on month 1, we set the blocksPerMonth to 5 blocks per month, but in real life, the blocks are being produced 1 block per second. That means that every second, the chain thinks that 1/5 of a month has occurred. So after 60 seconds (5 seconds for 5 blocks per month * 12 months), the token unlock schedule will start and the circulating supply will start to think that the circulating supply has gone up by the token investor unlock amount. 

Later, someone comes along and sets the blocksPerMonth to a huge value say 600000000. Because the logic for the vesting supply has no memory of previous token unlock events, the chain now thinks that the circulating supply is lower than it thought before. Accordingly, it adjusts the inflation rate to be higher, resulting in a too-high inflation rate. The correct thing to do instead is keep some memory of previous circulating supply changes so that you always err on the side of a too-low inflation rate, rather than too high.

@jmdkastro came up with an elegant "integral math" solution to this where you basically, every block, take some delta change in the vested token amount and accumulate it over time until the full unlock has been unlocked. 

That is totally functional but when I went to go implement it I observed that the actual on-chain tokenomics as currently implemented are stepwise anyways, since the token unlocks are written to happen on the first of a month when the unlock is supposed to happen, AND we only recalculate the target emission rate on a monthly basis anyways (for performance reasons). So no matter if the "integral math" was implemented it would still be a stepwise function ANYWAYS. 

So given that, I implemented a new keeper variable called "monthsUnlocked" which is meant to track the number of months that have passed since genesis. This is a value that is coerced to be between 0 and 36, and is stored in a strictly ever increasing value. This way, if an unlock of tokens ever happens, that unlock is not revertable in the logic of the chain. If the chain thinks that 200 blocks per second are being produced and that its time to unlock a bunch of tokens, and then later the chain thinks that only 1 block is being produced per second, when the chain goes to do the logic for token unlocks, it takes the stored value from the past, rather than the calculated value at the moment of what it thinks the token unlock should be. This is basically the same thing as the integral math solution but stepwise on a months cadence to match our existing on-chain logic, which makes for a smaller diff and faster integration on chain. 

Also while doing this I had to update the query service to return this additional new piece of data, and while I was doing that, the protobuf linter complained about the query service being called "QueryServer" instead of "QueryServiceServer" so theres a bit of boilerplate around that as well in this PR

PROTO-2787

We have an existing test suite for this in emissions_test.go, I have tweaked it to make it appropriately test the new code. I have not made any changes to the documentation as this is kind of a minor change.


fernandofcampos (dismissed):

x/mint/keeper/emissions.go
```diff
@@ -17,39 +17,48 @@ func GetLockedVestingTokens(
 	blocksPerMonth uint64,
 	blockHeight math.Int,
 	params types.Params,
-) (total, preseedInvestors, investors, team math.Int) {
+	monthsUnlocked math.LegacyDec,
+) (total, preseedInvestors, investors, team math.Int, updatedMonthsUnlocked math.Int) {
 	// foundation is unlocked from genesis
 	// participants are unlocked from genesis
 	// investors and team tokens are locked on a 1 year cliff three year vesting schedule
 	blocksInAYear := math.NewIntFromUint64(blocksPerMonth * 12)
```
xmariachi: Here, `blocksInAYear` would vary depending on the current value of `blocksPerMonth`. In a scenario where this goes down (or up) dramatically, it not only affects current month calculations but also the value for the whole year. 
Is this expected and covered?
E.g. it could mean that block `b` is `LT(blocksInAYear)` but block `b+1` is not.
Current setup is "estimating the past", isn't it?

Suggestion: we can keep track the block at which the monthUnlocks happened, so the yearly estimation does not extend "over the past", ie, we can keep track of what already happened, and only estimate what comes next.
A year can be thus computed as "last 11 months time + `blocksPerMonth`")

xmariachi (commented):
A comment on the solution ab "estimating the past"

x/mint/keeper/emissions.go
relyt29: `blocksInAYear` is only used in the if/else statement a couple lines below.

If block `b+1` is not less than blocksInAYear, but a previous block `b` went into the else branch because it was greater than blocksPerYear, the `updatedBlocksPerMonth` in the past would have been previously stored in the keeper later in `RecalculateTargetEmission` and so at the time of evaluation of block `b+1` in if branch `monthsUnlocked.LT(twelve)` would already be greater than 12.


We could track the specific block height but I think it makes more sense to just track the stored months already unlocked (monthsUnlocked)

relyt29: I think maybe the variable could be renamed to be better 

Edit: renamed to "monthsAlreadyUnlocked"

xmariachi: if the `blocksPerMonth` value decreases a lot, the `else` could be reached before `monthsAlreadyUnlocked` get to 12 (because it would not satisfy first condition). Is this ok? 
Wondering whether we should remove the yearly check `blockHeight.LT(blocksInAYear)` and just rely on monthly unlocks.

Another note:
since `monthsUnlocked` is based on a division over `blocksPerMonth`, which is variable, a decrease in this value could make `monthsUnlocked` to reach 36 faster which could make `monthsLocked := thirtySix.Sub(monthsAlreadyUnlocked)` to be 0 => in that case, will that mean all remaining tokens will be unlocked, or that no tokens will be _ever_ (because it is monotonic) unlocked at all?

xmariachi: Hmmm reviewing this, could it be that this `blockHeight.LT(blocksInAYear)` condition is just a sort of "pre-estimation" of monthsAlreadyUnlocked?  ie `monthsAlreadyUnlocked` at that point marking the "past" ("did we already reach 12 in the past?"), while this condition marking the present ("are we reaching it now under current yearly estimation numbers?").

In that case the mechanism looks ok - leaving the last note ab how a sudden change affects unlocks.

relyt29: yes, the code can be simplified, the if else statement is not necessary, since the math will work either way (the unlocked tokens will be zero before 12 months in the math). making changes to PR now

> since monthsUnlocked is based on a division over blocksPerMonth, which is variable, a decrease in this value could make monthsUnlocked to reach 36 faster 

Correct. In my opinion the right thing to do here is to always err on the side of saying the circulating supply is larger than it actually is in real life - this means that we err on the side of having too little inflation, rather than (potentially) too much inflation

This code doesn't actually "unlock" or send any tokens to these investors, it only is tracking how much they're supposed to have based on the unlock schedule. Our off-chain custodian partner responsible for the investor allocations will be the one legally responsible for introducing these tokens into the token supply by sending them to people who could use, transfer, or sell them

xmariachi (approved):
lgtm

x/mint/keeper/emissions.go
```diff
@@ -17,39 +17,46 @@ func GetLockedVestingTokens(
...
+	// before 1 year, all tokens are locked
+	// between 1 and 3 years, investors and team tokens are vesting and partially unlocked
+	// after 3 years, all tokens are unlocked
 	maxSupply := params.MaxSupply.ToLegacyDec()
 	percentInvestors := params.InvestorsPercentOfTotalSupply
 	percentPreseedInvestors := params.InvestorsPreseedPercentOfTotalSupply
 	percentTeam := params.TeamPercentOfTotalSupply
 	fullInvestors := percentInvestors.Mul(maxSupply).TruncateInt()
 	fullPreseedInvestors := percentPreseedInvestors.Mul(maxSupply).TruncateInt()
 	fullTeam := percentTeam.Mul(maxSupply).TruncateInt()
```
guilherme-brandao, line 34: Maybe it's already too late for this, but probably all these parameters should have been created with the same type. Some of them are `Dec`, while others are `Int`, so we need to keep converting them to be able to make various ops.

guilherme-brandao (approved):
lgtm! just a minor (optional) suggestion

x/mint/keeper/emissions.go
relyt29: yeah it's a mess but I think it's because the Int division is less precise that we had to use dec sometimes

x/mint/keeper/emissions.go
```diff
@@ -74,30 +81,33 @@ func GetCirculatingSupply(
...
-		return math.Int{}, math.Int{}, math.Int{}, math.Int{}, err
+		return math.Int{}, math.Int{}, math.Int{}, math.Int{}, math.Int{}, err
 	}
 	totalSupply = params.MaxSupply
-	lockedVestingTokens, _, _, _ = GetLockedVestingTokens(
+	lockedVestingTokens, _, _, _, updatedMonthsUnlocked = GetLockedVestingTokens(
 		blocksPerMonth,
 		math.NewIntFromUint64(blockHeight),
 		params,
+		monthsAlreadyUnlocked.ToLegacyDec(),
```
kpeluso, line 102: the concepts of `monthsAlreadyUnlocked` and `updatedMonthsUnlocked` seems so similar, yet are manipulated in 2 different scopes. for this, I'm not going to block, as there is some overlapping control flow (to calc updatedMonthsUnlocked and the other values, obvi) but I think it'd be cleaner to separate this logic from the compute of vested tokens

kpeluso (approved):

PR #691, opened 2024-12-04
Fix number of inferers considered in the reward cycle

fernandofcampos:

`k.lowestInfererScoreEma` was not being updated on a new Epoch. Then, after the 32 slots get filled up, and when the 33rd inference arrives, it was checked against the value in `k.lowestInfererScoreEma` from previous Epoch.

If the new inferer has a greater score, it gets added and the old one gets removed. But since the old one was from the previous epoch (that is, the worker in the stored `Score` struct of lowest score), it does not exist anymore, and the function that removes the inferer won't return and error if it does not exist. So we don't remove any inferer, and we add a new one. That's how we end up with 33.

This PR effectively resets that stored lowest score (stored by topic) upon the start of a new worker submission window. Therefore, whenever one attempts to move a worker out of the active set, we can be assured the worker is actually a part of the topic's current active set.

https://linear.app/alloralabs/issue/PROTO-2851/fix-number-of-inferers-considered-in-the-reward-cycle

   - Unit tests provided
   - Evident in the code and its comments
   - Added to stanza of next release

## Still Left Todo

### DIFF
--- a/x/emissions/keeper/keeper.go
+++ b/x/emissions/keeper/keeper.go
@@ -1112,19 +1112,13 @@ func (k *Keeper) AppendInference(
 		return errors.New("inference already submitted")
 	}
 
+	// Get active inferers for topic
 	workerAddresses, err := k.GetActiveInferersForTopic(ctx, topic.Id)
 	if err != nil {
 		return errorsmod.Wrap(err, "error getting active inferers for topic")
 	}
-	// If there are less than maxTopInferersToReward, add the current inferer
-	if uint64(len(workerAddresses)) < maxTopInferersToReward {
-		err := k.AddActiveInferer(ctx, topic.Id, inference.Inferer)
-		if err != nil {
-			return errorsmod.Wrap(err, "error adding active inferer")
-		}
-		return k.InsertInference(ctx, topic.Id, *inference)
-	}
 
+	// Get previous EMA score for the current inferer
 	previousEmaScore, err := k.GetInfererScoreEma(ctx, topic.Id, inference.Inferer)
 	if err != nil {
 		return errorsmod.Wrapf(err, "Error getting inferer score ema")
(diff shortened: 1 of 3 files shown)


kpeluso (dismissed):

xmariachi (dismissed):
lgtm

guilherme-brandao (changes requested):
updated comment below

x/emissions/keeper/keeper.go
```diff
@@ -1112,19 +1112,13 @@ func (k *Keeper) AppendInference(
 		return errors.New("inference already submitted")
 	}
 
+	// Get active inferers for topic
 	workerAddresses, err := k.GetActiveInferersForTopic(ctx, topic.Id)
 	if err != nil {
 		return errorsmod.Wrap(err, "error getting active inferers for topic")
 	}
```
guilherme-brandao: To avoid potential unnecessary queries, I'd move this below `GetLowestInfererScoreEma` since it's just used after getting the lowest score.

x/emissions/keeper/keeper.go
```diff
@@ -1160,6 +1172,16 @@ func (k *Keeper) AppendInference(
...
+
+		// Check if the inferer with lowest score is active before removing it, because remove will not fail if the inferer is not active
+		isActive, err := k.IsActiveInferer(ctx, topic.Id, lowestEmaScore.Address)
+		if err != nil {
+			return errorsmod.Wrap(err, "error checking if inferer is active")
+		}
+		if !isActive {
+			return errors.New("inferer with lowest score is not active")
+		}
+
```
guilherme-brandao, line 1176: I'm wondering if it should be handled this way. If we are trying to remove an active inferer that doesn't exist, it's a bug on our end. Maybe it's just a sanity check to avoid disrupting the reward flow, but if this happens, we should be made aware somehow. Lmk your thoughts.

guilherme-brandao (changes requested):
Nice job, I'm almost sure we need to also apply this to forecasters and reputers as well, since they follow the same logic. Can you double-check?

x/emissions/keeper/keeper.go
```diff
@@ -1134,21 +1128,39 @@ func (k *Keeper) AppendInference(
...
-		}
+	}
+	// If there are no lowest inferer score ema, it means it is the first inference for the topic
+	if !found {
+		lowestEmaScore = previousEmaScore
 		err = k.SetLowestInfererScoreEma(ctx, topic.Id, lowestEmaScore)
 		if err != nil {
 			return errorsmod.Wrap(err, "error setting lowest inferer score ema")
 		}
 	}
```
guilherme-brandao: I think this can be removed; this condition is already checked in the code below. 

It will avoid double persistence.

guilherme-brandao: Also please double-check if we still need `UpdateLowestScoreFromInfererAddresses` in this new implementation, I can see some cases that we persist the same lowest score 3 times.

guilherme-brandao (changes requested):

x/emissions/keeper/keeper.go
kpeluso: this "not found" check applies for the very first time the topic begins an epoch

kpeluso: We still do need `UpdateLowestScoreFromInfererAddresses` I believe because if we remove an actor, it may be any actor => we don't know the new lowest actor

x/emissions/keeper/keeper.go
kpeluso: This is an extra check indeed. "Defense in depth" applies here imo. Could def argue it's excessive, or warranted. imo it's good to have, much like our other excessive checks

x/emissions/keeper/keeper.go
guilherme-brandao: > this "not found" check applies for the very first time the topic begins an epoch

the condition below is already covering this case: `if uint64(len(workerAddresses)) == 0 `

fernandofcampos: Indeed, not found is not necessary, if it's just used for the very first time.

relyt29:
agree with Guilherme's comments, directionally looks good though, good work nailing this down

happy to approve whenever changelog conflict and guilherme's comments are addressed

kpeluso:
^^i resolved conflict in changelog, not "merge dev into this"

guilherme-brandao (dismissed):
lgtm!

kpeluso (dismissed):

x/emissions/keeper/keeper.go
fernandofcampos: Done

kpeluso (approved):

amimart (approved):
Well done! LGTM

PR #694, opened 2024-12-04
Make fuzzer whitelist aware

amimart:

Allow the fuzzer to consider the different whitelist states to be able to operate appropriately, this means that the different whitelists are tracked in the simulation data.

This also brings new transitions related to the management of whitelists:
- `AddToAdminWhitelist`
- `RemoveFromAdminWhitelist`
- `AddToGlobalWhitelist`
- `RemoveFromGlobalWhitelist`
- `AddToTopicCreatorWhitelist`
- `RemoveFromTopicCreatorWhitelist`
- `EnableTopicWorkerWhitelist`
- `DisableTopicWorkerWhitelist`
- `EnableTopicReputerWhitelist`
- `DisableTopicReputerWhitelist`
- `AddToTopicWorkerWhitelist`
- `RemoveFromTopicWorkerWhitelist`
- `AddToTopicReputerWhitelist`
- `RemoveFromTopicReputerWhitelist`

As only the whitelists managed by the fuzzer are kept in simulation data, in order to provide coherent transitions we now limit the topics we interact with to the ones created by the fuzzer.

_The initial setup as also been updated to add some actors in the different whitelists, not sure the choices I made are relevant though._
**EDIT**: The configuration file now takes inputs to specify the number of actors to put in the general whitelists, and all the transitions related to whitelists are performed at least once.

I've tested the fuzzer locally in its `behave` mode.

https://linear.app/alloralabs/issue/PROTO-2931/make-fuzzer-whitelist-aware


test/fuzz/.config.json.example
```diff
@@ -1,19 +1,33 @@
...
+		"CollectDelegatorRewards": 6,
+		"DoInferenceAndReputation": 20,
 		"UnregisterWorker": 4,
 		"UnregisterReputer": 4,
 		"UnstakeAsReputer": 6,
 		"UndelegateStake": 6,
 		"CancelStakeRemoval": 0,
-		"CancelDelegateStakeRemoval": 0
+		"CancelDelegateStakeRemoval": 0,
+		"AddToAdminWhitelist": 0,
```
relyt29: I think these would be better as 1 not zero

test/fuzz/actors_setup_test.go
```diff
@@ -509,23 +616,38 @@ func simulateAutomaticInitialState(
...
 	iterationCount++
-	// pick 4 reputers, 4 workers, and 2 delegators
-	startReputers, startWorkers, startDelegators := pickAutoSetupActors(m, data)
+
+	listTopics := data.getTopics()
+	require.Len(m.T, listTopics, 2)
+
+	// put 2 reputers & 2 workers in one topic whitelists
+	iterationCount = startAddToTopicWorkerWhitelist(m, data, faucet, startWorkers[2:], listTopics[1], iterationCount)
+	iterationCount = startAddToTopicReputerWhitelist(m, data, faucet, startReputers[2:], listTopics[1], iterationCount)
```
relyt29: Are we running through every possible state transition at least once? If not can you change this code so that it runs through every state transition that you have added, including the `removal` state transitions?

so e.g. instead of putting 2 reputers and 2 workers, you would put 3 in and remove 1 before the simulation fully started

relyt29 (changes requested):

kpeluso (commented):
This is very good already. However, also thinking that it could make sense to have a mode where the whitelist is essentially turned off and/or all workers added, and then the fuzzer proceeds as it was, so it fuzzes within the set of states where workers are allowed to do whatever.

This is value too obvi -- fuzzing through everything including whitelist-related state transitions

amimart:
> This is very good already. However, also thinking that it could make sense to have a mode where the whitelist is essentially turned off and/or all workers added, and then the fuzzer proceeds as it was, so it fuzzes within the set of states where workers are allowed to do whatever.

That makes sense, I'd like to propose instead of a mode some new configuration parameters to specify the initial setup related to whitelists:
- `initialActorsInAdminWhitelist: uint`: The number of actors added to the admin whitelist at initial setup;
- `initialActorsInGlobalWhitelist: uint`: The number of actors added to the global whitelist at initial setup;
- `initialActorsInTopicCreatorWhitelist: uint`: The number of actors added to the topic creator whitelist at initial setup;

By setting these parameters equals to the number of actors & setting 0 weight on whitelist transitions we'd have this behaviour, while giving a better configuration granularity.

What do you think?

test/fuzz/actors_setup_test.go
amimart: Oh ok I'll make sure we run through every transition at least once, this is clearly not the case right now

relyt29:
> I'd like to propose instead of a mode some new configuration parameters

I am fine with that and think it makes sense, just put it in the config.json with some sane default configuration and we're good to go

relyt29 (dismissed):
great work on this :)

fernandofcampos (dismissed):

xmariachi (approved):
lgtm

kpeluso (approved):
very clean!

PR #695, opened 2024-12-04
Mint module GenesisState, proto, and v3 no-op migration

relyt29:

As we're going through the v0.7.0 upgrade checklist I observed that PR #690 was missing `GenesisState` updates for the new keeper field. And since the errors are updated, AND since there's a new keeper field it probably makes sense to bump the `ConsensusVersion` of the mint module. I also added a migration for the mint module which is a no-op, just a print line that tells you that the mint module was upgraded!

PROTO-2787

These tests will have to be tested by hand as part of the v0.7.0 checklist upgrade flow.


amimart (dismissed):
Good catch 😉

x/mint/types/genesis.go
```diff
@@ -1,19 +1,23 @@
 package types
 
-import "cosmossdk.io/math"
+import (
```
xmariachi: single-line import instead?

x/mint/types/genesis.go
```diff
@@ -42,5 +47,14 @@ func ValidateGenesis(data GenesisState) error {
 		return ErrInvalidEcosystemTokensMinted
 	}
 
+	if data.MonthsUnlocked.IsNegative() {
+		return ErrInvalidMonthsUnlocked
+	}
+
+	thirtySix := math.NewInt(36)
+	if data.MonthsUnlocked.GT(thirtySix) {
```
xmariachi, line 53: MonthsUnlocked is defined as an int. Validation could also check < 0 values

relyt29: literally two lines above it checks `IsNegative`

x/mint/types/genesis.go
relyt29: uhhh yeah

x/mint/module/module.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 // ConsensusVersion defines the current x/mint module consensus version.
-const ConsensusVersion = 2
+const ConsensusVersion = 3
```
kpeluso: module mismatch. is it 3 or 4?

kpeluso (changes requested):

kpeluso (approved):

amimart (approved):

PR #699, opened 2024-12-09
Adapt install script to new release assets naming

amimart:

Update the `install.sh` script according to the new release assets name template, for instance:
`allorad_${OS}_${ARCH}` => `allora-chain_${VERSION_DIGITS}_${OS}_${ARCH}`

The installed binary is still `allorad`.

I took the opportunity to update the default version installed to the current latest (i.e. `0.7.0`).

Can be tested before merging with:
```bash
curl -sSL https://raw.githubusercontent.com/allora-network/allora-chain/arnaud%2Fadapt-install-script/install.sh | bash -s -- v0.7.0
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+* [#699](https://github.com/allora-network/allora-chain/pull/699) Make install script manage new release assets naming
+
 ## v0.7.0
 
 ### Added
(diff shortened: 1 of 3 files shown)


relyt29 (approved):
this is fine although I think it makes sense to start storing this kind of stuff in an .env or something in the repo because between the docker files and random install scripts and everything else it's easier to just have one env file that has the latest version

or even better rip this out from the actual version / consensusVersion in the golang so we never have to update anything at all

xmariachi (approved):

fernandofcampos (approved):

amimart:
> this is fine although I think it makes sense to start storing this kind of stuff in an .env or something in the repo

This script being used as an install oneliner through `curl -sSL https://raw.githubusercontent.com/allora-network/allora-chain/refs/heads/main/install.sh | bash` storing the latest version in a `.env` or resolving dynamically using the repo content won't work. But maybe updating it through a CI action could help keep it up to date.

PR #700, opened 2024-12-11
Register old proto messages

amimart:

Solve the unmarshalling issues of `sdk.Msg` implementations of previous `x/emissions` and `x/mint` versions by registering related types in the `InterfaceRegistry`.

This change is not consensus breaking as these messages are never manipulated in state mutations, there is no need to update validator nodes with this fix.

It'll allow to query transactions containing old versions of proto messages, for example the query below was returning an error:
```bash
> allorad --node "https://allora-rpc.testnet.allora.network" query tx 74E327C59995366ECF6009AE68122EB60B24815C78B636B01B52B2E96772932C
Error: unable to resolve type URL /emissions.v5.InsertWorkerPayloadRequest: tx parse error
Usage:
  allorad query tx --type=[hash|acc_seq|signature] [hash|acc_seq|signature] [flags]

Flags:
      --grpc-addr string   the gRPC endpoint to use for this chain
      --grpc-insecure      allow gRPC over insecure channels, if not the server must use TLS
      --height int         Use a specific height to query state at (this can error if the node is pruning state)
  -h, --help               help for tx
      --node string        <host>:<port> to CometBFT RPC interface for this chain (default "tcp://localhost:26657")
  -o, --output string      Output format (text|json) (default "text")
      --type string        The type to be used when querying tx, can be one of "hash", "acc_seq", "signature" (default "hash")

Global Flags:
      --home string         directory for config and data (default "/Users/arnaud/.allorad")
      --log_format string   The logging format (json|plain) (default "plain")
      --log_level string    The logging level (trace|debug|info|warn|error|fatal|panic|disabled or '*:<level>,<key>:<level>') (default "info")
      --log_no_color        Disable colored logs
      --trace               print out full stack trace on errors

unable to resolve type URL /emissions.v5.InsertWorkerPayloadRequest: tx parse error [cosmos/cosmos-sdk@v0.50.10/x/auth/tx/decoder.go:44]
```

I'll mention this in the upgrade checklist so we do the same for future versions.

https://linear.app/alloralabs/issue/PROTO-3030/investigate-issue-when-querying-transactions-in-testnet

### DIFF
--- a/x/mint/module/module.go
+++ b/x/mint/module/module.go
@@ -9,6 +9,7 @@ import (
 	"cosmossdk.io/core/store"
 	"cosmossdk.io/depinject"
 	modulev1 "github.com/allora-network/allora-chain/x/mint/api/mint/module/v1"
+	v1beta1 "github.com/allora-network/allora-chain/x/mint/api/mint/v1beta1"
 	"github.com/allora-network/allora-chain/x/mint/keeper"
 	"github.com/allora-network/allora-chain/x/mint/types"
 	"github.com/cosmos/cosmos-sdk/client"
@@ -50,6 +51,7 @@ func (AppModuleBasic) RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
 // RegisterInterfaces registers the module's interface types
 func (b AppModuleBasic) RegisterInterfaces(r cdctypes.InterfaceRegistry) {
 	types.RegisterInterfaces(r)
+	v1beta1.RegisterInterfaces(r)
 }
 
 // DefaultGenesis returns default genesis state as raw bytes for the mint
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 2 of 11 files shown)


guilherme-brandao (approved):
lgtm!

x/mint/module/module.go
```diff
@@ -50,6 +51,7 @@ func (AppModuleBasic) RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
 // RegisterInterfaces registers the module's interface types
 func (b AppModuleBasic) RegisterInterfaces(r cdctypes.InterfaceRegistry) {
 	types.RegisterInterfaces(r)
+	v1beta1.RegisterInterfaces(r)
```
relyt29, line 54: do we need to do this for `v2` `v3` and `v4` of the mint module? Maybe not because they don't affect the message server? v4 does update the query server

relyt29 (approved):

fernandofcampos (approved):

x/mint/module/module.go
amimart: Yes I couldn't find `v3` tx messages, the last `v4` version is already registered and is actually using `v2` messages

xmariachi (approved):
good one

PR #703, opened 2024-12-13
Diego/proto 2855 add outlier detection

xmariachi:

* Add new params `InferenceOutlierDetectionThreshold`, `InferenceOutlierDetectionAlpha` to control outlier-resistant inferences.
* Add stores of MAD ema per topic and last median.
* Use new state stores at `closeWorkerNonce` to compute MAD and last_median.
* Addition of query `network-inferences-at-block-outlier-resistant` that computes such network inferences on the fly, using the new state additions.

Migration created for `v0_8_0` package and `v7` protos. 

- [x] If tested, please describe how. If not, why tests are not needed. -- added unit tests for relevant new functions. Tested against local chain.
- [x] If documented, please describe where. If not, describe why docs are not needed. -- added to autocli


x/emissions/migrations/v6/migrate.go
```diff
@@ -106,9 +106,11 @@ func MigrateParams(store storetypes.KVStore, cdc codec.BinaryCodec) error {
...
 		PNormSafeDiv:                        oldParams.PNormSafeDiv,
 		// NEW PARAMS
-		GlobalWhitelistEnabled:       defaultParams.GlobalWhitelistEnabled,
-		TopicCreatorWhitelistEnabled: defaultParams.TopicCreatorWhitelistEnabled,
-		MinExperiencedWorkerRegrets:  defaultParams.MinExperiencedWorkerRegrets,
+		GlobalWhitelistEnabled:             defaultParams.GlobalWhitelistEnabled,
+		TopicCreatorWhitelistEnabled:       defaultParams.TopicCreatorWhitelistEnabled,
+		MinExperiencedWorkerRegrets:        defaultParams.MinExperiencedWorkerRegrets,
+		InferenceOutlierDetectionThreshold: defaultParams.InferenceOutlierDetectionThreshold,
+		InferenceOutlierDetectionAlpha:     defaultParams.InferenceOutlierDetectionAlpha,
```
kpeluso: No new migration needed because we could simply use the default values and/or set the values in a tx after the upgrade, right?

x/emissions/module/autocli.go
```diff
@@ -572,6 +572,15 @@ func (am AppModule) AutoCLIOptions() *autocliv1.ModuleOptions {
 						{ProtoField: "block_height_last_inference"},
 					},
 				},
+				{
+					RpcMethod: "GetNetworkInferencesAtBlockOutlierResistant",
+					Use:       "network-inferences-at-block-outlier-resistant [topic_id] [block_height_last_inference]",
+					Short:     "Get the Network Inferences for a topic at a block height where the last inference was made removing outliers",
+					PositionalArgs: []*autocliv1.PositionalArgDescriptor{
+						{ProtoField: "topic_id"},
+						{ProtoField: "block_height_last_inference"},
```
kpeluso, line 589: Should have a query to obtain this if we don't already.

And/or consider making a `latest-network-inference-outlier-resistant` query where the block arg is not even needed, paralleling the existing network inference query.

x/emissions/migrations/v6/migrate.go
xmariachi: That was my understanding about the params from previous additive non-migrations - does that also apply to the new stores defined in keeper (lastMedian, madEma) and the new queries?

x/emissions/module/autocli.go
xmariachi: true, will replicate the latest too.

x/emissions/migrations/v6/migrate.go
amimart: If a network is already running a 'v0.7.0' version these default values won't be set.

The additions in the stores are consensus breaking, a node with a previous version must be updated so we should increment the consensus version and add a migration to set default values for new params.

Regarding the proto API I think we can keep the `v6` as the addition of fields is not breaking, marshalling and unmarshalling of the `Params` struct will be backward and forward compatible with the addition of these parameters.

x/emissions/keeper/genesis.go
```diff
@@ -964,6 +964,29 @@ func (k *Keeper) InitGenesis(ctx context.Context, data *types.GenesisState) erro
 		}
 	}
 
+	// LastMedianInferences
+	if len(data.LastMedianInferences) != 0 {
+		for _, topicIdAndDec := range data.LastMedianInferences {
```
amimart:
```suggestion
	for _, topicIdAndDec := range data.LastMedianInferences {
```
This may be simplified I guess

x/emissions/keeper/genesis.go
```diff
@@ -964,6 +964,29 @@ func (k *Keeper) InitGenesis(ctx context.Context, data *types.GenesisState) erro
...
+					return errors.Wrap(err, "error setting lastMedianInferences")
+				}
+			}
+		}
+	}
+
+	// madInferences
+	if len(data.MadInferences) != 0 {
+		for _, topicIdDec := range data.MadInferences {
+			if err := k.SetMadInferences(ctx, topicIdDec.TopicId, topicIdDec.Dec); err != nil {
```
amimart, line 912: There's no not `nil` check  on the `topicIdDec` here

x/emissions/keeper/queryserver/query_server_inferences.go
```diff
@@ -98,6 +99,35 @@ func (qs queryServer) GetNetworkInferencesAtBlock(ctx context.Context, req *emis
 	return &emissionstypes.GetNetworkInferencesAtBlockResponse{NetworkInferences: result.NetworkInferences}, nil
 }
 
+// An outlier resistant version of GetNetworkInferencesAtBlock
+// TODO implement
```
amimart:
```suggestion
```

Seems implemented :)

x/emissions/types/params.go
```diff
@@ -611,3 +619,17 @@ func validateDataSendingFee(i cosmosMath.Int) error {
...
+	}
+	return nil
+}
+
+func validateInferenceOutlierDetectionAlpha(i alloraMath.Dec) error {
+	if err := ValidateDec(i); err != nil {
+		return err
+	}
+	return nil
+}
```
amimart: Is there more we can check? I imagine they should be positive and the alpha between 0 & 1

amimart (commented):
Seems pretty great, I've let some comments I'll let you consider :)

x/emissions/migrations/v6/migrate.go
xmariachi: Ty @amimart. 

> The additions in the stores are consensus breaking, a node with a previous version must be updated so we should increment the consensus version and add a migration to set default values for new params.

You mean that then a v0.8.0 will be needed instead of a v0.7.1, even if we do not need to set the default values (ie the param values can be set after mig, the stores are ok to start empty for the first epoch) ?

amimart: @xmariachi We can omit setting the default values in a migration and set them with an `UpdateParams` tx after, but having new stores in the state is consensus breaking as it changes the computed app hash for the block, this required a coordinated upgrade and that suggests a `v0.8.0` I guess

x/emissions/module/autocli.go
xmariachi: Done for 
network-inferences-at-block-outlier-resistant
latest-network-inferences-outlier-resistant
latest-available-network-inferences-outlier-resistant

x/emissions/keeper/genesis.go
xmariachi: Ty! And actually there are a bunch of these! 
Committing it separately since I also have to indent and drop ending curly brace

x/emissions/keeper/genesis.go
xmariachi: ty! Went over the `InitGenesis` and added where it wasn't

x/emissions/keeper/genesis.go
xmariachi: Done in all places where it happened in `InitGenesis`. Which was basically all places lol :P

x/emissions/types/params.go
xmariachi: good one, improved. 
[0,) in threshold
[0,1] in alpha

x/emissions/migrations/v6/migrate.go
xmariachi: Doing `v0.8.0` migration adding `v7` migration package, keeping `v6` proto API (adding params and queries)

error-070.log
amimart, line 1: May have been committed by mistake 😅

x/emissions/migrations/v6/migrate.go
xmariachi: Pushed `v0.8.0` migration tested locally.

error-070.log
xmariachi: yeah ty, removed  :smiley_cat:

kpeluso (dismissed):
just need to rebase! v nice!

kpeluso (dismissed):

amimart (changes requested):
I think we need to register the `v6` messages in the module `RegisterInterfaces` func, by adding:
```golang
v6.RegisterInterfaces(registry)
```
So that we'll be able to unmarshal old transactions containing `v6` messages through `allorad query tx $OLD_TX_HASH` for example

xmariachi:
> I think we need to register the `v6` messages in the module `RegisterInterfaces` func, by adding:
> 
> ```go
> v6.RegisterInterfaces(registry)
> ```
> 
> So that we'll be able to unmarshal old transactions containing `v6` messages through `allorad query tx $OLD_TX_HASH` for example

I initially did this, but it was not working, failing due to duplicate registration of v6 interfaces. Not entirely sure why though.

kpeluso (dismissed):

relyt29:
> I initially did this, but it was not working, failing due to duplicate registration of v6 interfaces. Not entirely sure why though.

pretty sure codec.go is made in the types folder that registers the "latest" interface always, so you don't have to double register

just the outdated interfaces

math/utils.go
```diff
@@ -567,3 +567,46 @@ func GetQuantileOfDecs(
 	}
 	return ret, nil
 }
+
+// MedianAbsoluteDeviation calculates the median absolute deviation of a slice of Dec values.
+// Returns (MAD, median, error)
+func MedianAbsoluteDeviation(data []Dec) (Dec, Dec, error) {
```
relyt29: totally a nitpick so feel free to ignore if lazy but you could name the variables rather than having to have the comment
```suggestion
// MedianAbsoluteDeviation calculates the median absolute deviation of a slice of Dec values.
func MedianAbsoluteDeviation(data []Dec) (medianAbsoluteDeviation Dec, median Dec, err error) {
```

x/emissions/proto/emissions/v6/params.proto
```diff
@@ -156,5 +156,4 @@ message Params {
   // and participate in all topics as workers and reputers
   bool topic_creator_whitelist_enabled = 51; // topic creator whitelist enabled => only topic creator whitelisted actors can create topics
   uint64 min_experienced_worker_regrets = 52; // minimum number of experienced workers required to use their regrets
-  // for calculating the topic initial regret
```
relyt29: usually this file would not change and be frozen with a new migration...

x/emissions/keeper/keeper.go
```diff
@@ -1072,17 +1137,102 @@ func (k *Keeper) GetLatestTopicInferences(ctx context.Context, topicId TopicId)
...
+		}
 	}
 
-	return &types.Inferences{
-		Inferences: make([]*types.Inference, 0),
-	}, 0, nil
+	return inferences, blockHeight, nil
+}
+
+// UpdateNetworkInferencesOutlierMetrics recalculates the network inferences outlier metrics
```
relyt29, line 1165: I don't love the use of the word "metrics" here because at least to me, it implies some kind of offchain database that is tracking performance data or something.

I'm also just unclear what this function is about based on the function header description. It seems like it updates the "previous" MAD and median values

x/emissions/migrations/v7/migrate.go
```diff
@@ -0,0 +1,114 @@
...
+		EpsilonReputer:                      oldParams.EpsilonReputer,
+		HalfMaxProcessStakeRemovalsEndBlock: oldParams.HalfMaxProcessStakeRemovalsEndBlock,
+		EpsilonSafeDiv:                      oldParams.EpsilonSafeDiv,
+		DataSendingFee:                      oldParams.DataSendingFee,
+		MaxElementsPerForecast:              oldParams.MaxElementsPerForecast,
+		MaxActiveTopicsPerBlock:             oldParams.MaxActiveTopicsPerBlock,
+		MaxStringLength:                     oldParams.MaxStringLength,
+		InitialRegretQuantile:               oldParams.InitialRegretQuantile,
+		PNormSafeDiv:                        oldParams.PNormSafeDiv,
+		GlobalWhitelistEnabled:              oldParams.GlobalWhitelistEnabled,
```
relyt29, line 102: above the comments say that the global whitelist enabled will be flipped on ... but here it seems to use the old value rather than forcing it to true?

does the comment just need to be fixed or does this migration for 0.8.0 emissions version 7 have some intended change to the whitelists on testnet? cc: @kpeluso

relyt29 (dismissed):
concerned about this whitelist thing otherwise the PR is very solid and clearly was a lot of work, so good job

x/emissions/migrations/v7/migrate.go
kpeluso: thanks for flagging, but not sure what whitelist comment you're referring to. whitelist is off for now, and will stay off for now until Foundation has better plans in place for their use. eventually will be on. they will be on in mainnet though

xmariachi:
> > I initially did this, but it was not working, failing due to duplicate registration of v6 interfaces. Not entirely sure why though.
> 
> pretty sure codec.go is made in the types folder that registers the "latest" interface always, so you don't have to double register
> 
> just the outdated interfaces

Yeah but v6 is not the latest :D 
I think this comment I made came from a different version (I had some confusion trying to reuse v6 and when creating the bare minimum v7), so I tried several ways and that error seem to happen then.
So - forget ab my comment.

However, one thing that was needed was to register the _types_ used by the Interfaces. @amimart suggested how, and I pushed in [this commit](https://github.com/allora-network/allora-chain/pull/703/commits/738dbe1124e0f25672ebb5636fc42b44f98d7c5a).

x/emissions/migrations/v7/migrate.go
xmariachi: Those params come from previous version, don't they? 
The flipping was done on `v0_7_0` migration package. This is `v0_8_0`. 

I think the only params from this version are those two re. outlier detection. Please shout otherwise
```
// NEW PARAMS
InferenceOutlierDetectionThreshold: defaultParams.InferenceOutlierDetectionThreshold,
InferenceOutlierDetectionAlpha:     defaultParams.InferenceOutlierDetectionAlpha,
```

x/emissions/keeper/keeper.go
xmariachi: > I don't love the use of the word "metrics" here

You got a point there on the metrics. Couldn't come up with a better name myself, so open to suggestions. "Data" also sound too vague and "stats" (maybe better?) also gives the idea that they are not used in the logic. 

Natif Inglish speaka, halp!

> I'm also just unclear what this function is about based on the function header description. It seems like it updates the "previous" MAD and median values

Indeed. The _metrics_ used for outlier detection.

x/emissions/proto/emissions/v6/params.proto
xmariachi: ty, it slipped in in my refactors. good catch, reverted it.

math/utils.go
xmariachi: good catch, added alongside more stuff so discarding your suggestion :heart:

amimart (approved):
All good for me well done! 💪

kpeluso (approved):
plz triple check all updates have been added here please @xmariachi 

https://www.notion.so/upshothq/Checklist-to-add-an-Upgrade-15234d40db488056970dd61953c20a39

PR #704, opened 2024-12-13
Correct Discrepancies with Whitepaper modified reward fractions vs plain EMA

relyt29:

Really all credit for this PR goes to @cal-hawk who did the hard work of finding the bug and tracing the actual lines of code related to it.

He found a discrepancy between the whitepaper and the allora-chain codebase where u-tilde (inferer, reputer, forecaster reward fractions) were being used for numberRatio calculations when it should be using f (the modifiedRewardFractions). And then also the reverse for setting the previous EMA score - we were using f instead of u-tilde in that case.

and in the other case it's:

f being used where it should be u-tilde

PROTO-3039, RES-579

To be quite honest I don't know of an easy way to test this in a unit test and show that the fixed code addresses issues other than just pointing to the whitepaper and saying read these formulas. The existing unit test suite at least will cover that directionally these changes do not break any major functionality of the chain.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,9 +57,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## v0.X.X
+
+### Added
+
+### Changed
+
+### Deprecated
+
+### Removed
+
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


cal-hawk (dismissed):
lgtm! thanks @relyt29 🙏

fernandofcampos (dismissed):

xmariachi (approved):

kpeluso (approved):

PR #712, opened 2024-12-17
Add sortition penalties

amimart:

Penalise underperforming actors that are missing epochs when they become live again.

A penalty will be applied to their EMA score, the penalty used in the initial EMA score, it is applied as many times as the number of previously missed epochs towards the initial score.

https://linear.app/alloralabs/issue/PROTO-3089/add-sortition-penalties

### DIFF
--- a/math/utils.go
+++ b/math/utils.go
@@ -54,17 +54,20 @@ func CalcEma(
 }
 
 // NCalcEma is a generalisation of calculating the EMA update n times.
+// This function computes = (1-α)^n * x + (1-(1-α)^n) * asymptote
+// As `n` approaches infinity, the function converges to the asymptote.
+// For smaller `n`, the function returns a value closer to `x`, as much as `alpha` affords.
 func NCalcEma(
 	alpha,
-	update,
-	score Dec,
+	asymptote,
+	x Dec,
 	n uint64,
 ) (Dec, error) {
-	if update.isNaN {
-		return ZeroDec(), errorsmod.Wrap(ErrNaN, "NCalcEma update EMA operand should not be NaN")
+	if asymptote.isNaN {
+		return ZeroDec(), errorsmod.Wrap(ErrNaN, "NCalcEma asymptote EMA operand should not be NaN")
 	}
-	if score.isNaN {
-		return ZeroDec(), errorsmod.Wrap(ErrNaN, "NCalcEma score EMA operand should not be NaN")
+	if x.isNaN {
+		return ZeroDec(), errorsmod.Wrap(ErrNaN, "NCalcEma x EMA operand should not be NaN")
 	}
 	if alpha.isNaN {
 		return ZeroDec(), errorsmod.Wrap(ErrNaN, "NCalcEma alpha EMA operand should not be NaN")
@@ -82,7 +85,7 @@ func NCalcEma(
 	if err != nil {
 		return ZeroDec(), err
 	}
-	oneMinusAlphaExpNTimesScore, err := oneMinusAlphaExpN.Mul(score)
+	oneMinusAlphaExpNTimesOldVal, err := oneMinusAlphaExpN.Mul(x)
 	if err != nil {
 		return ZeroDec(), err
 	}
@@ -91,12 +94,12 @@ func NCalcEma(
 	if err != nil {
 		return ZeroDec(), err
 	}
-	UpdateTimesOneMinusOneMinusAlphaExpN, err := update.Mul(oneMinusOneMinusAlphaExpN)
+	UpdateTimesOneMinusOneMinusAlphaExpN, err := asymptote.Mul(oneMinusOneMinusAlphaExpN)
 	if err != nil {
 		return ZeroDec(), err
 	}
 
-	ret, err := oneMinusAlphaExpNTimesScore.Add(UpdateTimesOneMinusOneMinusAlphaExpN)
+	ret, err := oneMinusAlphaExpNTimesOldVal.Add(UpdateTimesOneMinusOneMinusAlphaExpN)
 	if err != nil {
 		return ZeroDec(), err
 	}
@@ -541,7 +544,7 @@ func GetQuantileOfDecs(
 	decs []Dec,
 	quantile Dec,
 ) (Dec, error) {
-	// If there are no decs then the quantile of scores is 0.
+	// If there are no decs then the quantile of the `decs` is 0.
 	// This better ensures chain continuity without consequence because in this situation
 	// there is no meaningful quantile to calculate.
 	if len(decs) == 0 {
--- a/math/utils.go
+++ b/math/utils.go
@@ -53,6 +53,56 @@ func CalcEma(
 	return ret, nil
 }
 
+// NCalcEma is a generalisation of calculating the EMA update n times.
+func NCalcEma(
+	alpha,
+	update,
+	score Dec,
+	n uint64,
+) (Dec, error) {
+	if update.isNaN {
+		return ZeroDec(), errorsmod.Wrap(ErrNaN, "NCalcEma update EMA operand should not be NaN")
+	}
+	if score.isNaN {
+		return ZeroDec(), errorsmod.Wrap(ErrNaN, "NCalcEma score EMA operand should not be NaN")
+	}
+	if alpha.isNaN {
+		return ZeroDec(), errorsmod.Wrap(ErrNaN, "NCalcEma alpha EMA operand should not be NaN")
+	}
+
+	nDec, err := NewDecFromUint64(n)
+	if err != nil {
+		return ZeroDec(), err
+	}
+	oneMinusAlpha, err := OneDec().Sub(alpha)
+	if err != nil {
+		return ZeroDec(), err
+	}
+	oneMinusAlphaExpN, err := Pow(oneMinusAlpha, nDec)
+	if err != nil {
+		return ZeroDec(), err
+	}
+	oneMinusAlphaExpNTimesScore, err := oneMinusAlphaExpN.Mul(score)
+	if err != nil {
+		return ZeroDec(), err
+	}
+
+	oneMinusOneMinusAlphaExpN, err := OneDec().Sub(oneMinusAlphaExpN)
+	if err != nil {
+		return ZeroDec(), err
+	}
+	UpdateTimesOneMinusOneMinusAlphaExpN, err := update.Mul(oneMinusOneMinusAlphaExpN)
+	if err != nil {
+		return ZeroDec(), err
+	}
+
+	ret, err := oneMinusAlphaExpNTimesScore.Add(UpdateTimesOneMinusOneMinusAlphaExpN)
+	if err != nil {
+		return ZeroDec(), err
+	}
+	return ret, nil
+}
+
 // Generic function that sorts the keys of a map
 // Used for deterministic ranging of maps
 func GetSortedKeys[K cmp.Ordered, V any](m map[K]V) []K {
--- a/x/emissions/keeper/actor_penalties.go
+++ b/x/emissions/keeper/actor_penalties.go
@@ -1,6 +1,7 @@
 package keeper
 
 import (
+	"cosmossdk.io/math"
 	alloraMath "github.com/allora-network/allora-chain/math"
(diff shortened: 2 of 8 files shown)


x/emissions/keeper/actor_penalties.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return errorsmod.Wrap(err, "error calculating median absolute deviation")
+	}
+	x, err := standardDeviation.Mul(params.SortitionLambdaPenalty)
+	if err != nil {
+		return errorsmod.Wrap(err, "error calculating penalty")
+	}
+	penalty, err := lowestEMAScore.Score.Sub(x)
+	if err != nil {
+		return errorsmod.Wrap(err, "error calculating penalty")
```
xmariachi:
```suggestion
		return errorsmod.Wrap(err, "error calculating penalty substracting")
```

x/emissions/keeper/actor_penalties.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return errorsmod.Wrap(err, "error getting params")
+	}
+	standardDeviation, _, err := alloraMath.MedianAbsoluteDeviation(scores)
+	if err != nil {
+		return errorsmod.Wrap(err, "error calculating median absolute deviation")
+	}
+	x, err := standardDeviation.Mul(params.SortitionLambdaPenalty)
+	if err != nil {
+		return errorsmod.Wrap(err, "error calculating penalty")
```
xmariachi:
```suggestion
		return errorsmod.Wrap(err, "error calculating penalty multiplying")
```

xmariachi: avoiding same error message

x/emissions/keeper/actor_penalties.go
```diff
@@ -0,0 +1,126 @@
...
+
+	params, err := k.GetParams(ctx)
+	if err != nil {
+		return errorsmod.Wrap(err, "error getting params")
+	}
+	standardDeviation, _, err := alloraMath.MedianAbsoluteDeviation(scores)
+	if err != nil {
+		return errorsmod.Wrap(err, "error calculating median absolute deviation")
+	}
+	x, err := standardDeviation.Mul(params.SortitionLambdaPenalty)
```
xmariachi: change x for a more meaningful name (unless there is a justification on keeping name from paper equations?)

x/emissions/keeper/actor_penalties.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	// Save the penalised EMA score
+	return emaScore, k.SetInfererScoreEma(ctx, topic.Id, emaScore.Address, emaScore)
+}
+
+// applyPenalty applies the penalty to the EMA score for the given number of missed epochs while staying above provided limit.
+func applyPenalty(topic types.Topic, penalty, limit, emaScore alloraMath.Dec, missedEpochs uint32) (alloraMath.Dec, error) {
+	for i := uint32(0); i < missedEpochs; i++ {
+		penalisedScore, err := alloraMath.CalcEma(topic.MeritSortitionAlpha, penalty, emaScore, false)
```
xmariachi: This way of calculating feels like very cumbersome... 
0/ Is this one the one agreed as the way to apply an epoch-based loop penalty?
1/ Should we set a limit of max penalties? Just to try to not get too many epochs (=unbounded) which could affect block stability. Should be agreed though. (This is a common thing we want to avoid in general, worth having at the back of our minds lol)
2/ Could there be an alternative way to compute this penalty?

x/emissions/keeper/actor_penalties.go
```diff
@@ -0,0 +1,126 @@
...
+
+	return emaScore, nil
+}
+
+// CountWorkerContiguousMissedEpochs counts the number of contiguous missed epochs prior to the given nonce, given the
+// last worker submission and the current block heights.
+func countWorkerContiguousMissedEpochs(topic types.Topic, lastSubmissionHeight int64) uint32 {
+	count := uint32(0)
+	prevEpochStart := topic.EpochLastEnded - topic.EpochLength
+	for lastSubmissionHeight < prevEpochStart {
```
xmariachi: Could this be more efficient with a division? 

Like 

```
    if lastSubmissionHeight >= prevEpochStart {
        return 0
    }

    // Calculate missed epochs using division
    // Add 1 to match the original implementation's counting behavior
    missedEpochs := ((prevEpochStart - lastSubmissionHeight) / topic.EpochLength) + 1
    
    return uint32(missedEpochs)
```

amimart: Definitely, this was bad :)

x/emissions/keeper/actor_penalties.go
amimart: We could set a maximum number of `missedEpoch` to consider

xmariachi (changes requested):
Left some comments separately.

Remember adding new stores to genesis too etc. 
and do all checks [here](https://www.notion.so/upshothq/Checklist-to-add-an-Upgrade-15234d40db488056970dd61953c20a39)

x/emissions/keeper/actor_penalties.go
xmariachi: a limit set on emissions params, yep

amimart: It's been replaced by a new formula so no more loop :)

x/emissions/keeper/actor_penalties.go
```diff
@@ -0,0 +1,135 @@
...
+// actor last submission.
+func CountWorkerContiguousMissedEpochs(topic types.Topic, lastSubmissionHeight int64) int64 {
+	prevEpochStart := topic.EpochLastEnded - topic.EpochLength
+	return countContiguousMissedEpochs(prevEpochStart, topic.EpochLength, lastSubmissionHeight)
+}
+
+// CountReputerContiguousMissedEpochs counts the number of contiguous missed epochs prior to the given nonce, given the
+// actor last submission.
+func CountReputerContiguousMissedEpochs(topic types.Topic, lastSubmissionHeight int64) int64 {
+	prevEpochStart := topic.EpochLastEnded - topic.EpochLength + topic.WorkerSubmissionWindow + topic.GroundTruthLag
```
xmariachi: I believe this should be: 

`prevEpochStart := topic.EpochLastEnded - topic.EpochLength - topic.GroundTruthLag`

because the reputer's nonce will be referenced to the original epoch when the inferences were made (= it is same nonce all around, but it is evaluated after `GroundTruthLag` has happened).

xmariachi (dismissed):
Tests need fixing but preemptively approving re. the rest

math/utils.go
```diff
@@ -53,6 +53,56 @@ func CalcEma(
 	return ret, nil
 }
 
+// NCalcEma is a generalisation of calculating the EMA update n times.
+func NCalcEma(
+	alpha,
+	update,
+	score Dec,
```
kpeluso: terms should be more general than this e.g. currentVal, updatedVal. This may not always apply to scores, as this is a general Dec

math/utils.go
```diff
@@ -53,6 +53,56 @@ func CalcEma(
 	return ret, nil
 }
 
+// NCalcEma is a generalisation of calculating the EMA update n times.
```
kpeluso, line 56: Write the arithmetic for every math function

x/emissions/keeper/actor_penalties.go
```diff
@@ -0,0 +1,141 @@
...
+		func(topicId TopicId, score types.Score) error {
+			return k.SetReputerScoreEma(ctx, topicId, score.Address, score)
+		},
+		topic,
+		block,
+		emaScore,
+	)
+}
+
+func MayPenaliseActor(
```
kpeluso: Don't love the name but it's fine.

Great abstractions.

math/utils.go
kpeluso: I did it for you

x/emissions/keeper/actor_penalties.go
```diff
@@ -0,0 +1,141 @@
...
+	prevEpochStart := topic.EpochLastEnded - topic.EpochLength
+	return countContiguousMissedEpochs(prevEpochStart, topic.EpochLength, lastSubmittedNonce)
+}
+
+// CountReputerContiguousMissedEpochs counts the number of contiguous missed epochs prior to the given nonce, given the
+// actor last submission.
+func CountReputerContiguousMissedEpochs(topic types.Topic, lastSubmittedNonce int64) int64 {
+	prevEpochStart := topic.EpochLastEnded - topic.EpochLength - topic.GroundTruthLag
+	return countContiguousMissedEpochs(prevEpochStart, topic.EpochLength, lastSubmittedNonce)
+}
```
kpeluso, line 134: I'm not sure this is right. The reputer is sending payloads at the same `topic.EpochLength` blocks frequency as workers. They are just judging data that was submitted nearly `topic.GroundTruthLag` blocks ago.

kpeluso (changes requested):

x/emissions/keeper/actor_penalties.go
xmariachi: But this data refers back in time. 
Data "being sent now" = scores set now and they carry nonces from the past.

So I think moving all references a gtlag back is OK

x/emissions/keeper/actor_penalties.go
```diff
@@ -0,0 +1,141 @@
...
+// actor last submission.
+func CountWorkerContiguousMissedEpochs(topic types.Topic, lastSubmittedNonce int64) int64 {
+	prevEpochStart := topic.EpochLastEnded - topic.EpochLength
+	return countContiguousMissedEpochs(prevEpochStart, topic.EpochLength, lastSubmittedNonce)
+}
+
+// CountReputerContiguousMissedEpochs counts the number of contiguous missed epochs prior to the given nonce, given the
+// actor last submission.
+func CountReputerContiguousMissedEpochs(topic types.Topic, lastSubmittedNonce int64) int64 {
+	prevEpochStart := topic.EpochLastEnded - topic.EpochLength - topic.GroundTruthLag
```
xmariachi, line 132: Disallow negative blockheights , 

max(0, value)

Apply to other places too if relevant

x/emissions/keeper/actor_penalties.go
xmariachi: Agree on both hehe

amimart: Didn't find better but we can change it :)

x/emissions/keeper/actor_penalties.go
```diff
@@ -0,0 +1,141 @@
...
+	)
+}
+
+func MayPenaliseActor(
+	ctx sdk.Context,
+	missedEpochsFn func(topic types.Topic, lastSubmittedNonce int64) int64,
+	getPenaltyFn func(topicId TopicId) (alloraMath.Dec, error),
+	setScoreFn func(topicId TopicId, score types.Score) error,
+	topic types.Topic,
+	block types.BlockHeight,
```
kpeluso: rename: nonceBlockHeight

x/emissions/keeper/actor_penalties.go
```diff
@@ -0,0 +1,141 @@
...
+		topic,
+		block,
+		emaScore,
+	)
+}
+
+func MayPenaliseActor(
+	ctx sdk.Context,
+	missedEpochsFn func(topic types.Topic, lastSubmittedNonce int64) int64,
+	getPenaltyFn func(topicId TopicId) (alloraMath.Dec, error),
```
kpeluso: rename: `getAsymptoteFn`

x/emissions/keeper/actor_penalties.go
kpeluso: ApplyLivenessPenaltyToActor

x/emissions/keeper/actor_penalties.go
kpeluso: good call

kpeluso: @amimart added L137 below

kpeluso (approved):

PR #717, opened 2024-12-19
Table tests and fix zerolength array as error

xmariachi:
Fix access to empty array with test


amimart (approved):
lgtm

amimart (approved):

PR #720, opened 2024-12-19
Add initial ema score generation queries and events

guilherme-brandao:
Added necessary queries and events to facilitate inspection and monitoring of the initial EMA scores generation

https://linear.app/alloralabs/issue/PROTO-3082/fix-actor-score-ema-initialization


x/emissions/proto/emissions/v7/query.proto
```diff
@@ -1592,3 +1611,36 @@ message GetActiveInferersForTopicRequest {
 message GetActiveInferersForTopicResponse {
   repeated string inferers = 1;
 }
+
```
fernandofcampos, line 1614: consider adding (amino.dont_omitempty) = true otherwise it will remove the parameter if the value is zero (or zero value).

fernandofcampos: The comment was meant to be generic, not tied to a specific definition. I see there's non repeated tags (bellow) of type math.Dec, will it be returned as well even if zero, as they use a custom type?

fernandofcampos (dismissed):

kpeluso (approved):

PR #723, opened 2024-12-19
fix: adapt missed epoch calculation on worker activity

amimart:

Brings two fix to the penalty calculation introduced by https://github.com/allora-network/allora-chain/pull/712:

1. When a participant submits for the first time he is penalised because we assign it a score with the block height of 0, this is now checked to avoid the penalty

2. As scores of participants in the active set are updated when we compute the rewards, active workers are penalised when they shouldn't, we now change the missed epoch calculation depending on active or not

https://linear.app/alloralabs/issue/PROTO-3089/add-sortition-penalties

### DIFF
--- a/x/emissions/keeper/actor_penalties.go
+++ b/x/emissions/keeper/actor_penalties.go
@@ -1,8 +1,6 @@
 package keeper
 
 import (
-	"fmt"
-
 	"cosmossdk.io/math"
 	alloraMath "github.com/allora-network/allora-chain/math"
 	"github.com/allora-network/allora-chain/x/emissions/types"
@@ -80,7 +78,6 @@ func ApplyLivenessPenaltyToActor(
 	missedEpochs := missedEpochsFn(topic, emaScore.BlockHeight)
 	// No missed epochs == no penalty
 	if missedEpochs == 0 {
-		fmt.Println("--- NO PENALTY FOR ACTOR ---", nonceBlockHeight, emaScore)
 		ctx.Logger().Debug("no liveness penalty on actor",
 			"nonce", nonceBlockHeight,
 			"score", emaScore,
@@ -99,11 +96,6 @@ func ApplyLivenessPenaltyToActor(
 		return types.Score{}, err
 	}
 
-	fmt.Println("--- APPLY PENALTY TO ACTOR ---", "nonce", nonceBlockHeight,
-		"missed", missedEpochs,
-		"penalty", penalty,
-		"before", beforePenalty,
-		"after", emaScore)
 	ctx.Logger().Debug("apply liveness penalty on actor",
 		"nonce", nonceBlockHeight,
 		"missed", missedEpochs,
--- a/x/emissions/keeper/actor_penalties.go
+++ b/x/emissions/keeper/actor_penalties.go
@@ -103,8 +103,6 @@ func ApplyLivenessPenaltyToActor(
 		"before", beforePenalty,
 		"after", emaScore,
 	)
-
-	// Save the penalised EMA score
 	return emaScore, nil
 }
 
--- a/x/emissions/keeper/actor_penalties.go
+++ b/x/emissions/keeper/actor_penalties.go
@@ -1,13 +1,15 @@
 package keeper
 
 import (
+	"fmt"
+
 	"cosmossdk.io/math"
 	alloraMath "github.com/allora-network/allora-chain/math"
 	"github.com/allora-network/allora-chain/x/emissions/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-// ApplyLivenessPenaltyToInferer penalises an inferer for missing previous epochs. It saves and returns the new EMA score.
+// ApplyLivenessPenaltyToInferer penalises an inferer for missing previous epochs. It only returns the updated EMA score.
 // If the inferer didn't miss any epochs this is a no-op, the EMA score is returned as is.
 func (k *Keeper) ApplyLivenessPenaltyToInferer(
 	ctx sdk.Context,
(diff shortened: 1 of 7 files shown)


x/emissions/keeper/actor_penalties.go
```diff
@@ -85,26 +105,39 @@ func ApplyLivenessPenaltyToActor(
...
-	emaScore.BlockHeight = nonceBlockHeight
 
 	beforePenalty := emaScore
+	emaScore.BlockHeight = nonceBlockHeight
 	emaScore.Score, err = applyPenalty(topic, penalty, emaScore.Score, missedEpochs)
 	if err != nil {
 		return types.Score{}, err
 	}
 
 	ctx.Logger().Debug("apply liveness penalty on actor",
```
xmariachi, line 99: is this logging format Ok ? Like actually showing the text but not the data?
It will compile but it may not show the data. Using fmt.Sprintf can build up the string ensuring that.

x/emissions/keeper/actor_penalties_test.go
```diff
@@ -78,6 +115,41 @@ func (s *KeeperTestSuite) TestApplyLivenessPenaltyToForecaster() {
...
+func (s *KeeperTestSuite) TestApplyLivenessPenaltyToActiveForecaster() {
+	ctx := s.ctx
+	keeper := s.emissionsKeeper
+
+	givenTopic := types.Topic{
+		Id:                  uint64(1),
+		MeritSortitionAlpha: alloraMath.MustNewDecFromString("0.1"),
+		EpochLastEnded:      105,
+		EpochLength:         10,
+		GroundTruthLag:      5,
```
xmariachi: gt_lag needs to be equal or bigger than epochLength, not smaller. 
See https://github.com/allora-network/allora-chain/blob/6b2e5304c809773f84a6155ce091fe5fb8b02f6a/x/emissions/types/validations.go#L375

Separately, we should tetst with GTLag >> EpochLength which is where most problems will arise, since this leaves several rounds for active to overwrite inactive and so on.

x/emissions/keeper/actor_penalties.go
amimart: Testing that right now

x/emissions/keeper/actor_penalties_test.go
```diff
@@ -260,7 +332,62 @@ func TestCountWorkerContiguousMissedEpochs(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+// nolint: exhaustruct
+func TestCountActiveWorkerContiguousMissedEpochs(t *testing.T) {
+	topic := types.Topic{
+		EpochLastEnded: 105,
+		EpochLength:    10,
```
xmariachi, line 235: sam ehere about the topic definition validations, which are not run here but may lead to confusion

xmariachi (dismissed):
Some comments. Overalll lgtm though so I approve preemptively

x/emissions/keeper/actor_penalties_test.go
kpeluso: +1 here

x/emissions/keeper/actor_penalties.go
spooktheducks: This is definitely the preferred way, because it gives us the ability to destructure logs (either in json or logfmt) in Grafana and do lots of handy things with the extracted data.  If we're seeing the fields missing from the log output, that's a configuration issue.

x/emissions/keeper/actor_penalties.go
```diff
@@ -80,54 +73,63 @@ func ApplyLivenessPenaltyToActor(
...
 	getAsymptoteFn func(topicId TopicId) (alloraMath.Dec, error),
-	setScoreFn func(topicId TopicId, score types.Score) error,
 	topic types.Topic,
 	nonceBlockHeight types.BlockHeight,
 	emaScore types.Score,
 ) (types.Score, error) {
 	missedEpochs := missedEpochsFn(topic, emaScore.BlockHeight)
 	// No missed epochs == no penalty
 	if missedEpochs == 0 {
+		fmt.Println("--- NO PENALTY FOR ACTOR ---", nonceBlockHeight, emaScore)
```
guilherme-brandao: remove log

x/emissions/keeper/actor_penalties.go
```diff
@@ -80,54 +73,63 @@ func ApplyLivenessPenaltyToActor(
...
 	emaScore.Score, err = applyPenalty(topic, penalty, emaScore.Score, missedEpochs)
 	if err != nil {
 		return types.Score{}, err
 	}
 
+	fmt.Println("--- APPLY PENALTY TO ACTOR ---", "nonce", nonceBlockHeight,
+		"missed", missedEpochs,
+		"penalty", penalty,
+		"before", beforePenalty,
+		"after", emaScore)
```
guilherme-brandao: remove log

guilherme-brandao (dismissed):
Minor comments! preemptively approving.

guilherme-brandao (commented):
I think it would be good to add tests for `AppendInference`, `AppendForecast` and `AppendReputerLoss`, since is where the penalty flow is triggered.

x/emissions/keeper/actor_penalties.go
amimart: It is well structured!

x/emissions/keeper/keeper_test.go
```diff
@@ -5361,3 +5361,232 @@ func (s *KeeperTestSuite) TestInitialEmaScoreSettingInAppendReputer() {
...
+	}
+
+	topic, err := k.GetTopic(ctx, topic.Id)
+	s.Require().NoError(err)
+
+	// Append the inference
+	err = k.AppendInference(ctx, topic, blockHeight, inference, 4)
+	s.Require().NoError(err)
+
+	// Verify the worker received the initial EMA score
```
kpeluso:
```suggestion
	// Verify the worker's EMA score trended toward the topic initial score especially when there is a lapse in their liveness
```

kpeluso: here and for the 2 others

kpeluso (approved):

PR #740, opened 2025-01-25
fix: prevent git stash errors when no changes are present

mavci42:
- Added a check before running 'git stash' to avoid errors when no changes exist.
* Ensures smooth build process without unnecessary failures.
* Improves the stability of Docker builds by handling edge cases correctly.


spooktheducks:
Incorporated in #914

PR #747, opened 2025-02-14
New stdnorm calc + weights onchain

xmariachi:
This PR implements the rework on weights that fixes the latest outlier issues observed from experience and provides a cutoff on low weights for the stddev. 

* Adds stores for latestStdnorm and latestWeights
* regrets returned on calculation
* At `CloseReputerNonce`, calculate stdnorm based on old weights, store it and recalculate weights. 
* stdnorm used as normalization for all weights calculation cases.
* refactor and additional tests on weights.
* emits events on new stdnorm set, and on each new actor weight 
* autocli query addition of latest stdnorm, latest inferer weight and latest forecaster weight
* adds v0.9.0 migration

Tested with allora-simulator. The expectations from research are met.
 Internal implementation details, no need to post to the user. 

## Still Left Todo

* OPTIONAL: Optimization on weights from other actors taken from the latest ones. Implement a function that fills these weights from the store.
  * instead of calling "CalcWeightsGivenWorkers", take them from the store - however they will be different in the one-out  cases because the elements differ.


x/emissions/keeper/actor_utils/losses.go
```diff
@@ -145,6 +145,84 @@ func CloseReputerNonce(
 		return err
 	}
 
+	// Calculate the regret_stdnorm and the weights (multistep process).
```
felipead, line 149: Given the mission critical nature of these changes, can you cover them with some unit tests?

x/emissions/keeper/genesis.go
```diff
@@ -963,6 +963,33 @@ func (k *Keeper) InitGenesis(ctx context.Context, data *types.GenesisState) erro
 		}
 	}
 
+	// Initialize latest regret stdnorm
+	for _, stdnorm := range data.LatestRegretStdNorm {
```
felipead, line 967: same here...

x/emissions/keeper/genesis.go
```diff
@@ -2445,6 +2472,59 @@ func (k *Keeper) ExportGenesis(ctx context.Context) (*types.GenesisState, error)
 		})
 	}
 
+	// Export latest regret stdnorm
```
felipead, line 2475: ...and here

x/emissions/keeper/actor_utils/losses.go
felipead: I understand we are in a hurry here, but trust me - unit tests are the fastest way to ensure your changes are going to work in a distributed production environment such as the blockchain. 

Imagine if you ship a bug, then a new release cycle just to fix it and all the time wasted.

xmariachi: Haha, no need to convince me, I am a big fan myself of proper and comprehensive unit testing (for example, this is why a new `weights_test.go` file has been added which tests not only the new code but also some of the old code, and I have refactored to unit-test each step separately 😄  ).

Each one of the functions below in the `CloseWorkerNonce` ~should already have~ has unit test coverage (although I'll take this chance to review width and depth of these, so ty for putting stress on unit testing - it is never enough).

Re the `CloseWorkerNonce` as a whole, this is covered by older unit tests and also a new specific one for the new code (ie stdnorm and weights) under `rewards_test.go`.

xmariachi: BTW I pushed several commits on some missing tests, esp basic ones on query and keeper, and more importantly on CalcWeightGivenWorkers.

x/emissions/keeper/genesis.go
xmariachi: This is covered in genesis_test.go, although the coverage there is indeed basic. 
However this exceeds this ticket and release - also there are so many areas where unit testing can be improved.

x/emissions/keeper/msgserver/msg_server_params_test.go
kpeluso, line 1: you're missing adding the parameter to `msg_server_params.go`

please be sure to [consult this checklist](https://www.notion.so/upshothq/Adding-New-Global-Parameters-Checklist-6216783e3b4f4280a2fda6f3487f0164) in adding new global parameters

xmariachi: ty, good catch. added. Checklist complete.

x/emissions/keeper/actor_utils/losses.go
```diff
@@ -145,6 +145,84 @@ func CloseReputerNonce(
...
+	// 0. Get inferer and forecaster regrets
+	infererRegrets := regrets.InfererRegrets
+	inferers := make([]synth.Worker, 0)
```
amimart:
```suggestion
	inferers := make([]synth.Worker, 0, len(infererRegrets))
```

We can pass the capacity to make a single allocation and avoid allocation+copy when appending

x/emissions/keeper/actor_utils/losses.go
```diff
@@ -145,6 +145,84 @@ func CloseReputerNonce(
...
+	for inferer := range infererRegrets {
+		inferers = append(inferers, inferer)
+	}
+	forecasterRegrets := regrets.ForecasterRegrets
+	forecasters := make([]synth.Worker, 0)
```
amimart:
```suggestion
	forecasters := make([]synth.Worker, 0, len(forecasterRegrets))
```

Same here

x/emissions/keeper/actor_utils/losses.go
```diff
@@ -145,6 +145,84 @@ func CloseReputerNonce(
...
+	for forecaster := range forecasterRegrets {
+		forecasters = append(forecasters, forecaster)
+	}
+
```
kpeluso: these will result in nondeterminisms. you need to sort these arrays before you use them. there exist a util for this already you should use

x/emissions/migrations/v7/migrate.go
```diff
@@ -115,6 +115,7 @@ func MigrateParams(ctx sdk.Context, store storetypes.KVStore, cdc codec.BinaryCo
 		GlobalReputerWhitelistEnabled:      defaultParams.GlobalReputerWhitelistEnabled,
 		GlobalAdminWhitelistAppended:       defaultParams.GlobalAdminWhitelistAppended,
 		MaxWhitelistInputArrayLength:       defaultParams.MaxWhitelistInputArrayLength,
+		MinWeightThresholdForStdnorm:       defaultParams.MinWeightThresholdForStdnorm,
```
amimart: This migration from consensus version 6 to 7 has already been performed on the `v0.8.0` software upgrade so this will never been executed.

We should increment the consensus version of the `x/emissions` module, register a new migration for it in `x/emissions/module/module.go`, and introduce the `v0.9.0` migration at the app level.

I think it's ok to not introduce a new API version as we're only adding a parameter and some rpc queries, this is not a breaking change in proto3.

amimart (changes requested):
Great job thx! 
I've made some remarks, the main point is to introduce a new migration for the next software upgrade to `v0.9.0`.

kpeluso (changes requested):
Few comments + Upgrade Flow then gtg

x/emissions/keeper/genesis.go
xmariachi: replied above

x/emissions/keeper/actor_utils/losses.go
xmariachi: fixed, using GetSortedKeys now. Great catch.

x/emissions/migrations/v7/migrate.go
xmariachi: We are also adding a param + two stores on keeper as well, plus queries, though these are additive and not proto-breaking indeed.
However, shouldn't it be just cleaner to generate new protos then?

xmariachi: pushed complete migration, undoing previous changes to v7, upgrade tested locally from v0.8.3.

x/emissions/migrations/v8/migrate.go
```diff
@@ -0,0 +1,120 @@
...
+		GlobalWhitelistEnabled:              oldParams.GlobalWhitelistEnabled,
+		TopicCreatorWhitelistEnabled:        oldParams.TopicCreatorWhitelistEnabled,
+		MinExperiencedWorkerRegrets:         oldParams.MinExperiencedWorkerRegrets,
+		InferenceOutlierDetectionThreshold:  defaultParams.InferenceOutlierDetectionThreshold,
+		InferenceOutlierDetectionAlpha:      defaultParams.InferenceOutlierDetectionAlpha,
+		LambdaInitialScore:                  defaultParams.LambdaInitialScore,
+		GlobalWorkerWhitelistEnabled:        defaultParams.GlobalWorkerWhitelistEnabled,
+		GlobalReputerWhitelistEnabled:       defaultParams.GlobalReputerWhitelistEnabled,
+		GlobalAdminWhitelistAppended:        defaultParams.GlobalAdminWhitelistAppended,
+		MaxWhitelistInputArrayLength:        defaultParams.MaxWhitelistInputArrayLength,
```
amimart: We should keep the old params values for them as they're already in the state

x/emissions/migrations/v7/migrate.go
amimart: I'm not sure this is cleaner as it brings a breaking change in the API and all the clients will need to be updated, but I'm fine with both practices

x/emissions/migrations/v8/migrate.go
xmariachi: Done! good catch

kpeluso (dismissed):

kpeluso (dismissed):

x/emissions/keeper/inference_synthesis/weight.go
```diff
@@ -389,3 +498,82 @@ func CalcWeightFromNormalizedRegret(
...
+	}
+
+	return nil
+}
+
+// StoreLatestNormalizedWeights sets the latest weights for the given topic
+func StoreLatestNormalizedWeights(ctx sdk.Context, k keeper.Keeper, topicId TopicId, weights RegretInformedWeights) error {
+	// Set inferer weights. nolint maprange: order is not relevant
+	for worker, weight := range weights.Inferers { // nolint: maprange
+		err := k.SetLatestInfererWeight(ctx, topicId, worker, weight)
```
amimart: I think the order of update in the state should be deterministic, the underlying IAVL tree could be balanced differently

x/emissions/keeper/inference_synthesis/weight.go
```diff
@@ -389,3 +498,82 @@ func CalcWeightFromNormalizedRegret(
...
+		if err != nil {
+			return errorsmod.Wrapf(err, "error setting latest inferer weight for worker %s", worker)
+		}
+	}
+
+	// Set forecaster weights. nolint maprange: order is not relevant.
```
amimart: Same here

x/emissions/keeper/inference_synthesis/weight.go
kpeluso: yes i think so too

amimart (approved):
LGTM

kpeluso (approved):

x/emissions/keeper/inference_synthesis/weight.go
xmariachi: great catch, ty - better to go safe.

fernandofcampos (approved):
LGTM. Great job, especially with the test coverage.

felipead (approved):

PR #748, opened 2025-02-17
Add nonce to rewards events

xmariachi:

* Emit nonce (and also blockheight_tx) on rewards events. This aligns `rewards` with other events that use `nonce`.

- [x] If tested, please describe how. If not, why tests are not needed. -- Covered by unit tests. 
- [x] If documented, please describe where. If not, describe why docs are not needed. -- no need, this change does not modify behaviour


fernandofcampos (dismissed):

kpeluso (commented):
great!

kpeluso (approved):

x/emissions/module/rewards/rewards.go
```diff
@@ -577,12 +577,13 @@ func payoutRewards(
 	ctx sdk.Context,
 	k keeper.Keeper,
 	rewards []types.TaskReward,
+	nonce BlockHeight,
```
felipead, line 580: 👍

x/emissions/types/events_utils.go
```diff
@@ -75,11 +75,12 @@ func NewRewardsSetEventBase(actorType ActorType, blockHeight BlockHeight, reward
...
-		TopicId:     topicId,
-		BlockHeight: blockHeight,
-		Addresses:   addresses,
-		Rewards:     rewardValues,
+		ActorType:     actorType,
+		TopicId:       topicId,
+		BlockHeight:   blockHeight,
+		Addresses:     addresses,
+		Rewards:       rewardValues,
+		BlockHeightTx: blockHeightTx,
```
felipead, line 83: 👍

x/emissions/proto/emissions/v7/events.proto
```diff
@@ -47,6 +47,7 @@ message EventRewardsSettled {
     (gogoproto.customtype) = "github.com/allora-network/allora-chain/math.Dec",
     (gogoproto.nullable) = false
   ];
+  int64 block_height_tx = 6;
```
felipead, line 50: 👍

x/emissions/types/events_utils.go
```diff
@@ -66,7 +66,7 @@ func NewEMAScoresSetEventBase(actorType ActorType, scores []Score, activations m
 /// Rewards
 
 // Assumes length of `rewards` is at least 1
-func NewRewardsSetEventBase(actorType ActorType, blockHeight BlockHeight, rewards []TaskReward) proto.Message {
+func NewRewardsSetEventBase(actorType ActorType, blockHeight, blockHeightTx BlockHeight, rewards []TaskReward) proto.Message {
```
felipead, line 69: I just didn't understand the difference between `blockHeight` and `blockHeightTx`, can you explain better?

Does the former mean current block height, and the later is the actual nonce?

If you can, please use better variable names to make this information more explicit.

xmariachi: `blockHeight` is the blockHeight that this event is referring to. Every action on the topic lifecycle is related to a "nonce". This nonce is what defines a particular cycle. 
A cycle contains a worker cycle (inference / forecasting), a reputer cycle, and ends with rewards computation. 
So the events emitted at any time during the cycle are related to that nonce (which is the starting blockHeight of the cycle).

Separately, these events are emitted at different block times, typically, where the transaction / operation happens. This has been expressed as `blockheightTx` in other events too, to express "the current blockheight at which the tx is happening", so here we are aligning with other events too.

amimart (approved):
LGTM

PR #754, opened 2025-02-27
Fix error message inconsistency in whitelist.go

DeVikingMark:
Corrected error messages in whitelist functions to properly reference "actor id" instead of "admin id" in validation error messages. This improves error clarity and consistency when validating actor identifiers.


spooktheducks:
Incorporated in #914

PR #757, opened 2025-03-04
ENGN-3538 ENGN-3539 Incoming value limiting (with migration)

xmariachi:
* Added `DataLimitExponent` to limit the size and precision of the incoming data. Generous enough to avoid under/overflows. 
*  Migration to v0.10.0 with tests
* Fixed v2 migration Params (added old and new types for v1 and v2) - it was failing due to the Params usage on validations.
* Added validations on incoming data with new validation_test.go. 
* Unit tests to ensure on keeper incoming data InsertInference / InsertForecasts / InsertLosses.

Tested via:  existing unit tests, added unit tests for logic and migration, and also run the full migration using upgrade_test.go successfully. ✅  
- [x] If documented, please describe where. If not, describe why docs are not needed. -- added in README on inf synth.


x/emissions/types/validations.go
```diff
@@ -19,6 +21,52 @@ var (
...
+func ValidateIncomingDataDec(value alloraMath.Dec, params Params) error {
+	if ValidateDec(value) != nil {
+		return errors.Wrap(sdkerrors.ErrInvalidType, "value is invalid")
+	}
+	if value.IsZero() {
+		return nil
+	}
+
+	baseStringPositive := "1e"
+	baseStringNegative := "-1e"
```
amimart: Maybe worth validating the absolute value, this would simplify the verifications

amimart (commented):
I'd like to discuss the need of a global parameter to limit input data, because I think this should be part of the API contract and not be allowed to change over time for an API version. We could also take some benefits of this by making these validations before the messages being routed.

We could for example introduce a new type for that purpose that would encapsulate the logic related to these restrictions by implementing gogoproto marshalling interfaces, and specifying this type in the proto spec would help the API being self-documented. And as gRPC clients would also use this type they can be sure they adhere to the API contract.

Let me know what you think

Dockerfile.upgrade
```diff
@@ -2,8 +2,8 @@ FROM golang:1.22-bookworm AS builder
 
 ENV GO111MODULE=on
 ENV GOMODCACHE=/gocache
-ENV BASELINE_VERSION_TAG=v0.8.0
```
kpeluso, line 5: is this in a checklist somewhere? can it be automated?

kpeluso: answer: yes

https://www.notion.so/upshothq/Checklist-to-add-an-Upgrade-15234d40db488056970dd61953c20a39

x/emissions/keeper/genesis.go
```diff
@@ -449,7 +449,7 @@ func (k *Keeper) InitGenesis(ctx context.Context, data *types.GenesisState) erro
 	//AllLossBundles []*TopicIdBlockHeightReputerValueBundles
 	for _, topicIdBlockHeightReputerValueBundles := range data.AllLossBundles {
 		if topicIdBlockHeightReputerValueBundles != nil {
-			if err := topicIdBlockHeightReputerValueBundles.ReputerValueBundles.Validate(); err != nil {
+			if err := topicIdBlockHeightReputerValueBundles.ReputerValueBundles.Validate(data.Params); err != nil {
```
kpeluso: if the value of the parameter got lowered at some point before this was ran, then this could fail

x/emissions/keeper/keeper.go
```diff
@@ -1430,10 +1430,14 @@ func (k *Keeper) InsertInference(
...
 ) error {
+	params, err := k.GetParams(ctx)
+	if err != nil {
+		return errorsmod.Wrap(err, "error getting params")
+	}
 	if err := types.ValidateTopicId(topicId); err != nil {
 		return errorsmod.Wrap(err, "topic id validation failed")
 	}
-	err := inference.Validate()
+	err = inference.Validate(params)
```
kpeluso: I don't really love how so many of the validations need to become stateful now because of this, and lowering this value affects many values stored in state anyway. I also don't view this param as being something gov is likely to change much. Therefore, I'd like to open a conversation as to whether it's best to have this be a global constant instead of variable => Changing it will require a network upgrade, which may also implicate changing many values stored in state. @xmariachi @amimart

amimart: Yes I agree, I've noted some elements related to that in a [comment](https://github.com/allora-network/allora-chain/pull/757#pullrequestreview-2671386002)

xmariachi:
> I'd like to discuss the need of a global parameter to limit input data, because I think this should be part of the API contract and not be allowed to change over time for an API version. We could also take some benefits of this by making these validations before the messages being routed.
> 
> We could for example introduce a new type for that purpose that would encapsulate the logic related to these restrictions by implementing gogoproto marshalling interfaces, and specifying this type in the proto spec would help the API being self-documented. And as gRPC clients would also use this type they can be sure they adhere to the API contract.
> 
> Let me know what you think

Interesting point. Thing is we currently are not sure about how this will operate in practice, and if we in the future would like to allow bigger numbers.

Re. the type, we are already using the Dec type, and we are ok with it being bigger as a result from the calculations - we just want to limit the incoming values, so we are happy to keep the whole space for Dec - just not for incoming data. A new type just for incoming data could be indeed created to clarify that and sgtm. However, how to act in case of change of data type - would be another version `vX` of the proto? That may not need a param nor global variable in that case.

x/emissions/keeper/keeper.go
xmariachi: commented [here](https://github.com/allora-network/allora-chain/pull/757#issuecomment-2711288533) I think encapsulating that def in a type makes it cleaner

amimart:
> However, how to act in case of change of data type - would be another version `vX` of the proto? That may not need a param nor global variable in that case.

We are changing the type of an existing field so that's a breaking change in the proto spec, we should increase the API & Consensus version of the `x/emissions` module, this is already done though :)

xmariachi:
Converted to draft to impl changes on type.

math/dec.go
```diff
@@ -40,14 +40,15 @@ const mathCodespace = "math"
...
-	ErrOutOfRange         = errorsmod.Register(mathCodespace, 8, "value is out of range")
+	ErrInvalidDecString       = errorsmod.Register(mathCodespace, 1, "invalid decimal string")
+	ErrUnexpectedRounding     = errorsmod.Register(mathCodespace, 2, "unexpected rounding")
+	ErrNonIntegral            = errorsmod.Register(mathCodespace, 3, "value is non-integral")
+	ErrInfiniteString         = errorsmod.Register(mathCodespace, 4, "value is infinite")
+	ErrOverflow               = errorsmod.Register(mathCodespace, 5, "overflow")
+	ErrNaN                    = errorsmod.Register(mathCodespace, 6, "NaN not permitted in this context")
+	ErrNotMatchingLength      = errorsmod.Register(mathCodespace, 7, "slices are not of the same length")
+	ErrOutOfRange             = errorsmod.Register(mathCodespace, 8, "value is out of range")
+	ErrInvalidBoundedExp40Dec = errorsmod.Register(mathCodespace, 9, "invalid bounded exp40 dec")
```
amimart: It could also make sense to reuse the already existing `ErrOutOfRange`, wdyt?

x/emissions/proto/emissions/v9/inputworker.proto
```diff
@@ -0,0 +1,68 @@
+syntax = "proto3";
+package emissions.v9;
+
+import "emissions/v3/nonce.proto";
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/allora-network/allora-chain/x/emissions/types";
+
+message BoundedInference {
```
amimart: About naming wdyt if we prefix these messages with `Input` instead, e.g. `InputInference`? I think it's better here to indicate that it's an input type than that it's bounded

x/emissions/keeper/msgserver/msg_server_reputer_payload.go
```diff
@@ -16,32 +16,38 @@ func (ms msgServer) InsertReputerPayload(ctx context.Context, msg *types.InsertR
...
+	if err != nil {
+		return nil, errorsmod.Wrapf(err, "Error getting params for reputer: %v", &msg.ReputerValueBundle.ValueBundle.Reputer)
+	}
+	rvb, err := msg.ReputerValueBundle.Convert()
+	if err != nil {
+		return nil, errorsmod.Wrapf(err,
+			"Reputer bad data format for block: %d", blockHeight)
+	}
+
+	err = rvb.Validate()
```
amimart: Just an idea: we could also provide a constructor `func NewReputerValueBundleFromInput(rvb BoundedReputerValueBundle) (*ReputerValueBundle, error)` that performs the validation, a constructor is expected to output a valid state, and we minimize the risk of forgetting the validation step

amimart (commented):
That looks good, I've noted some suggestions I'll let you consider.
I'll need to review more deeply but the approach sounds good to me

x/emissions/types/validations.go
xmariachi: done

x/emissions/keeper/genesis.go
xmariachi: By using input layer validation, this would not be required anymore. Chain is able to store `Dec` in a stateless way and all validations occur on input layer.

x/emissions/keeper/keeper.go
xmariachi: Resolve this convo bc of new input layer validations makes this stateless.

x/emissions/proto/emissions/v9/inputworker.proto
xmariachi: good idea

math/dec.go
xmariachi: true

x/emissions/keeper/msgserver/msg_server_reputer_payload.go
xmariachi: yeah makes sense

x/emissions/keeper/keeper.go
```diff
@@ -1916,7 +1916,8 @@ func (k *Keeper) InsertNetworkLossBundleAtBlock(
 	if err := types.ValidateBlockHeight(block); err != nil {
 		return errorsmod.Wrap(err, "block height validation failed")
 	}
-	if err := lossBundle.Validate(); err != nil {
+	err := lossBundle.Validate()
+	if err != nil {
```
```suggestion
	if err := lossBundle.Validate(); err != nil {
```
Maybe better to keep it as it was

x/emissions/migrations/v9/migrate.go
```diff
@@ -0,0 +1,118 @@
...
+func MigrateStore(ctx sdk.Context, emissionsKeeper keeper.Keeper) error {
+	ctx.Logger().Info("STARTING EMISSIONS MODULE MIGRATION FROM VERSION 8 TO VERSION 9")
+	ctx.Logger().Info("MIGRATING STORE FROM VERSION 8 TO VERSION 9")
+	storageService := emissionsKeeper.GetStorageService()
+	store := runtime.KVStoreAdapter(storageService.OpenKVStore(ctx))
+	cdc := emissionsKeeper.GetBinaryCodec()
+
+	ctx.Logger().Info("MIGRATING PARAMS FROM VERSION 8 TO VERSION 9")
+	// This also flips on global and topic creator whitelists
+	if err := MigrateParams(ctx, store, cdc); err != nil {
```
amimart, line 31: We have no changes in the params right? I think we don't need to migrate them, btw the comments seems copy pasted from and old migration and not relevant here

x/emissions/types/conversions.go
```diff
@@ -0,0 +1,307 @@
+package types
+
+import (
+	"cosmossdk.io/errors"
+)
+
+// NewInferenceFromInput converts InputInference to Inference
+func (bi *InputInference) NewInferenceFromInput() (*Inference, error) {
```
amimart:
```suggestion
func NewInferenceFromInput(bi *InputInference) (*Inference, error) {
```
Not a big deal, it's just more idiomatic to take inputs as arguments for a constructor

amimart (commented):
Thanks that looks pretty good! I've let some minor remarks I'll let you consider

x/emissions/migrations/v9/migrate.go
xmariachi: yeah I left that in because we'll have another migration in the same release :P

x/emissions/types/conversions.go
xmariachi: applied to all functions

x/emissions/migrations/v9/migrate.go
amimart: Oh ok makes sense 😉

kpeluso (approved):

amimart (approved):

PR #758, opened 2025-03-04
Fix: ensure stake removal processing

amimart:

When processing stake removals in the `x/emissions` end blocker there's a limit to the number of stakes to process which when reached was causing an error, resulting with the stake removals until not to be processed.

The resolution consists of not returning such error anymore but emitting a log instead, allowing to process these removals.

Another fix is covered here: in the case stake removals were not process in the block they were expected to be completed because of this limit, they were not managed afterwards because the provided block when removing them was systematically the current block.

Some refactoring has been necessary to allow this module part to be testable by mocking the emissions keeper.

https://linear.app/alloralabs/issue/ENGN-3494/135-logic-error-in-the-removestakes-function-could-lead-to-a-denial-of


xmariachi (dismissed):
lgtm

kpeluso (dismissed):
perfect

kpeluso (approved):

PR #760, opened 2025-03-04
ENGN-3256 totalsumprevioustopicweights can be set to any value

xmariachi:
* Fix vulnerability: if topic has not been (re)activated, then the topic weights should not change.
* Change `isAdded, error` for just `error` on `addTopicToActiveSetRespectingLimitsWithoutMinWeightReset`, adding some error types. The flag was not really adding value, while not processing the flag would lead to a vulnerability like the one found.

- [x] If tested, please describe how. If not, why tests are not needed. -- existing unit tests 
- [x] If documented, please describe where. If not, describe why docs are not needed. -- no need, changes in functions not user-facing.

### DIFF
--- a/x/emissions/keeper/topic_activation.go
+++ b/x/emissions/keeper/topic_activation.go
@@ -279,7 +279,7 @@ func (k *Keeper) ActivateTopic(ctx context.Context, topicId TopicId) error {
 	epochEndBlock := currentBlock + topic.EpochLength
 
 	err = k.activateTopicAndResetLowestWeightAtBlock(ctx, topicId, epochEndBlock)
-	if errors.IsOf(err, types.ErrTopicAlreadyActive) || errors.IsOf(err, types.ErrTopicCannotBeActivated) {
+	if errors.IsOf(err, types.ErrTopicAlreadyActive, types.ErrTopicCannotBeActivated) {
 		sdkCtx.Logger().Info(fmt.Sprintf("Failed to add topic at next epoch %d, %d, %v", topicId, epochEndBlock, err))
 		return nil
 	}
--- a/x/emissions/keeper/topic_activation.go
+++ b/x/emissions/keeper/topic_activation.go
@@ -116,7 +116,7 @@ func (k *Keeper) inactivateTopicWithoutMinWeightReset(ctx context.Context, topic
 		return errors.Wrap(err, "failed to check if topic exists")
 	}
 	if !topicExists {
-		return nil
+		return types.ErrTopicDoesNotExist
 	}
 
 	// Check if this topic is activated or not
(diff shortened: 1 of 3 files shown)


x/emissions/keeper/topic_activation.go
```diff
@@ -339,7 +343,10 @@ func (k *Keeper) AttemptTopicReactivation(ctx context.Context, topicId TopicId)
 	}
 
 	err = k.activateTopicAndResetLowestWeightAtBlock(ctx, topicId, epochEndBlock)
-	if err != nil {
+	if errors.IsOf(err, types.ErrTopicAlreadyActive) || errors.IsOf(err, types.ErrTopicCannotBeActivated) {
```
amimart:
```suggestion
	if errors.IsOf(err, types.ErrTopicAlreadyActive, types.ErrTopicCannotBeActivated) {
```
This can be simplified

x/emissions/keeper/topic_activation.go
```diff
@@ -279,6 +279,10 @@ func (k *Keeper) ActivateTopic(ctx context.Context, topicId TopicId) error {
 	epochEndBlock := currentBlock + topic.EpochLength
 
 	err = k.activateTopicAndResetLowestWeightAtBlock(ctx, topicId, epochEndBlock)
+	if errors.IsOf(err, types.ErrTopicAlreadyActive) || errors.IsOf(err, types.ErrTopicCannotBeActivated) {
```
amimart:
```suggestion
	if errors.IsOf(err, types.ErrTopicAlreadyActive, types.ErrTopicCannotBeActivated) {
```

amimart (dismissed):
Sounds good thanks, just noticed minor simplification

amimart (dismissed):

guilherme-brandao (dismissed):

amimart (dismissed):

amimart (approved):

fernandofcampos (approved):

x/emissions/keeper/topic_activation.go
```diff
@@ -375,15 +382,10 @@ func (k *Keeper) removeCurrentTopicFromBlock(ctx context.Context, topicId TopicI
...
 	if err != nil {
-		sdkCtx.Logger().Warn(fmt.Sprintf("Failed to add topic at next epoch %d, %d", topicId, epochEndBlock))
 		return errors.Wrap(err, "failed to add topic to active set respecting limits without min weight reset")
 	}
-	if !isAdded {
-		sdkCtx.Logger().Info(fmt.Sprintf("Failed to add topic at next epoch %d, %d", topicId, epochEndBlock))
-		return nil
-	}
 
 	err = k.SetTopicToNextPossibleChurningBlock(ctx, topicId, epochEndBlock)
```
kpeluso, line 390: You say:
> Fix vulnerability: if topic has not been (re)activated, then the topic weights should not change.

but that's precisely what this if-statement you took out was doing... so i'm confused...

xmariachi: Same logic is applied, only that the err mgmt is one layer up, to the public function `ActivateTopic` .
The new code on `activateTopicAndResetLowestWeightAtBlock` will currently return the incoming error instead of `nil`directly, and its error handling has been moved one layer up, where different logic can be applied depending on the error case. The management of the errors makes it more explicit what to do and on which cases.
 
```
if errors.IsOf(err, types.ErrTopicAlreadyActive, types.ErrTopicCannotBeActivated) { ... 
```

guilherme-brandao (approved):

PR #761, opened 2025-03-04
fix close worker nonce boundaries

xmariachi:
Fix worker nonce boundaries. Currently they are wider than they should be, and conceptually they were wrongly defined.
This will not have a direct impact on inner workings.

- [x] If tested, please describe how. If not, why tests are not needed. -- running unit tests
- [x] If documented, please describe where. If not, describe why docs are not needed. -- no need, changes align with current definition.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 * [#757](https://github.com/allora-network/allora-chain/pull/757) Incoming value limiting
+* [#761](https://github.com/allora-network/allora-chain/pull/761) Fix close worker nonce boundaries
 
 ### Security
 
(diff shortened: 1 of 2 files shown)


amimart (dismissed):

kpeluso:
this change was made on purpose here:
https://github.com/allora-network/allora-chain/commit/4b037016a3a421b674f1262382828022eed4863d

trying to recover original meaning. the comments, which should normally convey this, suggest the current PR's changes are correct, not what is currently there

xmariachi:
> this change was made on purpose here: [4b03701](https://github.com/allora-network/allora-chain/commit/4b037016a3a421b674f1262382828022eed4863d)
> 
> trying to recover original meaning. the comments, which should normally convey this, suggest the current PR's changes are correct, not what is currently there

The only gotcha I can see here, is if we ever want to turn the `CloseWorkerNonce` into a transaction (instead of sth that runs on `EndBlocker` and thus in a particular block we can control), which is triggered externally. Or if we modify when it is called (potentially another `EndBlocker`), so it can be called within the period of time where `WorkerSubmissionWindow`(wsw) ends, up to the end of the epoch (`nonce + EpochLength`).

I believe the current code may be based on old assumptions.
Allowing to close the window until the ground truth appears may seem sensible at first. However, this would mean allowing to keep such number of worker windows open - we are now keeping one only. Also, because wsw is no bigger than EpochLength, this again does not make sense - but prior to the introduction of wsw, it may have. 

While it is true that, if we want to ever change when the worker nonce gets closed, we need to extend this window definition.

amimart (approved):

guilherme-brandao (approved):

PR #769, opened 2025-03-13
Caching NetworkInferences

xmariachi:
* Stores Network Inferences and Outlier-Resistant Network Inferences
* Keeps them the same amount of time, pruning as usual after rewards are done. 
* They get filled up at `CloseWorkerNonce`. A test suite has been added to `worker_test.go`(there were none) to test validity of the function in general.
* Related migration changes related (genesis, query) 
* Removes duplicity of `GetLatestNetworkAvailableInferences` and `GetLatestNetworkInferences`. Keeping the latter name.
* Fix reward_test.go tests force-advancing to the end of the wsw.
* Removes coefficients, weights and also lossBlockHeight from the `GetLatestNetworkInferencesResponse`.
___
* Adds validations for input layer objects

- [x] If tested, please describe how. If not, why tests are not needed. -- Current unit test fixed, new unit tests added. 
- [x] If documented, please describe where. If not, describe why docs are not needed. -- This is an internalization and fixing of issues, though these are not documented nor need further documentation.


x/emissions/proto/emissions/v9/query.proto
```diff
@@ -13,8 +12,8 @@ import "emissions/v3/stake.proto";
 import "emissions/v3/topic.proto";
 import "emissions/v3/types.proto";
 import "emissions/v3/worker.proto";
-// v9
-import "emissions/v9/params.proto";
+// v8
+import "emissions/v8/params.proto";
```
amimart, line 16: We should use `v9` and not rely on old versions. We already rely on some old messages versions but this should be addressed at some point to ensure a self-sufficient version

x/emissions/proto/emissions/v9/query.proto
```diff
@@ -162,15 +161,15 @@ service QueryService {
...
+  // }
 
-  rpc GetLatestAvailableNetworkInferencesOutlierResistant(GetLatestAvailableNetworkInferencesOutlierResistantRequest) returns (GetLatestAvailableNetworkInferencesOutlierResistantResponse) {
-    option (cosmos.query.v1.module_query_safe) = true;
-    option (google.api.http).get = "/emissions/v9/latest_available_network_inferences_outlier_resistant/{topic_id}";
-  }
+  // rpc GetLatestAvailableNetworkInferencesOutlierResistant(GetLatestAvailableNetworkInferencesOutlierResistantRequest) returns (GetLatestAvailableNetworkInferencesOutlierResistantResponse) {
+  //   option (cosmos.query.v1.module_query_safe) = true;
+  //   option (google.api.http).get = "/emissions/v9/latest_available_network_inferences_outlier_resistant/{topic_id}";
+  // }
```
amimart: Can be removed

amimart (commented):
Nice work, I've let a few remarks

amimart (dismissed):

guilherme-brandao (approved):

guilherme-brandao (changes requested):
Just did a test adding max inferers and forecasters in the network inference generation and it took approx 300ms for just one topic. We need to discuss this number before moving forward since now we will be processing the network inference in the `EndBlock`.

x/emissions/keeper/actor_utils/worker.go
```diff
@@ -103,6 +104,39 @@ func CloseWorkerNonce(k *keeper.Keeper, ctx sdk.Context, topic types.Topic, nonc
 		return err
 	}
 
+	// Calculate and store network inferences for this block.
+	networkInferencesResult, err := synth.GetNetworkInferences(
```
amimart, line 107: From what I understand this under the hood fetches data from the `allInferences` & `allForecasts` state collections which are filled just before from `inferences` & `forecasts` collections, if we keep them in memory we can access them only once and reduce state access

xmariachi: very good point. Inferences is big and so we could be improving IO a lot there.

xmariachi: Done for inferences

xmariachi:
> Just did a test adding max inferers and forecasters in the network inference generation and it took approx 300ms for just one topic. We need to discuss this number before moving forward since now we will be processing the network inference in the `EndBlock`.

Different other efforts are being done separately to improve performance.

x/emissions/types/validations.go
```diff
@@ -32,6 +32,18 @@ func ValidateDec(value alloraMath.Dec) error {
 	return nil
 }
 
+func ValidateBoundedExp40Dec(value alloraMath.BoundedExp40Dec) error {
```
amimart: The `value.IsFinite()` check is already present [here](https://github.com/allora-network/allora-chain/blob/fce74cb0544c9ccac5fecb7d1c2d11bb3e5fcafb/math/bounded_dec.go#L35), maybe we can just add the `!value.isNan()` there

amimart (dismissed):
LGTM, just a minor remark up to you :)

xmariachi:
Rebased to @guilherme-brandao 's work on #772. 
Also found a problem while testing with offchain nodes: latest inferences are taking the last worker commit which is wrong. Should just get the latest from the store.

x/emissions/types/validations.go
xmariachi: ty! actually `IsFinite()` checks against `NaN` internally too, so the `isNaN` check is redundant, so I'll remove it and change with a more descriptive error msg :P

amimart: So I think we can remove this `ValidateBoundedExp40Dec` func as this is already checked when unmarshaling

xmariachi: correct - there is no operations in-between input layer and validation, so this can be dropped. Pushed.

amimart (approved):
Looks good! 🚀

guilherme-brandao (approved):

PR #771, opened 2025-03-17
Fix reputer/worker window boundaries, tests added

xmariachi:
* Fix: Reputer remainder lag incorrectly computed.
* Alignment of inclusiveness (all inclusive, consistent in both reputer and worker windows) so as to remove confusion.
* Overflow protection.
* Redundant check removed on worker submission window at `EndBlocker`.
* Exhaustive unit tests for both windows

✅  Added new unit tests
✅ Ran all tests and integration tests
- [x] If documented, please describe where. If not, describe why docs are not needed. -- Added `Data Submission` section to `./README.md`.


README.md
```diff
@@ -347,8 +347,16 @@ ssh -NL 2345:localhost:2345 user@remote.ip
 
 9- At Cursor, go to the debug pannel, pick "Remote Debug" configuration and start debugging.
 
+## Data Submission Windows
 
-### References
+Data is submitted by workers and reputers within the allowed submission windows.
+
+For workers, the submission window is defined at the topic level, as the blocks starting at `nonce.BlockHeight` and ending at `nonce.BlockHeight + topic.WorkerSubmissionWindow`.
```
amimart, line 354: It may be worth to specify if these bounds are inclusive or exclusive

x/emissions/keeper/actor_utils/__debug_bin3232807552
amimart, line 1: Sounds like it could be removed :)

x/emissions/keeper/keeper.go
```diff
@@ -3996,17 +3997,44 @@ func (k *Keeper) PruneReputerNonces(ctx context.Context, topicId uint64, blockHe
...
+	if math.MaxInt64-lowerBound < topic.WorkerSubmissionWindow {
+		return false, errorsmod.Wrapf(types.ErrInvalidValue,
+			"nonce block height %d is too high, adding window %d would overflow",
+			nonce.BlockHeight,
+			topic.WorkerSubmissionWindow)
+	}
+	return lowerBound <= blockHeight && blockHeight <= upperBound, nil
 }
 
 // Return true if the nonce is within the worker submission window for the topic
```
```suggestion
// Return true if the nonce is within the reputer submission window for the topic
```

x/emissions/keeper/keeper.go
```diff
@@ -3996,17 +3997,44 @@ func (k *Keeper) PruneReputerNonces(ctx context.Context, topicId uint64, blockHe
...
-// Inclusive of the start block height and exclusive of the end block height
-func (k *Keeper) BlockWithinWorkerSubmissionWindowOfNonce(topic types.Topic, nonce types.Nonce, blockHeight int64) bool {
-	return nonce.BlockHeight <= blockHeight && blockHeight < topic.WorkerSubmissionWindow+nonce.BlockHeight
+// Inclusive of the start block height and of the end block height.
+func (k *Keeper) BlockWithinWorkerSubmissionWindowOfNonce(topic types.Topic, nonce types.Nonce, blockHeight int64) (bool, error) {
```
```suggestion
func BlockWithinWorkerSubmissionWindowOfNonce(topic types.Topic, nonce types.Nonce, blockHeight int64) (bool, error) {
```

We may not attach this func to the `Keeper` if we don't need to access its internal elements

x/emissions/keeper/keeper.go
```diff
@@ -3996,17 +3997,44 @@ func (k *Keeper) PruneReputerNonces(ctx context.Context, topicId uint64, blockHe
...
+		return false, errorsmod.Wrap(types.ErrInvalidValue, "worker submission window cannot be 0")
+	}
+	lowerBound := nonce.BlockHeight
+	upperBound := nonce.BlockHeight + topic.WorkerSubmissionWindow
+	if math.MaxInt64-lowerBound < topic.WorkerSubmissionWindow {
+		return false, errorsmod.Wrapf(types.ErrInvalidValue,
+			"nonce block height %d is too high, adding window %d would overflow",
+			nonce.BlockHeight,
+			topic.WorkerSubmissionWindow)
+	}
```
```suggestion
	lowerBound := nonce.BlockHeight
	if math.MaxInt64-lowerBound < topic.WorkerSubmissionWindow {
		return false, errorsmod.Wrapf(types.ErrInvalidValue,
			"nonce block height %d is too high, adding window %d would overflow",
			nonce.BlockHeight,
			topic.WorkerSubmissionWindow)
	}
	upperBound := nonce.BlockHeight + topic.WorkerSubmissionWindow
```

It would overflow before checking for it :)
BTW do you think this is worth checking? If we manage `nonce.BlockHeight` only internally we should not have problems before billions of years, and then it won't be our problem anymore 😉

x/emissions/keeper/keeper.go
```diff
@@ -3996,17 +3997,44 @@ func (k *Keeper) PruneReputerNonces(ctx context.Context, topicId uint64, blockHe
...
+		extraLag = topic.EpochLength - extraLag
+	}
+	// The block at which ground truth is revealed
+	revealedGroundTruthBlock := nonce.ReputerNonce.BlockHeight + topic.GroundTruthLag
+
+	// Allow 1 EpochLength + any extra lag as defined above.
+	lowerBound := revealedGroundTruthBlock
+	upperBound := revealedGroundTruthBlock + extraLag + topic.EpochLength
+
+	if math.MaxInt64-lowerBound < topic.WorkerSubmissionWindow {
```
amimart: Sounds like wrong check, if we want to check I think there's both the lower and upper bounds to check here

x/emissions/keeper/actor_utils/__debug_bin3232807552
xmariachi: done and added rule to gitignore

x/emissions/keeper/keeper.go
xmariachi: Good point. I think I want to change it to another file though.

x/emissions/keeper/keeper.go
xmariachi: I had my doubts as well but I normally prefer to bulletproof. I was thinking in potential malicious topic definitions.

amimart: Better to be bulletproof than the opposite :),  and yeah if see that we don't restrict values on window related configs so better to check these

x/emissions/keeper/keeper.go
xmariachi: done

x/emissions/keeper/keeper.go
xmariachi: yeah, worker submission window does not apply here. Fixed.

amimart (dismissed):
LGTM

guilherme-brandao (dismissed):

amimart (approved):

guilherme-brandao (approved):

PR #772, opened 2025-03-18
Add one out forecast-implied inference in network inference bundle

guilherme-brandao:
This PR adds a new feature to calculate forecast-implied inferences when each inferer is removed from the network. These values are critical for evaluating the influence of individual inferers on forecasters' implied predictions.

- Added `GetOneOutInfererForecastImpliedInferences` function to calculate what each forecaster's implied inference would be when each inferer is removed
- Updated `CalcNetworkInferences` to include these values in the network inference bundle
- Used existing `OneOutInfererForecasterValues` type to maintain consistent structure with other one-out calculations

### Performance Related Changes:
- The GetNetworkInferences function employs an in-memory cache that temporarily stores results of expensive mathematical operations (in this case, just `Exp`) to avoid redundant calculations during network inference processing; the cache is enabled at the function's start, utilized during calculations, and then intentionally disabled and cleared upon completion.
- Other minor optimizations.

✅  Added new unit tests
- [x] If documented, please describe where. If not, describe why docs are not needed. -- Added `Data Submission` section to `./README.md`.


xmariachi (commented):
This contains a triple-nested for.
I am concerned about the time this will take when executing in an endBlocker.
Could we time-measure this by having a full max inferers and max forecasters test?
If this has an impact we want to know ... it could add some non-negligible time to the block
cc @spooktheducks @amimart

amimart:
>  it could add some non-negligible time to the block

It seems to be used on queries only so it should not impact block production

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -107,6 +107,150 @@ func getForecastImpliedInferences(
...
+						filteredInfererToRegret[inferer] = regret
+					}
+				}
+			}
+
+			// Get this forecaster's forecast and filter out the withheld inferer
+			forecast, ok := args.ForecasterToForecast[forecaster]
+			if !ok {
+				continue
+			}
```
amimart: This should be done the soonest we can, i.e. at the beginning of the first loop

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -107,6 +107,150 @@ func getForecastImpliedInferences(
...
+				if inferer != withheldInferer {
+					filteredInferers = append(filteredInferers, inferer)
+					if inference, ok := args.InfererToInference[inferer]; ok {
+						filteredInfererToInference[inferer] = inference
+					}
+					if regret, ok := args.InfererToRegret[inferer]; ok {
+						filteredInfererToRegret[inferer] = regret
+					}
+				}
+			}
```
amimart, line 169:
```suggestion
		for _, withheldInferer := range args.Inferers {
			// Filter out the inferer we want to withhold
			filteredInferers = filteredInferers[:0]
			copy(filteredInferers, args.Inferers)
			filteredInferers[i] = filteredInferers[0]
			filteredInferers = filteredInferers[1:]
			
			filteredInfererToInference := maps.Clone(args.InfererToInference)
			delete(filteredInfererToInference, withheldInferer)
			filteredInfererToRegret := maps.Clone(args.InfererToRegret)
			delete(filteredInfererToRegret, withheldInferer)
```

Here's a proposal for better performance:
- Allocate `filteredInferers ` only once by reseting its slice every loop iteration (should be allocated with `filteredInferers := make([]Inferer, 0, len(args.Inferers))` at the beginning of this func);
- Faster map copy using `maps.Clone(..)` as it rely on compiler implementation;

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -107,6 +107,150 @@ func getForecastImpliedInferences(
...
+			if !ok {
+				continue
+			}
+
+			filteredForecastElements := make([]*emissions.ForecastElement, 0)
+			for _, element := range forecast.ForecastElements {
+				if element.Inferer != withheldInferer {
+					filteredForecastElements = append(filteredForecastElements, element)
+				}
+			}
```
amimart, line 176:
```suggestion
			filteredForecastElements = filteredForecastElements[:0]
			copy(filteredForecastElements, forecast.ForecastElements)
			if i := slices.IndexFunc(forecast.ForecastElements, func(element *emissions.ForecastElement) bool {
				return element.Inferer == withheldInferer
			}); i != -1 {
				filteredForecastElements[i] = filteredForecastElements[0]
				filteredForecastElements = filteredForecastElements[1:]
			}
```

Performance improvement as well, the `filteredForecastElements ` should be allocated at the beginning of the func

amimart (commented):
I've let some time/alloc optimization remarks to see if we can improve it

x/emissions/keeper/inference_synthesis/network_inference_builder.go
guilherme-brandao: Thank you for the comment! I added changes similar to the one mentioned with the goal of optimizing memory usage. Unfortunately, the changes did not have any impact on the function’s speed.

math/cache.go
```diff
@@ -0,0 +1,92 @@
...
+
+// MathOperationsCacheKey represents a key for the mathematical operations cache
+type MathOperationsCacheKey struct {
+	Op   string
+	Base string
+}
+
+// MathOperationsCache provides a cache for expensive mathematical operations
+type MathOperationsCache struct {
+	mu      sync.RWMutex
```
amimart: I don't think this is needed, in the context we use it it's not async

math/dec.go
```diff
@@ -372,20 +396,42 @@ func Ln(x Dec) (Dec, error) {
...
 		return Dec{}, errorsmod.Wrapf(ErrNaN, "cannot Exp a NaN %s", x.String())
 	}
+
+	// Only use cache if operations cache is enabled (implementation from cache.go)
+	key := MathOperationsCacheKey{
+		Op:   "exp",
+		Base: x.String(),
+	}
+
+	if cached, ok := mathOperationsCache.Get(key); ok {
```
amimart: I wonder if the cache layer should be managed here, this seems not compliant with the separation of concerns principle.

Wdyt about managing the cache directly [here](https://github.com/allora-network/allora-chain/blob/4318716dceea3f316338236fc0e0db29e7fa4b6a/x/emissions/keeper/inference_synthesis/weight.go#L500) where we calculate the gradient, we may even gain a little more in performance. So the math layer is not responsible for managing the cache, we manage it only where we need it

amimart (commented):
Left some comments that may be great to discuss

math/dec.go
xmariachi: makes sense, worth trying. This would defer the decision of using cache or not to each usage, right?

guilherme-brandao: yeah makes sense! Just added the suggestion

x/emissions/keeper/inference_synthesis/network_inference_builder.go
xmariachi: Yes, if I'm reading this right this can be done at line 146, since it is read-only. Good catch.

guilherme-brandao: Good catch! Added

amimart (approved):
LGTM 👍

xmariachi (approved):

xmariachi:
> > it could add some non-negligible time to the block
> 
> It seems to be used on queries only so it should not impact block production

I think this was left unreplied, but just for completeness: this is internalized into endblocker in another PR

PR #773, opened 2025-03-18
Update the topic's state and weight on RemoveStakes()

fernandofcampos:
The issue:
- The RemoveStakes and RemoveDelegateStakes functions were reducing a topic's staking amount
- However, they were not updating the topic's weight or the totalSumPreviousTopicWeights
- This was causing unfair distribution of rewards, as inactive topics or topics with reduced stake were still receiving a share of rewards they shouldn't

The fix:
- When stake is removed from a topic, the topic's weight is immediately recalculated
- The totalSumPreviousTopicWeights value is updated accordingly
- This ensures that reward distribution is fair and accurate
- Topics that should receive a larger share now do

https://linear.app/alloralabs/issue/ENGN-3511/136-the-removestakes-function-does-not-update-the-topics-state-or

## Still Left Todo

### DIFF
--- a/x/emissions/keeper/keeper.go
+++ b/x/emissions/keeper/keeper.go
@@ -2342,6 +2342,11 @@ func (k *Keeper) RemoveReputerStake(
 		return errorsmod.Wrapf(err, "Setting total stake failed")
 	}
 
+	// Calculate new weight and update totalSumPreviousTopicWeights
+	if err := k.updateTopicWeightAfterStakeChange(ctx, topicId); err != nil {
+		return err
+	}
+
 	// remove stake withdrawal information
 	err = k.DeleteStakeRemoval(ctx, blockHeight, topicId, reputer)
 	if err != nil {
@@ -2494,6 +2499,12 @@ func (k *Keeper) RemoveDelegateStake(
 	if err := k.SetTotalStake(ctx, totalStakeNew); err != nil {
 		return errorsmod.Wrapf(err, "Setting total stake failed")
 	}
+
+	// Calculate new weight and update totalSumPreviousTopicWeights
+	if err := k.updateTopicWeightAfterStakeChange(ctx, topicId); err != nil {
+		return err
+	}
+
 	if err := k.DeleteDelegateStakeRemoval(ctx, stakeRemovalBlockHeight, topicId, reputer, delegator); err != nil {
 		return errorsmod.Wrapf(err, "Deleting delegate stake removal from queue failed")
 	}
@@ -3053,6 +3064,7 @@ func (k *Keeper) GetPreviousTopicWeight(ctx context.Context, topicId TopicId) (t
 }
 
 // Set the previous weight during rewards calculation for a topic
+// This function also updates the total sum of previous topic weights
 func (k *Keeper) SetPreviousTopicWeight(ctx context.Context, topicId TopicId, weight alloraMath.Dec) error {
 	if err := types.ValidateTopicId(topicId); err != nil {
 		return errorsmod.Wrap(err, "topic id validation failed")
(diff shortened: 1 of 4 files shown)


guilherme-brandao (changes requested):
Looks good! Just need to add tests that check the topic weight recalculation

fernandofcampos:
> Looks good! Just need to add tests that check the topic weight recalculation

Done.

x/emissions/keeper/keeper.go
```diff
@@ -2194,6 +2210,65 @@ func (k *Keeper) RemoveReputerStake(
...
+		)
+		if err != nil {
+			return errorsmod.Wrap(err, "error calculating new topic weight")
+		}
+
+		// Update previous topic weight
+		if err := k.SetPreviousTopicWeight(ctx, topicId, newWeight); err != nil {
+			return errorsmod.Wrapf(err, "Setting previous topic weight failed")
+		}
+
```
xmariachi: some of this code could be refactorized? it is also used at inactivateTopicWithoutMinWeightReset

x/emissions/keeper/keeper.go
```diff
@@ -2194,6 +2210,65 @@ func (k *Keeper) RemoveReputerStake(
 		return errorsmod.Wrapf(err, "Setting total stake failed")
 	}
 
+	// If the topic was active, calculate new weight and update totalSumPreviousTopicWeights
+	if isActive && !noPrior {
```
xmariachi: I need to think about it more, but I think we don't want to check activity here? 

ie if the topic is not active we still allow to remove stake, and in that case its weights change, regardless of topic being active or not. It may be reactivated at some point. And the total `totalSumPreviousTopicWeights` would not get the right number.

We want to check whether this topic had an actual previous topic weight, I believe. And if so, then apply the operation regardless of activity.
wdyt?

xmariachi (commented):
Left some comments

x/emissions/keeper/keeper.go
amimart: I second that

fernandofcampos: I agree

x/emissions/keeper/keeper.go
fernandofcampos: Done

x/emissions/keeper/keeper.go
fernandofcampos: Done

guilherme-brandao (dismissed):

x/emissions/keeper/keeper.go
```diff
@@ -2194,6 +2200,13 @@ func (k *Keeper) RemoveReputerStake(
 		return errorsmod.Wrapf(err, "Setting total stake failed")
 	}
 
+	// If the topic had a previous topic weight, calculate new weight and update totalSumPreviousTopicWeights
+	if !noPrior {
+		if err := k.updateTopicWeightAfterStakeChange(ctx, topicId); err != nil {
+			return err
+		}
+	}
+
```
amimart: This check it should be made in the `Keeper#updateTopicWeightAfterStakeChange` func to avoid duplication IMO

amimart (changes requested):
I noticed that there's an [end blocker updating topic weights](https://github.com/allora-network/allora-chain/blob/43a8632de73ee75e398f672b5738a1979c1d749f/x/emissions/module/abci.go#L40), AFAIU topic weights depends on stake only, so since we now update topic weights when stake is added or removed we should removed this end blocker cc @xmariachi @guilherme-brandao

x/emissions/keeper/keeper.go
xmariachi: > topic weights depends on stake only

This is not correct - it also depends on the `topicFeeRevenue`. See `GetTargetWeight` (which current version, in another PR, removes the epochLength).

> we should removed this end blocker

I don't think so, since this 
1. gets the active topics due this block (which is block-based)
2. updates their topic weights.

amimart: Oh yes understood thx, we should keep the end blocker then :p

About this check on the topic having previous weight, I still think it should be done in `Keeper#updateTopicWeightAfterStakeChange`

amimart (dismissed):
LGTM 👍

guilherme-brandao (dismissed):

x/emissions/keeper/keeper.go
xmariachi: Nice refactor, ty. 
One note: currently there is a similar operation on `inactivateTopicWithoutMinWeightReset` that also touches this. 

Although it is not in the initial scope, I think that code that performs the removal should be refactored too so the code that removes a topic weights is similarly specified as the current other methods in keeper, whose only responsibility is to do just that.
 
ie. 
do a new method 
`func (k *Keeper) RemoveTopicFromPreviousTopicWeights(ctx context.Context, topicId TopicId) error {`
that contains the related code, at first glance seems to be this code:  

```
    // Remove previous weight from total sum of previous topic weights, but keep on list
	totalSumPreviousTopicWeights, err := k.totalSumPreviousTopicWeights.Get(ctx)
	if err != nil {
		if !errors.IsOf(err, collections.ErrNotFound) {
			return errors.Wrap(err, "failed to get total sum of previous topic weights")
		}
		totalSumPreviousTopicWeights = alloraMath.ZeroDec()
	}

	previousTopicWeight, err := k.previousTopicWeight.Get(ctx, topicId)
	if err != nil {
		if !errors.IsOf(err, collections.ErrNotFound) {
			return errors.Wrap(err, "failed to get previous topic weight")
		}
		// no previous weight, no need to remove

		return nil
	}
	totalSumPreviousTopicWeights, err = totalSumPreviousTopicWeights.Sub(previousTopicWeight)
	if err != nil {
		return errors.Wrap(err, "failed to subtract previous topic weight from total sum of previous topic weights")
	}
	err = k.totalSumPreviousTopicWeights.Set(ctx, totalSumPreviousTopicWeights)
	if err != nil {
		return errors.Wrap(err, "failed to set total sum of previous topic weights")
	}
```

ty

xmariachi (changes requested):
Please do not merge this just yet; there is a quirk I want to check with research re. cadence of updates. Setting review as "Request Changes" to block it explicitly

x/emissions/keeper/keeper.go
fernandofcampos: Done

xmariachi:
> I noticed that there's an [end blocker updating topic weights](https://github.com/allora-network/allora-chain/blob/43a8632de73ee75e398f672b5738a1979c1d749f/x/emissions/module/abci.go#L40), AFAIU topic weights depends on stake only, so since we now update topic weights when stake is added or removed we should removed this end blocker cc @xmariachi @guilherme-brandao

Hey @amimart , topic weights depend on reputers' stake _and topicFeeRevenue_. In the `EndBlocker` the `topicFeeRevenue` is modified in the `EndBlocker` (where the feeRevenue is "dripped") so we shouldn't remove it from there, I believe.

amimart:
> Hey @amimart , topic weights depend on reputers' stake _and topicFeeRevenue_. In the `EndBlocker` the `topicFeeRevenue` is modified in the `EndBlocker` (where the feeRevenue is "dripped") so we shouldn't remove it from there, I believe.

@xmariachi Yeah you're right we need to keep this `EndBlocker`

xmariachi (dismissed):
I think it is consistent with the other topic operations (which happen at that time too). 
LGTM, please rebase.

amimart (dismissed):

guilherme-brandao (dismissed):

amimart (approved):

xmariachi (approved):
lgtm!

PR #777, opened 2025-03-20
Allow forecasts with unregistered inferers to proceed

fernandofcampos:

This addresses the vulnerability where inferer deregistrations would  cause a forecaster's entire submission to fail. Now, when an inferer is not registered in a topic, we skip that specific inferer in the forecast rather than failing the whole transaction.

- Modified message handler to continue processing when an inferer is unregistered
- Added test case to verify the fix works correctly

https://linear.app/alloralabs/issue/ENGN-3519/172-inferer-de-registrations-will-cause-forecasting-submission-to

## Still Left Todo

### DIFF
--- a/x/emissions/keeper/msgserver/msg_server_worker_payload.go
+++ b/x/emissions/keeper/msgserver/msg_server_worker_payload.go
@@ -142,8 +142,8 @@ func (ms msgServer) InsertWorkerPayload(ctx context.Context, msg *types.InsertWo
 				return nil, err
 			}
 			if !isInfererRegistered {
-				return nil, errorsmod.Wrapf(err,
-					"Error forecasted inferer address is not registered in this topic")
+				sdkCtx.Logger().Info("Skipping unregistered inferer", "inferer", el.Inferer)
+				continue
 			}
 
 			notAlreadySeen := !seenInferers[el.Inferer]
--- a/x/emissions/keeper/msgserver/msg_server_worker_payload.go
+++ b/x/emissions/keeper/msgserver/msg_server_worker_payload.go
@@ -136,16 +136,6 @@ func (ms msgServer) InsertWorkerPayload(ctx context.Context, msg *types.InsertWo
 		acceptedForecastElements := make([]*types.ForecastElement, 0)
 		seenInferers := make(map[string]bool)
 		for _, el := range forecast.ForecastElements {
-			// Check if the forecasted inferer is registered in the topic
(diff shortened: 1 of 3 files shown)


amimart (dismissed):

x/emissions/keeper/msgserver/msg_server_worker_payload.go
```diff
@@ -139,8 +139,8 @@ func (ms msgServer) InsertWorkerPayload(ctx context.Context, msg *types.InsertWo
 				return nil, err
 			}
 			if !isInfererRegistered {
-				return nil, errorsmod.Wrapf(err,
-					"Error forecasted inferer address is not registered in this topic")
+				sdkCtx.Logger().Info("Skipping unregistered inferer", "inferer", el.Inferer)
```
xmariachi: I think that, because we are validating separately the input, and this is a transaction (which means that it depends on the block it gets executed in), this would break the expectations from different forecasts. 

Example: 
* forecast is done with inferers on the active set.  
* meanwhile, one inferer in the active set unregisters itself in the middle of the submission window.
* so , some forecasts before that block within the window will find that inferer registered, and some will find it unregistered - so this inferer would be added in some forecasts and not in others, causing data inconsistency downstream.

So, because this check depends on changes of state within an eopch, I believe we should plain remove this registration check completely - and trust the forecast validation sending the right inferers.

@amimart @guilherme-brandao  wdyt?

xmariachi (changes requested):
Left a comment that needs revision @fernandofcampos @amimart @guilherme-brandao

x/emissions/keeper/msgserver/msg_server_worker_payload.go
amimart: From what I understand when we close the wsw we filter the forecast elements with the inferers of the active set, so if we do not make this check it'll be filtered later on anyway, is that right?

xmariachi: yep. 
Actually, reviewing the `CloseWorkerNonce`, that will limit the forecasts to the right inferers.
However, I think that checking if the inferer is registered should not be done here at tx time. It will be checked if it is in the active set (and thus if it is registered as well) at one point in time at `CloseWorkerNonce`.

amimart: Yes ok I'm fine with that, let's reduce forecasters fees 😊

amimart (dismissed):

xmariachi:
LGTM, please rebase @fernandofcampos

amimart (approved):

xmariachi (approved):

PR #778, opened 2025-03-21
fix spelling issues 

rnkrtt:
app/ante.go
keeper keeper - keeper `fixed duplicate`

app/app.yaml
`deleted space`


spooktheducks:
Incorporated in #914

PR #779, opened 2025-03-22
Fix max distant value in stake weighted calc

guilherme-brandao:
This PR improves the implementation of GetStakeWeightedLossMatrix to correctly identify the most distant value from the consensus using absolute differences. The function is used to calculate stake-weighted loss matrices and identify outliers in reported losses.

- Modified the logic in GetStakeWeightedLossMatrix to properly calculate absolute differences by taking the absolute value of the distance between each value and the consensus
- Updated tests to verify the correct behavior with values both above and below consensus

https://linear.app/alloralabs/issue/ENGN-3540/182-incorrect-maxdistant-value-causes-reputers-who-havent-submitted

### DIFF
--- a/x/emissions/module/rewards/rewards_internal.go
+++ b/x/emissions/module/rewards/rewards_internal.go
@@ -292,12 +292,9 @@ func GetStakeWeightedLossMatrix(
 				return nil, nil, err
 			}
 
-			absoluteDifference := distance
-			if distance.IsNegative() {
-				absoluteDifference, err = distance.Abs()
-				if err != nil {
-					return nil, nil, err
-				}
+			absoluteDifference, err := distance.Abs()
+			if err != nil {
+				return nil, nil, err
 			}
 
 			// Update the most distant value if this is the first valid distance
--- a/x/emissions/module/rewards/rewards_internal.go
+++ b/x/emissions/module/rewards/rewards_internal.go
@@ -299,7 +299,7 @@ func GetStakeWeightedLossMatrix(
 					return nil, nil, err
 				}
 			}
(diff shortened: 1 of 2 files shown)


x/emissions/module/rewards/rewards_internal.go
```diff
@@ -293,9 +291,21 @@ func GetStakeWeightedLossMatrix(
 			if err != nil {
 				return nil, nil, err
 			}
-			if distance.Gt(maxDistance) {
-				maxDistance = distance
+
+			absoluteDifference := distance
```
xmariachi: I guess we can use `Dec.Abs()`? :D

xmariachi (changes requested):
Left a suggestion, otherwise lgtm.

x/emissions/module/rewards/rewards_internal.go
```diff
@@ -293,9 +291,21 @@ func GetStakeWeightedLossMatrix(
...
+			if distance.IsNegative() {
+				absoluteDifference, err = distance.Abs()
+				if err != nil {
+					return nil, nil, err
+				}
+			}
```
amimart:
```suggestion
			absoluteDifference, err = distance.Abs()
			if err != nil {
				return nil, nil, err
			}
```

This could be simplified

guilherme-brandao: oh yeah, nice catch!

amimart (approved):

fernandofcampos (approved):

xmariachi (approved):

PR #781, opened 2025-03-24
Enable pebbledb ENGN-3576

fernandofcampos:

## Still Left Todo

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -15,6 +15,7 @@ builds:
     tags:
       - netgo
       - ledger
+      - pebbledb
 
     ldflags:
       - -X github.com/cosmos/cosmos-sdk/version.BuildTags=netgo,ledger
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 # v0.12.0
 
 ### Added
+* [#781](https://github.com/allora-network/allora-chain/pull/781/files) Enable pebbledb
 
 ### Changed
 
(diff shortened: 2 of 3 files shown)


amimart (commented):
Please add to changelog

cmd/allorad/cmd/root.go
```diff
@@ -106,6 +106,8 @@ func NewRootCmd() *cobra.Command {
 			cmtCfg := cmtcfg.DefaultConfig()
 			cmtCfg.Consensus.TimeoutCommit = 3 * time.Second
 			cmtCfg.LogLevel = "*:error,p2p:info,state:info" // better default logging
+			// overwrite the default db backend
+			cmtCfg.BaseConfig.DBBackend = "pebbledb"
```
xmariachi, line 110: Will setting this make usage of `pebbledb` the default one?
Aren't we keeping this optional? ie if we just merge it like it is, it will force everyone to use pebbledb. Cannot this be set via a flag only, and keeping the prev as default?

xmariachi: Will setting this make usage of `pebbledb` the default one?
Aren't we keeping this optional? ie if we just merge it like it is, it will force everyone to use pebbledb, won't it?
 Cannot this be set via a flag only, and keeping the prev as default?

xmariachi (commented):
A comment on defaults

cmd/allorad/cmd/root.go
amimart: This set `pebbledb` as default backend in the `config.toml` when initialising a new node, the `config.toml` can still be edited.

But I agree to keep the default for now and remove this

fernandofcampos: Done

amimart (approved):

xmariachi (approved):

PR #786, opened 2025-03-30
Remove unsupported formatters from golangci-lint config

Fallengirl:
Remove from enable tools that are not in the current list of linters or that are formatters:

- `gofmt` (formatter)
- `goimports` (formatter)
- `gosimple` (deprecated, missing)
- `typecheck` (missing)
- `tenv` (also not in the list of supported linters)

Without these changes, golangci-lint fails with errors. Locally tested per the [official migration guide](https://golangci-lint.run/product/migration-guide/).


spooktheducks:
Incorporated in #914

PR #794, opened 2025-04-04
Optimize topic weights fetch and update flow

guilherme-brandao:
- Remove repeated and unnecessary fetches for the module params and topic structure during the topic `GetCurrentTopicWeight` and `DripTopicFeeRevenue`

No ticket was created for this. This optimization opportunity was found during the execution of this ticket: 
https://linear.app/alloralabs/issue/ENGN-3608/test-topic-weight-changes

### DIFF
--- a/x/emissions/keeper/keeper.go
+++ b/x/emissions/keeper/keeper.go
@@ -3312,7 +3312,7 @@ func (k *Keeper) AddTopicFeeRevenue(ctx context.Context, topicId TopicId, amount
 
 // return the blocks per week
 // defined as the blocks per month divided by 4.345
-var weeksPerMonth = alloraMath.MustNewDecFromString("4.345")
+var weeksPerMonth = alloraMath.MustNewDecFromString("4.345") //nolint:gochecknoglobals // constant
 func (k *Keeper) CalculateBlocksPerWeek(ctx context.Context, blocksPerMonth uint64) (alloraMath.Dec, error) {
 	blocksPerMonthDec, err := alloraMath.NewDecFromUint64(blocksPerMonth)
 	if err != nil {
--- a/x/emissions/keeper/keeper.go
+++ b/x/emissions/keeper/keeper.go
@@ -3312,21 +3312,13 @@ func (k *Keeper) AddTopicFeeRevenue(ctx context.Context, topicId TopicId, amount
 
 // return the blocks per week
 // defined as the blocks per month divided by 4.345
-func (k *Keeper) calculateBlocksPerWeek(ctx context.Context) (alloraMath.Dec, error) {
-	moduleParams, err := k.GetParams(ctx)
-	if err != nil {
-		return alloraMath.Dec{}, errorsmod.Wrap(err, "error getting params")
-	}
-	blocksPerMonth, err := alloraMath.NewDecFromUint64(moduleParams.BlocksPerMonth)
+var weeksPerMonth = alloraMath.MustNewDecFromString("4.345")
+func (k *Keeper) CalculateBlocksPerWeek(ctx context.Context, blocksPerMonth uint64) (alloraMath.Dec, error) {
+	blocksPerMonthDec, err := alloraMath.NewDecFromUint64(blocksPerMonth)
 	if err != nil {
 		return alloraMath.Dec{}, errorsmod.Wrap(err, "error creating blocks per month")
 	}
-	// 4.345 weeks per month on average
-	weeksPerMonth, err := alloraMath.NewDecFromString("4.345")
-	if err != nil {
-		return alloraMath.Dec{}, errorsmod.Wrap(err, "error creating weeks per month")
-	}
-	blocksPerWeek, err := blocksPerMonth.Quo(weeksPerMonth)
+	blocksPerWeek, err := blocksPerMonthDec.Quo(weeksPerMonth)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


x/emissions/keeper/keeper.go
```diff
@@ -3312,21 +3312,13 @@ func (k *Keeper) AddTopicFeeRevenue(ctx context.Context, topicId TopicId, amount
 
 // return the blocks per week
 // defined as the blocks per month divided by 4.345
-func (k *Keeper) calculateBlocksPerWeek(ctx context.Context) (alloraMath.Dec, error) {
-	moduleParams, err := k.GetParams(ctx)
-	if err != nil {
-		return alloraMath.Dec{}, errorsmod.Wrap(err, "error getting params")
-	}
-	blocksPerMonth, err := alloraMath.NewDecFromUint64(moduleParams.BlocksPerMonth)
+var weeksPerMonth = alloraMath.MustNewDecFromString("4.345") //nolint:gochecknoglobals // constant
```
xmariachi: This var is too general, not specific to the keeper. It may also be used in other places so maybe there is a better, more general place to define it. Though I'm not sure where, we don't have a place for such constants (we probably should!)
So for now, pls move it to be defined at start of file after imports (go convention).

xmariachi (changes requested):
small change but otherwise lgtm

x/emissions/keeper/keeper.go
```diff
@@ -3361,49 +3353,41 @@ func (k *Keeper) SetLastDripBlock(ctx context.Context, topicId TopicId, block Bl
 // where C_{t,i} is the topic fee revenue
 // and N_{epochs,w} is the number of epochs per week
 // and this decay or drip happens each epoch
-func (k *Keeper) DripTopicFeeRevenue(ctx sdk.Context, topicId TopicId, block BlockHeight) error {
-	if err := types.ValidateTopicId(topicId); err != nil {
+func (k *Keeper) DripTopicFeeRevenue(ctx sdk.Context, topic types.Topic, blocksPerWeek alloraMath.Dec, block BlockHeight) error {
+	if err := types.ValidateTopicId(topic.Id); err != nil {
```
amimart, line 3342: Do we still need to validate the topic id? I feel that having already fetched the `types.Topic` should ensure it exists and is valid

x/emissions/keeper/keeper.go
```diff
@@ -3312,21 +3312,13 @@ func (k *Keeper) AddTopicFeeRevenue(ctx context.Context, topicId TopicId, amount
...
 // return the blocks per week
 // defined as the blocks per month divided by 4.345
-func (k *Keeper) calculateBlocksPerWeek(ctx context.Context) (alloraMath.Dec, error) {
-	moduleParams, err := k.GetParams(ctx)
-	if err != nil {
-		return alloraMath.Dec{}, errorsmod.Wrap(err, "error getting params")
-	}
-	blocksPerMonth, err := alloraMath.NewDecFromUint64(moduleParams.BlocksPerMonth)
+var weeksPerMonth = alloraMath.MustNewDecFromString("4.345") //nolint:gochecknoglobals // constant
+func (k *Keeper) CalculateBlocksPerWeek(ctx context.Context, blocksPerMonth uint64) (alloraMath.Dec, error) {
```
```suggestion
func CalculateBlocksPerWeek(ctx context.Context, blocksPerMonth uint64) (alloraMath.Dec, error) {
```

If we don't interact with the keeper maybe we can detach the func from it

amimart (commented):
Just small remarks I'll let up to you

x/emissions/keeper/keeper.go
guilherme-brandao: I already questioned these redundant validations, but the argument is that they are part of a defense-in-depth approach. I think we should revisit whether we want to keep this in the future.

x/emissions/keeper/keeper.go
guilherme-brandao: Moved to `math/utils`

x/emissions/keeper/keeper.go
guilherme-brandao: Moved to `math/utils` alongside the `CalculateBlocksPerWeek` func

xmariachi:
Needs fixing linter and compilation is failing in tests, otherwise it lgtm.
```
# github.com/allora-network/allora-chain/x/emissions/keeper_test [github.com/allora-network/allora-chain/x/emissions/keeper.test]
Error: x/emissions/keeper/keeper_test.go:4747:45: not enough arguments in call to k.DripTopicFeeRevenue
	have ("github.com/cosmos/cosmos-sdk/types".Context, uint64, int64)
	want ("github.com/cosmos/cosmos-sdk/types".Context, "github.com/allora-network/allora-chain/x/emissions/types".Topic, "github.com/allora-network/allora-chain/math".Dec, "github.com/allora-network/allora-chain/x/emissions/keeper".BlockHeight)
Error: x/emissions/keeper/keeper_test.go:4750:45: not enough arguments in call to k.DripTopicFeeRevenue
	have ("github.com/cosmos/cosmos-sdk/types".Context, uint64, int64)
```

amimart (dismissed):

xmariachi (changes requested):
Add to changelog, otherwise lgtm, good work

amimart (approved):

xmariachi (approved):

PR #796, opened 2025-04-08
Fix: rework nondet math operations

amimart:

Re-implement in a deterministic way some math func, here are the details on the chosen algorithms:

- `Exp`: Uses a Taylor polynomial with a lookup table for powers of `10^(1/n)` of size 300 which seems sufficient for 34 precision digits;
- `Ln`: Implements the Newton's method that uses exp 3 times;
- `Log10`: Use `d = ln(x) * 1/ln(10)`;
- `Pow`: Implement `apd`'s algorithm but using our `Ln` and `Exp` impls;

The `Exp10` func has been removed as not used.

The algorithms choices and implementations were made with the idea on improving performances, we are now much faster, see the benchmarks below. In terms of accuracy it is almost identical to previous implementations.

The perf improvements helped the network inferences calculation to drop from 250ms to 100ms.

Benchmarks before:
```
BenchmarkLn/input_1.2-12                    12574	     95334 ns/op
BenchmarkLog10/input_1.2-12              17979	     66241 ns/op
BenchmarkExp/input_1.2-12                  44658	     26917 ns/op
BenchmarkPow/input_1.2_1.2-12          10000	    103102 ns/op
```

Benchmarks after:
```
BenchmarkLn/input_1.2-12                    45258	     26759 ns/op
BenchmarkLog10/input_1.2-12              44979	     27504 ns/op
BenchmarkExp/input_1.2-12                  153902	      7737 ns/op
BenchmarkPow/input_1.2_1.2-12          36795	     32447 ns/op
```


math/dec_bench_test.go
```diff
@@ -1,30 +1,88 @@
...
+					b.Fatal(err)
+				}
+				b.StopTimer()
+			}
+		})
+	}
+}
+
+func BenchmarkPow(b *testing.B) {
+	for _, d := range benchExpData {
```
xmariachi: Here we are using expData for `pow`. Shouldn't we use different testing data for pow, even if for clarity? 

Also - I think the tests could be table driven including expecting values (eg here we are testing that there are no Fatal errors, but not the expected values themselves, right?) 
(also, is `Fatal` the right error here?)

xmariachi (commented):
Left a comment on testing. lgtm in general, good work!

math/dec_bench_test.go
amimart: I'll add another set of data for pow for clarity sake you're right.

The values are not checked here because these are benchmarks not tests, the fatal error is because it makes no sense to benchmark error cases. Some tests already exists in `dec_test.go` but they're a bit limited so I'll add some more complete ones

math/dec_test.go
```diff
@@ -180,7 +180,8 @@ func TestDec(t *testing.T) {
 	// See https://github.com/allora-network/allora-chain/pull/789
 	logDeterminism, err := alloraMath.Ln(alloraMath.MustNewDecFromString("1.6285091944505809264504560045920167"))
 	require.NoError(t, err)
-	require.True(t, alloraMath.MustNewDecFromString("0.4876649916811116824516548471782886").Equal(logDeterminism))
+	fmt.Println(logDeterminism.String())
```
guilherme-brandao: Remove `Println` ?

guilherme-brandao (dismissed):
Minor comment. Already approving! Nice work!

math/dec_test.go
amimart: Oh yeah good catch!

xmariachi (approved):

guilherme-brandao (approved):

PR #799, opened 2025-04-11
Correct Monthly PreviousPercentageRewardToStakedReputers Calculation

guilherme-brandao:

This PR introduces functionality to track and manage rewards distributed within the emissions module on a monthly basis. The primary goal is to fix inaccuracies and timing issues in the calculation of `PreviousPercentageRewardToStakedReputers`.

**Problem 1: Inaccurate Calculation Basis**
Previously, the percentage calculation was based only on the rewards distributed during the specific block/epoch where the calculation occurred. This didn't accurately reflect the true distribution.

**Problem 2: Conditional Update Timing**
The update to `PreviousPercentageRewardToStakedReputers` previously only happened if the designated month-end block (where `blockHeight % BlocksPerMonth == 0`) *also* coincided with active reward emissions for some topic in that same block via the `EmitRewards` function. If no topics were rewarded on that specific block, the percentage wasn't updated for the concluded month.

**Solution: Monthly Reward Tracking & Logic Relocation**

1.  **Persistent Monthly Counters:** Introduced `MonthlyReputerRewards` and `MonthlyTopicRewards` state entries to the keeper. These counters are now incremented whenever rewards are distributed throughout the month.
2.  **Accurate Calculation:** The calculation logic, now located directly in the `EndBlocker`, triggers reliably whenever `blockHeight % BlocksPerMonth == 0`. It retrieves the *total* accumulated monthly rewards (`GetMonthlyReputerRewards`, `GetMonthlyTopicRewards`) to compute an accurate percentage based on the *entire month's* activity.
3.  **Reliable Timing:** By moving the logic outside the conditional `EmitRewards` flow, the percentage update now correctly occurs at the end of every month period defined by `BlocksPerMonth`, regardless of whether rewards happen to be emitted in that specific end block.
4.  **State Reset:** The monthly counters are reset after the calculation prepares for the next month.

https://linear.app/alloralabs/issue/ENGN-3478/address-f-stakers-tokenomics-issues-found-by-research

## Still Left Todo

- Need to create migration to fill `MonthlyReputerRewards` and `MonthlyTopicRewards`


x/emissions/module/abci.go
```diff
@@ -100,5 +102,61 @@ func EndBlocker(ctx context.Context, am AppModule) error {
 			sdkCtx.Logger().Warn("Error deleting worker window blockheight", "error", err)
 		}
 	}
+
+	// If the current block is a multiple of the number of blocks per month,
+	// update the percentage of rewards going to staked reputers and reset monthly counters.
+	if uint64(blockHeight)%moduleParams.BlocksPerMonth == 0 {
```
xmariachi: I think this whole part should be refactored into a function that is called `defer`red at the start.
If an error happens before this, it wont get done.

Also making it into a separate function will make it easy to be unit-tested. We probably want this moved into ... the keeper module?

x/emissions/module/rewards/rewards.go
```diff
@@ -152,29 +149,23 @@ func EmitRewards(args EmitRewardsArgs) error {
...
+			return errors.Wrapf(err, "failed to add monthly reputer rewards")
+		}
+		// Add to monthly topic rewards
+		topicRewardInt, err := topicReward.SdkIntTrim()
 		if err != nil {
-			return errors.Wrapf(err, "failed to calculate percentage to staked reputers")
+			return errors.Wrapf(err, "failed to convert topic reward to int")
 		}
-		err = args.K.SetPreviousPercentageRewardToStakedReputers(args.Ctx, percentageToStakedReputers)
+		err = args.K.AddMonthlyTopicRewards(args.Ctx, topicRewardInt)
```
xmariachi: Small thing here, but wouldn't it be better to just store and add at the end? For efficiency (one write to the store).
  
Also makes me think the "partial modification" state that it will be left if the logic fails in some rewardableTopic, preventing others from being processed. 
1. Should those be a `continue`, and in that case, it shouldn't do any modification?
2. If it still should return an error, it that partial state change correct?

x/emissions/module/abci.go
xmariachi: Also consider the internal error handling within this new code. 
If some of those happen - the values wont be reset, and they will remain like this until next month. I'm not sure of the best behaviour - but let's define it :)

xmariachi (changes requested):

x/emissions/module/abci.go
guilherme-brandao: The other alternative would be to halt the chain and add a `panic` error.

guilherme-brandao: I’m not sure about moving it to the keeper. We should reduce the size of the keeper, it has a bunch of things. But I agree with putting it in a separate function and adding it in the `defer`.

x/emissions/module/rewards/rewards.go
guilherme-brandao: I think we need to start thinking about where a `panic(err)` is actually desirable — anything related to tokenomics is a strong candidate for that. 

Some anomalies can happen that I'm sure would be better to just not move forward if they happen.

CHANGELOG.md
```diff
@@ -65,6 +65,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### API Breaking Changes
+
+#### Removed
+
+#### Added 
+
+#### Changed
+
+# [Unreleased]
```
amimart:
```suggestion
# [Released]
```

There's the old `# [Released]` marker below to remove as well

x/emissions/keeper/genesis.go
```diff
@@ -1007,6 +1007,35 @@ func (k *Keeper) InitGenesis(ctx context.Context, data *types.GenesisState) erro
...
+			return errors.Wrap(err, "monthly reputer rewards validation failed")
+		}
+		if err := k.SetMonthlyReputerRewards(ctx, data.MonthlyReputerRewards); err != nil {
+			return errors.Wrap(err, "error setting monthlyReputerRewards")
+		}
+	} else {
+		if err := k.SetMonthlyReputerRewards(ctx, cosmosMath.ZeroInt()); err != nil {
+			return errors.Wrap(err, "error setting monthlyReputerRewards to zero int")
+		}
+	}
```
amimart:
```suggestion
	// MonthlyReputerRewards
	if err := types.ValidateSdkIntRepresentingMonetaryValue(data.MonthlyReputerRewards); err != nil {
		return errors.Wrap(err, "monthly reputer rewards validation failed")
	}
	if err := k.SetMonthlyReputerRewards(ctx, data.MonthlyReputerRewards); err != nil {
		return errors.Wrap(err, "error setting monthlyReputerRewards")
	}
```

This should be enough I guess, same simplification can apply for the `MonthlyTopicRewards`

x/emissions/module/abci.go
```diff
@@ -13,16 +14,28 @@ import (
...
 	sdkCtx.Logger().Debug("---------------- Emissions EndBlock -------------------", "blockHeight", blockHeight)
+
 	moduleParams, err := am.keeper.GetParams(sdkCtx)
 	if err != nil {
 		sdkCtx.Logger().Error("Error Getting module params", err)
 		return err
 	}
+
+	defer func() {
+		telemetry.ModuleMeasureSince(emissionstypes.ModuleName, time.Now(), telemetry.MetricKeyEndBlocker)
```
amimart: The `time.Now` in this line must be executed very first which here is called in the deferred func, and the `telemetry.ModuleMeasureSince` must be executed very last which is not the case either.

As defer are LIFOed we should keep this in a first defer:
```
defer telemetry.ModuleMeasureSince(emissionstypes.ModuleName, telemetry.Now(), telemetry.MetricKeyEndBlocker)
```

x/emissions/module/rewards/rewards.go
amimart: This is usually the way it is managed, and this doesn't require a panic but just returning the error in the end blocker instead of silencing it.

I'd tend to prefer dealing with chain halts than state inconsistencies, the latter could be a nightmare on the long term.

To mitigate this, a lot of errors can occur on unhealthy nodes with difficulties reading/writing to the database, in such cases the node should stop

amimart (changes requested):
As this is state breaking the `x/emissions` consensus version must be incremented and a migration registered

x/emissions/module/rewards/rewards.go
amimart: This is usually the way it is managed, and this doesn't require a panic but just returning the error in the end blocker instead of silencing it.

I'd tend to prefer dealing with chain halts than state inconsistencies, the latter could be a nightmare on the long term.

To mitigate this, a lot of errors can occur on unhealthy nodes with difficulties reading/writing to the database, in such cases the node should stop

x/emissions/proto/emissions/v10/genesis.proto
amimart, line 1: This can be added to `v9` instead of introducing a new version

x/emissions/module/abci.go
xmariachi: > The other alternative would be to halt the chain and add a panic error.

I mean that, eg what happens if the first Reset is OK but the second Reset fails? Is it ok to leave it in such inconsistent state (where the first one worked, the second one didn't)?
Or if the `GetMonthlyReputerRewards` fails - shouldn't the reset happen anyway?
(i don't think any of that merits a panic though)

> We should reduce the size of the keeper, it has a bunch of things. 

Yes I'm concerned about the size of the `keeper.go` file, but keeper-related functions need not go in the `keeper.go` file (eg `actor_penalties.go`, `topic_weight.go`, `topic_activation.go` contain keeper object functions but in separate files for better readability). 
So if it functionally makes sense to be an object method, I think it should be (regardless of how many object-based methods we have in the keeper).

x/emissions/module/rewards/rewards.go
xmariachi: +1 on reviewing the usage of panic and make it consistent through the codebase.

x/emissions/keeper/genesis.go
```diff
@@ -2578,6 +2595,18 @@ func (k *Keeper) ExportGenesis(ctx context.Context) (*types.GenesisState, error)
 		})
 	}
 
+	// Get Monthly Reputer Rewards
+	monthlyReputerRewards, err := k.GetMonthlyReputerRewards(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get monthly reputer rewards")
```
xmariachi, line 2597: Hey. I know it goes out of the scope of this PR but, following the recent comments: shouldn't all of these errors on the genesis be a `panic` instead of an error? (Checked a few modules out there, it is normally the case to `panic` on such genesis actions)
cc @amimart @spooktheducks

x/emissions/module/abci.go
```diff
@@ -100,5 +113,60 @@ func EndBlocker(ctx context.Context, am AppModule) error {
...
+			return errors.Wrapf(err, "Failed to convert total topic rewards to alloraMath.Dec")
+		}
+
+		percentageToStakedReputersDec, err = totalReputersDec.Quo(totalTopicDec)
+		if err != nil {
+			return errors.Wrapf(err, "Failed to calculate percentage to staked reputers")
+		}
+	}
+
+	err = am.keeper.SetPreviousPercentageRewardToStakedReputers(sdkCtx, percentageToStakedReputersDec)
```
xmariachi: Should we send an event for this specifically?

x/emissions/module/abci.go
```diff
@@ -100,5 +113,60 @@ func EndBlocker(ctx context.Context, am AppModule) error {
 			sdkCtx.Logger().Warn("Error deleting worker window blockheight", "error", err)
 		}
 	}
+
+	return nil
+}
+
+// handleMonthlyRewardsReset calculates the percentage of rewards that went to staked reputers
+// in the previous month, stores it, and resets the monthly counters.
+func handleMonthlyRewardsReset(sdkCtx sdk.Context, am AppModule) (err error) {
```
xmariachi: Not a strong opinion though, but I think this should be a keeper method (even if in a separate file or sth), in the sense that it feels totally keeper related. Curious what others think

x/mint/scripts/protocgen.sh
```diff
@@ -4,7 +4,7 @@ set -e
 
 echo "Generating gogo proto code"
 cd proto
-proto_dirs=$(find . -path -prune -o -name '*.proto' -print0 | xargs -0 -n1 dirname | sort | uniq)
+proto_dirs=$(find . -path -prune -o -name '*.proto' -print0 | xargs -0 -n1 dirname | sort -V | uniq)
```
xmariachi, line 7: 👍

x/emissions/module/abci.go
guilherme-brandao: We should! Adding...

x/emissions/keeper/genesis.go
amimart: If an error is returned here it'll panic in the module's ABCI related method [here](https://github.com/allora-network/allora-chain/blob/5c90507f2faed841b39447649175ba8b229f20e4/x/emissions/module/module.go#L179), so in the end we have the same behaviour.

I personally prefer having a single panic call and a method containing the logic returning errors like we do, I also wouldn't mind to panic directly :)

x/emissions/module/abci.go
```diff
@@ -13,16 +13,28 @@ import (
...
 	sdkCtx.Logger().Debug("---------------- Emissions EndBlock -------------------", "blockHeight", blockHeight)
+
 	moduleParams, err := am.keeper.GetParams(sdkCtx)
 	if err != nil {
 		sdkCtx.Logger().Error("Error Getting module params", err)
 		return err
 	}
+
+	defer func() {
+		if err == nil && uint64(blockHeight)%moduleParams.BlocksPerMonth == 0 {
```
amimart: If the intention of using defer is to execute that even if there's an error I think the err nil check should be removed

x/emissions/module/abci.go
amimart: Could be put in the keeper yeah but I don't mind having this logic here.

This only thing I'd consider important to put in the keeper though is a `ResetMonthlyRewards` func ensuring that `SetPreviousPercentageRewardToStakedReputers`, `ResetMonthlyReputerRewards` and `ResetMonthlyTopicRewards` are systematically called together: The keeper being the guardian of state integrity, I think partial state update in its public api should be avoid

x/emissions/keeper/genesis.go
xmariachi: gotcha, then makes sense to keep as it is, at least for now. In general I think we should be abiding by common standards though.

x/emissions/keeper/keeper.go
```diff
@@ -4597,3 +4603,77 @@ func (k Keeper) SetLatestForecasterWeight(ctx context.Context, topicId TopicId,
 	}
 	return k.latestForecasterWeights.Set(ctx, collections.Join(topicId, worker), weight)
 }
+
+// GetMonthlyReputerRewards retrieves the total reward for the month going to reputers.
+func (k *Keeper) GetMonthlyReputerRewards(ctx context.Context) (cosmosMath.Int, error) {
```
amimart, line 4608: In all their usage, reputer and topic rewards are used together, and updated together, what do you think about exposing only one set of funcs (e.g. `AddMonthlyRewards(ctx, reputer, topic)`) to manage both of them? And if in their usage their should be either reseted or added, I think the setter shouldn't exists or at least be private

This would ensure by design they are updated together, as they should.

x/emissions/module/abci.go
amimart: Could be put in the keeper yeah but I don't mind having this logic here.

This only thing I'd consider important to put in the keeper though is a `ResetMonthlyRewards` func ensuring that `SetPreviousPercentageRewardToStakedReputers`, `ResetMonthlyReputerRewards` and `ResetMonthlyTopicRewards` are systematically called together: The keeper being the guardian of state integrity, I think partial state update in its public api should be avoid

amimart (dismissed):
LGTM

xmariachi (approved):
lgtm.
A reminder that having a migration and new stores within the PR, to double-check the checklists for upgrading and adding stores :) 
Approving preemptively

amimart (approved):

PR #800, opened 2025-04-11
Ensure listening coefficients never drops to 0

fernandofcampos:

## Still Left Todo

### DIFF
--- a/math/dec.go
+++ b/math/dec.go
@@ -53,13 +53,12 @@ var (
 var (
 	logOfEbase2 = MustNewDecFromString("1.442695040888963407359924681001892137")
 	// 0.5 with sufficient precision digit to allow right shifting keeping dec128 precision
-	log2B      apd.Decimal
-	oneBigInt  = apd.NewBigInt(1)
-	tenBigInt  = apd.NewBigInt(10)
-	zeroDec    = apd.New(0, 0)
-	oneDec     = apd.New(1, 0)
-	twoDec     = apd.New(2, 0)
-	epsilonDec = apd.New(1, -18) // Represents 1e-18
+	log2B     apd.Decimal
+	oneBigInt = apd.NewBigInt(1)
+	tenBigInt = apd.NewBigInt(10)
+	zeroDec   = apd.New(0, 0)
+	oneDec    = apd.New(1, 0)
+	twoDec    = apd.New(2, 0)
 )
 
 func init() {
@@ -69,13 +68,6 @@ func init() {
 	}
 	enforceDecimalPrecision(b, dec128Context.Precision+2)
 	log2B = *b
-}
-
-// EpsilonDec represents a small positive value (1e-18), used to prevent division by zero or issues with zero values.
-func EpsilonDec() Dec {
-	var d apd.Decimal
-	d.Set(epsilonDec)
-	return Dec{dec: d, isNaN: false}
 }
 
 // The number 0 encoded as Dec
--- a/math/dec.go
+++ b/math/dec.go
@@ -53,12 +53,13 @@ var (
 var (
 	logOfEbase2 = MustNewDecFromString("1.442695040888963407359924681001892137")
 	// 0.5 with sufficient precision digit to allow right shifting keeping dec128 precision
-	log2B     apd.Decimal
-	oneBigInt = apd.NewBigInt(1)
-	tenBigInt = apd.NewBigInt(10)
-	zeroDec   = apd.New(0, 0)
-	oneDec    = apd.New(1, 0)
-	twoDec    = apd.New(2, 0)
+	log2B      apd.Decimal
+	oneBigInt  = apd.NewBigInt(1)
+	tenBigInt  = apd.NewBigInt(10)
+	zeroDec    = apd.New(0, 0)
+	oneDec     = apd.New(1, 0)
+	twoDec     = apd.New(2, 0)
+	epsilonDec = apd.New(1, -18) // Represents 1e-18
 )
 
 func init() {
@@ -68,6 +69,13 @@ func init() {
 	}
 	enforceDecimalPrecision(b, dec128Context.Precision+2)
 	log2B = *b
+}
+
+// EpsilonDec represents a small positive value (1e-18), used to prevent division by zero or issues with zero values.
+func EpsilonDec() Dec {
+	var d apd.Decimal
+	d.Set(epsilonDec)
+	return Dec{dec: d, isNaN: false}
 }
 
 // The number 0 encoded as Dec
--- a/x/emissions/module/rewards/rewards_internal.go
+++ b/x/emissions/module/rewards/rewards_internal.go
@@ -10,6 +10,11 @@ import (
 	"github.com/allora-network/allora-chain/x/emissions/types"
 )
 
+const epsilon = "0.0000001"
+
+// EpsilonDec represents a small positive value (1e-18), used to prevent division by zero or issues with zero values.
+var EpsilonDec = alloraMath.MustNewDecFromString(epsilon)
+
 // flatten converts a double slice of alloraMath.Dec to a single slice of alloraMath.Dec
 func flatten(arr [][]alloraMath.Dec) []alloraMath.Dec {
 	var flat []alloraMath.Dec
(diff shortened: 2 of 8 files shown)


x/emissions/types/errors.go
```diff
@@ -97,4 +97,5 @@ var (
 	ErrMaxWhitelistInputArrayLengthExceeded      = errors.Register(ModuleName, 92, "max whitelist input array length exceeded")
 	ErrTopicAlreadyActive                        = errors.Register(ModuleName, 93, "topic already active")
 	ErrTopicCannotBeActivated                    = errors.Register(ModuleName, 94, "topic cannot be activated")
+	ErrDivisionByZero                            = errors.Register(ModuleName, 95, "division by zero")
```
xmariachi: There is a `ErrFractionDivideByZero` and `ErrNumberRatioDivideByZero`. 
We probably want to either use the Fraction if that makes sense, or otherwise create a personalised 
 `ErrCoefficientDivisionByZero`

math/dec.go
```diff
@@ -53,12 +53,13 @@ var (
...
-	zeroDec   = apd.New(0, 0)
-	oneDec    = apd.New(1, 0)
-	twoDec    = apd.New(2, 0)
+	log2B      apd.Decimal
+	oneBigInt  = apd.NewBigInt(1)
+	tenBigInt  = apd.NewBigInt(10)
+	zeroDec    = apd.New(0, 0)
+	oneDec     = apd.New(1, 0)
+	twoDec     = apd.New(2, 0)
+	epsilonDec = apd.New(1, -18) // Represents 1e-18
```
xmariachi: Is this usage of epsilon correct, or should we use topic's epsilon? Moving this to convo with research on ticket. I don't think an epsilon should be enforced at the type level - because different usages will need different epsilons.

xmariachi (commented):
Some comments (a full review on the most complicated parts is pending)

math/dec.go
amimart: I think this should be declared under the `rewards` package as it is where it's used

x/emissions/module/rewards/rewards_internal.go
```diff
@@ -225,89 +226,190 @@ func GetStakeWeightedLoss(reputersStakes, reputersReportedLosses []alloraMath.De
...
+	for _, row := range losses {
+		if len(row) != numCols {
+			return nil, nil, fmt.Errorf("all loss rows must have same length")
+		}
+	}
+
+	// First check if all stakes are zero or epsilon
+	allZeroOrEpsilon := true
+	for _, stake := range stakes {
+		if !stake.IsZero() && !stake.Equal(alloraMath.EpsilonDec()) {
```
amimart: It'd be better to have a global var with the epsilon dec instead of allocating a new one each iterations

x/emissions/module/rewards/rewards_internal.go
```diff
@@ -528,6 +630,29 @@ func GetAllReputersOutput(
...
+		if err != nil {
+			return nil, nil, errors.Wrap(err, "GetAllReputersOutput, err summing new coefficients")
+		}
+
+		// Avoid division by zero if sum is zero (shouldn't happen with EpsilonDec clamp, but safety first)
+		if coeffSum.IsZero() || coeffSum.Lt(alloraMath.EpsilonDec()) {
+			// If the sum is effectively zero even after epsilon clamp, something is wrong. Maybe return initial state or error.
+			// For now, let's return an error, as this state indicates a potential deeper issue.
+			return nil, nil, errors.Wrap(types.ErrDivisionByZero, "GetAllReputersOutput, coefficient sum effectively zero during normalization")
+		}
```
amimart:
```suggestion
```

This seems safe to remove IMHO :)

x/emissions/module/rewards/rewards_internal.go
```diff
@@ -528,6 +630,29 @@ func GetAllReputersOutput(
...
+
+		normalizedCoefficients := make([]alloraMath.Dec, len(newCoefficients))
+		for j := range newCoefficients {
+			normalizedCoefficients[j], err = newCoefficients[j].Quo(coeffSum)
+			if err != nil {
+				return nil, nil, errors.Wrapf(err, "GetAllReputersOutput, error normalizing coefficient %d", j)
+			}
+		}
+		// Update newCoefficients to use the normalized values for subsequent checks
+		newCoefficients = normalizedCoefficients
```
amimart:
```suggestion
		for j := range newCoefficients {
			newCoefficients[j], err = newCoefficients[j].Quo(coeffSum)
			if err != nil {
				return nil, nil, errors.Wrapf(err, "GetAllReputersOutput, error normalizing coefficient %d", j)
			}
		}
```

I guess we can reuse the same array instead of allocating a new one

x/emissions/module/rewards/rewards_internal.go
```diff
@@ -594,6 +719,23 @@ func GetAllReputersOutput(
...
+	if !finalCoeffSum.IsZero() && !finalCoeffSum.Equal(alloraMath.OneDec()) {
+		normalizedCoefficients := make([]alloraMath.Dec, len(coefficients))
+		for j := range coefficients {
+			normalizedCoefficients[j], err = coefficients[j].Quo(finalCoeffSum)
+			if err != nil {
+				return nil, nil, errors.Wrapf(err, "GetAllReputersOutput, error normalizing final coefficient %d", j)
+			}
+		}
+		coefficients = normalizedCoefficients
+	}
```
amimart:
```suggestion
	if !finalCoeffSum.Equal(alloraMath.OneDec()) {
		for j := range coefficients {
			coefficients[j], err = coefficients[j].Quo(finalCoeffSum)
			if err != nil {
				return nil, nil, errors.Wrapf(err, "GetAllReputersOutput, error normalizing final coefficient %d", j)
			}
		}
	}
```

We can reuse the `coefficients` array, and I think the !0 check is not needed

amimart (commented):
Thanks, I go some first remarks here, didn't check the whole PR yet

x/emissions/keeper/actor_utils/losses.go
```diff
@@ -295,7 +295,7 @@ func CloseReputerNonce(
 	}
 	types.EmitNewReputerLastCommitSetEvent(ctx, topic.Id, blockHeight, &nonce)
 
-	return nil
+	return err
```
xmariachi: probs from a rebase mistake?

amimart: There's 3 commits that should not be there, it should be fixed by running:
```bash
git rebase --onto a08bf877b1b9028ebcfa4b3ea7b9409020af1a17 363d6f982bafc00879cd6886dc83e845a5f6331f HEAD
```

x/emissions/module/rewards/rewards_internal.go
```diff
@@ -9,6 +10,11 @@ import (
 	"github.com/allora-network/allora-chain/x/emissions/types"
 )
 
+const epsilon = "1e-18"
+
+// EpsilonDec represents a small positive value (1e-18), used to prevent division by zero or issues with zero values.
+var EpsilonDec = alloraMath.MustNewDecFromString(epsilon)
```
guilherme-brandao: We already have this value in emissions params, it's called `EpsilonSafeDiv`, please reuse it.

x/emissions/module/rewards/rewards_test.go
```diff
@@ -956,7 +1001,46 @@ func (s *RewardsTestSuite) TestFixingTaskRewardAlphaDoesNotChangePerformanceImpo
...
+
+	// Fund the second topic with the same significant initial stake
+	s.MintTokensToAddress(s.addrs[reputerIndexes[0]], initialStake)
+	fundTopicMessage = types.FundTopicRequest{
+		Sender:  s.addrsStr[reputerIndexes[0]],
+		TopicId: topicId1,
+		Amount:  initialStake,
+	}
+	_, err = s.msgServer.FundTopic(s.ctx, &fundTopicMessage)
+	require.NoError(err)
```
guilherme-brandao: It seems like you're staking and topic funding both inside and outside the `setUpTopic` function. Is that intended? If not, please review other cases

x/emissions/module/rewards/scores_test.go
```diff
@@ -174,7 +174,8 @@ func (s *RewardsTestSuite) TestGetReputersScoresFromCsv() {
 		epoch301Get("reputer_score_4"),
 	}
 	for i, reputerScore := range scores {
-		testutil.InEpsilon5(s.T(), reputerScore.Score, expectedScores[i].String())
+		// Use a more lenient epsilon value (1% instead of 0.001%)
+		testutil.InEpsilon2(s.T(), reputerScore.Score, expectedScores[i].String())
```
guilherme-brandao: Are these changes compromising reputer scoring accuracy? If yes, has it already been double-checked with the research team to see if it's ok?

guilherme-brandao (changes requested):

x/emissions/keeper/actor_utils/losses.go
fernandofcampos: Done

x/emissions/module/rewards/rewards_test.go
fernandofcampos: I didn't realize setUpTopic was already doing that. I cleaned up the tests.

x/emissions/module/rewards/scores_test.go
fernandofcampos: Yes, because of the normalization (check at x/emissions/module/rewards/scores.go). I will double check with research.

x/emissions/module/rewards/rewards_internal.go
fernandofcampos: The problem with using that is that it implies reading from the keeper when we need it. I will see if I can optimize it.

x/emissions/module/rewards/rewards_test.go
fernandofcampos: Done

x/emissions/module/rewards/rewards_internal.go
```diff
@@ -225,89 +231,190 @@ func GetStakeWeightedLoss(reputersStakes, reputersReportedLosses []alloraMath.De
...
+	allZeroOrEpsilon := true
+	for _, stake := range stakes {
+		if !stake.IsZero() && !stake.Equal(EpsilonDec) {
+			allZeroOrEpsilon = false
+			break
+		}
+	}
+
+	// If all stakes are zero, use epsilon values instead to avoid division by zero
+	if allZeroOrEpsilon {
```
fstecker: Isn't the whole `allZeroOrEpsilon` thing redundant? If the only thing that happens in there is also inside the `if allZero`?

x/emissions/module/rewards/rewards_internal.go
```diff
@@ -513,7 +620,7 @@ func GetAllReputersOutput(
 			}
 			max, err := alloraMath.Max(
 				coefficientsPlusLearningRateTimesGradient,
-				alloraMath.ZeroDec(),
+				EpsilonDec,
```
fstecker: This wouldn't allow any of the listening coefficients to go below epsilon? I think that should still be allowed, we only need to check that they don't all become zero (or equivalently, their sum isn't zero). I think we already do this elsewhere, so this shouldn't be necessary?

x/emissions/module/rewards/scores.go
```diff
@@ -106,6 +106,25 @@ func GenerateReputerScores(
 			}
 			reputerListeningCoefficients[i] = cappedCoefficient
 		}
+
+		// Only normalize coefficients when there are multiple reputers
+		// When there's only one reputer, normalization is unnecessary and would force the value to 1.0
+		if len(reputerListeningCoefficients) > 1 {
```
fstecker: I don't think that's necessary. If there's only one reputer, its listening coefficient doesn't matter anyway (the losses will be whatever this reputer says, what else should it be).

x/emissions/module/rewards/rewards_internal.go
```diff
@@ -528,6 +635,19 @@ func GetAllReputersOutput(
...
+			return nil, nil, errors.Wrap(err, "GetAllReputersOutput, err summing new coefficients")
+		}
+
+		for j := range newCoefficients {
+			newCoefficients[j], err = newCoefficients[j].Quo(coeffSum)
+			if err != nil {
+				return nil, nil, errors.Wrapf(err, "GetAllReputersOutput, error normalizing coefficient %d", j)
+			}
+		}
+
```
fstecker, line 638: I'm having second thoughts about the normalization, and now think it might be better to normalize so that the max listening coefficient is 1, instead of normalizing so the sum is 1. That would make the `listenedStakeFraction` as high as possible and the `if listenedStakeFraction.Lt(params.MinStakeFraction)` branch less likely to be taken. Which might improve convergence. Maybe. I'll think about it some more.

fstecker (changes requested):
As we discussed today, it would be great if you could wait with this PR until we have some potential problems in the design figured out.

PR #803, opened 2025-04-14
Add reset/cleanup functions to defer in close nonce

guilherme-brandao:
This PR enhances the robustness of the CloseReputerNonce function by implementing a defer block to ensure critical cleanup operations (fulfilling the nonce, resetting active reputers, resetting submissions) are always executed, even if errors occur during the processing of reputer bundles.

The primary goal is to prevent the system from getting stuck in an inconsistent state if CloseReputerNonce fails mid-execution. By guaranteeing cleanup actions via defer, we ensure the nonce is properly fulfilled and associated state (active reputers, submissions) is reset, improving the reliability of the nonce management and reward distribution process.

https://linear.app/alloralabs/issue/ENGN-3672/required-functions-not-being-called-not-deferd

## Still Left Todo
- PR Review 
- Check if more tests need to be done

### DIFF
--- a/x/emissions/keeper/actor_utils/losses.go
+++ b/x/emissions/keeper/actor_utils/losses.go
@@ -23,8 +23,11 @@ func CloseReputerNonce(
 	k *keeper.Keeper,
 	ctx sdk.Context,
 	topic types.Topic,
-	nonce types.Nonce) error {
-	/// All filters should be done in order of increasing computational complexity
+	nonce types.Nonce,
+) (err error) {
+	blockHeight := ctx.BlockHeight()
+
+	// All filters should be done in order of increasing computational complexity
 	// Check if the worker nonce is unfulfilled
 	workerNonceUnfulfilled, err := k.IsWorkerNonceUnfulfilled(ctx, topic.Id, &nonce)
 	if err != nil {
@@ -53,10 +56,50 @@ func CloseReputerNonce(
 		)
 	}
 	// Check if the window time has passed
-	blockHeight := ctx.BlockHeight()
 	if blockHeight < nonce.BlockHeight+topic.GroundTruthLag {
 		return types.ErrReputerNonceWindowNotAvailable
 	}
+
+	defer func() {
+		if err != nil {
+			ctx.Logger().Error(
+				"Error occurred before finalization in CloseReputerNonce, attempting cleanup anyway",
+				"topicId", topic.Id,
+				"nonce", nonce,
+				"error", err,
+			)
+		}
+
+		_, fulfillErr := k.FulfillReputerNonce(ctx, topic.Id, &nonce)
+		if fulfillErr != nil {
+			ctx.Logger().Error(
+				"Error fulfilling reputer nonce during deferred cleanup",
+				"topicId", topic.Id,
+				"nonce", nonce,
+				"error", fulfillErr,
+			)
+		}
+
+		resetActiveErr := k.ResetActiveReputersForTopic(ctx, topic.Id)
+		if resetActiveErr != nil {
+			ctx.Logger().Error(
+				"Error resetting active reputers during deferred cleanup",
+				"topicId", topic.Id,
+				"error", resetActiveErr,
+			)
+		}
+
+		resetSubmissionsErr := k.ResetReputersIndividualSubmissionsForTopic(ctx, topic.Id)
+		if resetSubmissionsErr != nil {
+			ctx.Logger().Error(
+				"Error resetting reputer individual submissions during deferred cleanup",
+				"topicId", topic.Id,
+				"error", resetSubmissionsErr,
+			)
+		}
+
+		ctx.Logger().Info("Closed reputer nonce", "topicId", topic.Id, "nonce", nonce)
+	}()
 
 	params, err := k.GetParams(ctx)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


x/emissions/keeper/actor_utils/losses.go
```diff
@@ -217,34 +264,31 @@ func CloseReputerNonce(
...
-
-	err = k.ResetReputersIndividualSubmissionsForTopic(ctx, topic.Id)
-	if err != nil {
-		return err
-	}
-
 	types.EmitNewReputerLastCommitSetEvent(ctx, topic.Id, blockHeight, &nonce)
-	ctx.Logger().Info("Closed reputer nonce", "topicId", topic.Id, "nonce", nonce)
-	return nil
+
```
xmariachi, line 297: I would keep the Closed reputer nonce, even if at DEBUG level.

x/emissions/keeper/actor_utils/losses.go
```diff
@@ -274,6 +317,9 @@ func FilterUnacceptedWorkersFromReputerValueBundle(
 	// Get the accepted forecasters of the associated worker response payload
 	forecasts, err := k.GetForecastsAtBlock(ctx, topicId, reputerRequestNonce.ReputerNonce.BlockHeight)
 	if err != nil {
+		if errors.Is(err, collections.ErrNotFound) {
+			return nil, errorsmod.Wrapf(sdkerrors.ErrNotFound, "no forecasts found at block height %d for topic %d", reputerRequestNonce.ReputerNonce.BlockHeight, topicId)
```
xmariachi: if there are no forecasts, will it exit with error? Is this ok?

x/emissions/keeper/actor_utils/losses.go
guilherme-brandao: I sent it to the end of the `defer` function

x/emissions/keeper/actor_utils/losses.go
guilherme-brandao: `GetForecastsAtBlock` returns an empty array when the error is `ErrNotFound`, so this won’t trigger. Removing…

x/emissions/keeper/actor_utils/losses.go
```diff
@@ -92,7 +138,8 @@ func CloseReputerNonce(
 		// if they're left with no valid losses.
 		filteredBundle, err := FilterUnacceptedWorkersFromReputerValueBundle(k, ctx, topic.Id, *bundle.ValueBundle.ReputerRequestNonce, &bundle)
 		if err != nil {
-			continue
+			ctx.Logger().Warn("Could not filter bundle for reputer, skipping", "reputer", bundle.ValueBundle.Reputer, "topicId", topic.Id, "error", err)
+			continue // Skip this reputer
 		}
 
 		/// Filtering done now, now write what we must for inclusion
```
xmariachi, line 145: We should check below this loop, that there are enough remaining reputer losses (at least 1), otherwise we can cut it right there, can't we?

xmariachi:
Please do similarly for the CloseWorkerNonce counterparts: `ResetActiveWorkersForTopic` , `ResetWorkersIndividualSubmissionsForTopic`, `FulfillWorkerNonce`

xmariachi:
Additionally. 
I think we should move these into its own function and take them up to the point where we calculate the (topic, block) activity - so when we detect this, we can call defer these operations that we want to happen in any case.

For example, in [this function](https://github.com/allora-network/allora-chain/blob/0b3596e82de6ef5c0f14520be8ccaa9ee3cc7843/x/emissions/module/rewards/topic_rewards.go#L50) we detect that and ultimately we `CloseReputerNonce`. However, if there is any error in between, `CloseReputerNonce` won't be called and thus the cleaning won't happen.

So I believe we should move both function cleanings (worker, reputer) closest to the point where the (topic, block) is detected (and thus we know we need to ultimately clean).

x/emissions/keeper/actor_utils/losses.go
guilherme-brandao: make sense! added the check

x/emissions/keeper/actor_utils/losses.go
```diff
@@ -217,34 +268,31 @@ func CloseReputerNonce(
...
-	err = k.ResetReputersIndividualSubmissionsForTopic(ctx, topic.Id)
-	if err != nil {
-		return err
-	}
-
 	types.EmitNewReputerLastCommitSetEvent(ctx, topic.Id, blockHeight, &nonce)
-	ctx.Logger().Info("Closed reputer nonce", "topicId", topic.Id, "nonce", nonce)
-	return nil
+
+	return err
```
xmariachi: why returning err here? This should still return nil, right?

xmariachi (commented):
an additional comment, lgtm otherwise

x/emissions/keeper/actor_utils/losses.go
guilherme-brandao: right!

guilherme-brandao: It will return the error if it exists, but the function that is calling it will just log the error

xmariachi (approved):
lgtm!

amimart (approved):
LGTM

zale144 (approved):
lgtm

PR #807, opened 2025-04-16
Add v0.11.0 upgrade

amimart:

Increment the `x/emissions` module consensus version as changes in this release will be consensus breaking. An empty migration for consensus version from 9 to 10 has been added.

The `v0.11.0` upgrade handler has also been added.

The upgrade has been testnet locally.


x/emissions/module/module.go
```diff
@@ -41,7 +42,7 @@ var (
 )
 
 // ConsensusVersion defines the current module consensus version.
-const ConsensusVersion = 9
+const ConsensusVersion = 10
```
xmariachi, line 45: I understand that if we also RegisterTypes and RegisterInterfaces (some lines below), it will complain about duplicated types? 
However - in that case we better add a comment explaining that v9 need not be added to the lists (it would be pretty tempting in next migrations to just add it)

amimart: The types and interfaces to register are linked to the API version not the consensus version, so next time we increment the API version we'll still need to register `v9` types

xmariachi (dismissed):

x/emissions/module/module.go
```diff
@@ -137,6 +138,11 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
 	}); err != nil {
 		panic(fmt.Sprintf("failed to migrate x/%s from version 8 to 9: %v", types.ModuleName, err))
 	}
+	if err := cfg.RegisterMigration(types.ModuleName, 9, func(ctx sdk.Context) error {
```
xmariachi, line 141: a comment: what would it happen if we just do not exercise this migration here at all? Since the `MigrateStore` does not have any effect.

xmariachi (approved):

x/emissions/module/module.go
amimart: If you don't register a migration the nodes would crash after restarting with the new binary with the following log:
```
3:21PM ERR error in proxyAppConn.FinalizeBlock err="no migration found for module emissions from version 9 to version 10: not found" module=consensus
```

guilherme-brandao (approved):

PR #812, opened 2025-04-29
fix: validate reputer bundle contents against network inference

zale144:

Adding inferer address validation to insert reputer payload handler

https://linear.app/alloralabs/issue/ENGN-3677/validate-reputer-bundle-contents-against-network-inference

## Still Left Todo


test/integration/worker_inference_and_forecast_test.go
```diff
@@ -149,48 +150,22 @@ func InsertReputerBundle(m testCommon.TestConfig, topic *types.Topic, BlockHeigh
...
-		// Just as valid:
-		// OneOutInfererValues:    []*types.WithheldWorkerAttributedValue{},
-		// OneOutForecasterValues: []*types.WithheldWorkerAttributedValue{},
-		OneInForecasterValues: []*types.InputWorkerAttributedValue{
-			{
-				Worker: workerAddr,
-				Value:  alloraMath.MustNewBoundedExp40Dec(alloraMath.NewDecFromInt64(100)),
-			},
-		},
+		ForecasterValues:              nil,
```
xmariachi, line 153: Here, the forecast has been removed. Why?

x/emissions/keeper/actor_utils/losses.go
```diff
@@ -132,18 +121,7 @@ func CloseReputerNonce(
...
-		// if they're left with no valid losses.
-		filteredBundle, err := FilterUnacceptedWorkersFromReputerValueBundle(k, ctx, topic.Id, *bundle.ValueBundle.ReputerRequestNonce, &bundle)
-		if err != nil {
-			ctx.Logger().Warn("Could not filter bundle for reputer, skipping", "reputer", bundle.ValueBundle.Reputer, "topicId", topic.Id, "error", err)
-			continue // Skip this reputer
-		}
-
-		/// Filtering done now, now write what we must for inclusion
-		lossBundlesByReputer = append(lossBundlesByReputer, filteredBundle)
+		lossBundlesByReputer = append(lossBundlesByReputer, &bundle)
```
xmariachi, line 124: why are we removing the filtering here - is it because this would not be needed anymore given the new input validations?
Note that this also checks duplicates.

xmariachi:
I'm reluctant to merge this before running the simulator and the fuzzer on this branch, which would test if there is any additional change to do, on the chain or on the offchain side, to make sure this won't create disruption.

x/emissions/keeper/actor_utils/losses.go
amimart: About duplicates this is already checked at submission right? A worker cannot send two payloads in the same window

x/emissions/keeper/msgserver/msg_server_reputer_payload.go
```diff
@@ -103,3 +111,96 @@ func (ms msgServer) InsertReputerPayload(ctx context.Context, msg *types.InsertR
...
+	networkInferences, err := ms.k.GetNetworkInferences(ctx, topicId, blockHeight)
+	if err != nil {
+		return errorsmod.Wrapf(err, "error getting inferences")
+	}
+
+	if err := validateWorkerValues(
+		networkInferences.InfererValues,
+		valueBundle.InfererValues,
+	); err != nil {
+		return fmt.Errorf("inferer values mismatch: %w", err)
```
amimart, line 125:
```suggestion
		return errorsmod.Wrapf(sdkerrors.ErrInvalidRequest, "inferer values mismatch: %w", err)
```

As the error returned here is the one that'll appear in the tx result it should be registered with a code. I propose to use the generic `sdkerrors.ErrInvalidRequest` for that

amimart (commented):
Looks good to me, I only have a remark on error management.

I agree with @xmariachi that we need to ensure the current reputer implementation will pass these validations, not sure the simulator/fuzzer are enough to check that

zale144:
@xmariachi @amimart would the payload from the actual reputer implementation be different than the payloads used in unit/integration tests here?

test/integration/worker_inference_and_forecast_test.go
zale144: I think on the first cycle there wouldn't be any forecasts, otherwise, the validation would break

x/emissions/keeper/actor_utils/losses.go
zale144: the point of input validation is so we can remove filtering in the end blocker, right?

x/emissions/keeper/msgserver/msg_server_reputer_payload.go
zale144: makes sense, I'll do it above at the call site

PR #815, opened 2025-05-02
Refund ecosystem account when invalid topic to reward

guilherme-brandao:
This PR modifies the rewards distribution logic to handle cases where a topic is eligible for rewards based on its weight but cannot distribute them due to lack of active participation (e.g., no reward nonce submitted or errors during reward processing).

Previously, these rewards might have remained undistributed in the rewards module. Now, any rewards allocated to a topic that cannot be processed are transferred to the `EcosystemModuleName` account.

**Changes:**

-   Added `keeper.MoveCoinsFromAlloraRewardsToEcosystem` function to facilitate the transfer.
-   Updated `rewards.EmitRewards` to call the new keeper function when a topic's rewards cannot be processed.
-   Added the test case `TestNoActiveParticipantsNoRewardsForTopic` to verify that rewards for topics without active participation are correctly moved to the ecosystem module account.

https://linear.app/alloralabs/issue/ENGN-3525/189-185-apy-over-12percent

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+* [#815](https://github.com/allora-network/allora-chain/pull/815) Refund ecosystem account when invalid topic to reward 
 
 ### Deprecated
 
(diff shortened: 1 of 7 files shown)


xmariachi (approved):
ok, lgtm! 
Approving preemptively.
My only suggestion would be to accumulate the rewards to be given back, to optimize the sending of rewards for all failing topics in one go, instead of having one per failing topic. In practice, since it's such an uncommon scenario, it shouldn't matter that much.

amimart (approved):

PR #816, opened 2025-05-06
Generic labels in metrics, usage in events and q/tx

xmariachi:
* Metrics methods refactor to provide labels generically 
* Modified events, tx and queries metrics to send labels where appropriate (eg topicId, nonce, address, etc).

- [x] If tested, please describe how. If not, why tests are not needed. -- Unit tests updated, tested with integration tests. 
- [x] If documented, please describe where. If not, describe why docs are not needed. -- Added to the emissions module README (also making it worthwhile - previous one was incorrect based on a copy from simd)


x/emissions/keeper/msgserver/msg_server_registrations.go
```diff
@@ -14,7 +15,17 @@ import (
...
 func (ms msgServer) Register(ctx context.Context, msg *types.RegisterRequest) (_ *types.RegisterResponse, err error) {
-	defer metrics.RecordMetrics("Register", time.Now(), &err)
+	actorLabel := "reputer"
+	if !msg.IsReputer {
+		actorLabel = "worker"
+	}
+
+	labels := map[string]string{
+		"topic_id": strconv.FormatUint(msg.TopicId, 10),
+		"address":  msg.Sender,
```
amimart, line 25: We should avoid labels with high cardinality, I think setting `address` as label would create too much metric entries which can impacts node performance

x/emissions/keeper/msgserver/msg_server_reputer_payload.go
```diff
@@ -14,27 +15,54 @@ import (
...
-	} else if !canSubmit {
-		return nil, types.ErrNotPermittedToSubmitReputerPayload
+	// if validated, record metrics with appropriate labels
+	labels := map[string]string{
+		"address":     msg.ReputerValueBundle.ValueBundle.Reputer,
+		"topic_id":    strconv.FormatUint(msg.ReputerValueBundle.ValueBundle.TopicId, 10),
+		"nonce":       strconv.FormatUint(uint64(msg.ReputerValueBundle.ValueBundle.ReputerRequestNonce.ReputerNonce.BlockHeight), 10),
+		"blockHeight": strconv.FormatInt(blockHeight, 10),
 	}
+	defer metrics.RecordMetrics("InsertReputerPayload", time.Now(), &err, labels)
```
amimart, line 65: I have multiple remarks here:
- The `time.Now()` being executed here would provide wrong latency measurement;
- The `address`, `nonce`, `blockHeight` introduce a very high cardinality, that can be dangerous;

I think we should have only one `defer` statement at the beginning, and if we need to evaluate vars at execution we can use `defer func() {}` instead

x/emissions/keeper/msgserver/msg_server_reputer_payload.go
```diff
@@ -14,27 +15,54 @@ import (
...
 // A tx function that accepts a individual loss and possibly returns an error
 func (ms msgServer) InsertReputerPayload(ctx context.Context, msg *types.InsertReputerPayloadRequest) (_ *types.InsertReputerPayloadResponse, err error) {
-	defer metrics.RecordMetrics("InsertReputerPayload", time.Now(), &err)
-
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
 	blockHeight := sdkCtx.BlockHeight()
 
 	moduleParams, err := ms.k.GetParams(ctx)
 	if err != nil {
+		defer metrics.RecordMetrics("InsertReputerPayload", time.Now(), &err, map[string]string{"error": err.Error()})
```
amimart, line 23: Having error messages as label can creates a lot of metric entries and are hard to use (e.g. error msgs can contains values creating multiple error labels for the same one), it's preferable to use error codes instead

x/emissions/keeper/queryserver/query_server_inferences.go
```diff
@@ -14,7 +15,11 @@ import (
 
 // GetWorkerLatestInferenceByTopicId handles the query for the latest inference by a specific worker for a given topic.
 func (qs queryServer) GetWorkerLatestInferenceByTopicId(ctx context.Context, req *emissionstypes.GetWorkerLatestInferenceByTopicIdRequest) (_ *emissionstypes.GetWorkerLatestInferenceByTopicIdResponse, err error) {
-	defer metrics.RecordMetrics("GetWorkerLatestInferenceByTopicId", time.Now(), &err)
+	labels := map[string]string{
+		"worker_address": req.WorkerAddress,
+		"topic_id":       strconv.FormatUint(req.TopicId, 10),
+	}
+	defer metrics.RecordMetrics("GetWorkerLatestInferenceByTopicId", time.Now(), &err, labels)
```
amimart, line 22: Is it needed to have labels on queries? I understand the need of labels when it's related to a mutation in the state but I'm not sure it's relevant for queries

x/emissions/metrics/metrics_util.go
```diff
@@ -7,53 +7,103 @@ import (
...
 	)
 }
 
-func RecordMetrics(apiMethod string, startTime time.Time, err *error) {
-	IncrementRpcRequestCounter(apiMethod, err)
-	MeasureRpcRequestLatency(apiMethod, startTime)
+func RecordMetrics(apiMethod string, startTime time.Time, err *error, labels map[string]string) {
+	if labels == nil {
+		labels = make(map[string]string)
+	}
```
amimart, line 74:
```suggestion
	if labels == nil {
		labels = make(map[string]string)
	}
```
We can read from a `nil` map so let it be `nil`

x/emissions/types/events_emitters.go
```diff
@@ -1,67 +1,70 @@
...
 
 /// Scores
 
-func EmitNewInfererScoresSetEvent(ctx sdk.Context, scores []Score) {
+func EmitNewInfererScoresSetEvent(ctx sdk.Context, scores []Score, topicId TopicId) {
 	if len(scores) < 1 {
 		return
 	}
-	metrics.IncrProducerEventCount(metrics.INFERER_SCORE_EVENT)
+	metrics.IncrProducerEventCountWithLabels(metrics.INFERER_SCORE_EVENT, map[string]string{"topic_id": fmt.Sprintf("%d", topicId)})
```
amimart, line 18: Maybe better to use `strconv.FormatUint` as it is done in other places

amimart (changes requested):
Hey I've let some remarks,

My main concern is about metrics, having labels that can carry a lot of different values can be dangerous for the node's memory, I even think it is a bad idea to have topic ids in metrics..

If we really need such labels I think we should consider implementing a metrics exporter that would expose them by watching the chain's state

xmariachi:
> Hey I've let some remarks,
> 
> My main concern is about metrics, having labels that can carry a lot of different values can be dangerous for the node's memory, I even think it is a bad idea to have topic ids in metrics..
> 
> If we really need such labels I think we should consider implementing a metrics exporter that would expose them by watching the chain's state

Thanks for the review. Yeah you're totally right about the high cardinality. I'm reviewing this. TopicId would be very useful, but the concern stays and we may just want to get that from offchain data only, in which case this PR may just be discarded.

felipead (approved):
Thanks Diego, 

I agree with Arnaud's argument on the node's memory.

Can we do a simpler PR just with the `TopicID` and the essentials for monitoring mainnet?

I am afraid the metrics exported, although the ideal solution, will not be ready on time.

xmariachi:
It was decided to move out from adding cardinality to the node and use offchain data for topicId-based alerts. Closing PR.

PR #819, opened 2025-05-08
Mint module metrics + refactor on metric names

xmariachi:
* Add mint metrics on events

- [x] If tested, please describe how. If not, why tests are not needed.  -- unit and integration tests 
- [x] If documented, please describe where. If not, describe why docs are not needed. -- fixed docs on module `README`s


mmallora (dismissed):
Good! @felipead @spooktheducks please add your thoughts or approve.

x/emissions/metrics/metrics_util.go
```diff
@@ -29,10 +29,10 @@ func IncrementRpcRequestCounter(endpoint string, err *error) {
 // Measures the RPC request latency in milliseconds
 // Metric Name:
 //
-//	allora_rpc_request_latency_ms
+//	allora_request_latency_ms
 func MeasureRpcRequestLatency(endpoint string, startTime time.Time) {
 	metrics.MeasureSinceWithLabels(
-		[]string{"allora", "request", "latency_ms"},
+		[]string{"allora", "emissions", "request", "latency_ms"},
```
amimart:
```suggestion
		[]string{"allora", "emissions", "request", "duration_ms"},
```

I think `duration` would be more appropriate as latency suggests a delay

x/emissions/metrics/metrics_util.go
```diff
@@ -13,11 +13,11 @@ import (
 // Measures RPC endpoint request throughput
 // Metric Name:
 //
-//	allora_rpc_request_counter
+//	allora_emissions_request_counter
```
amimart, line 16: I think we can remove this metric as the `MeasureRpcRequestLatency` provides a counter already

x/emissions/metrics/metrics_util.go
```diff
@@ -44,10 +44,10 @@ func MeasureRpcRequestLatency(endpoint string, startTime time.Time) {
 // This metric counts the number of events produced by the system.
 // Metric Name:
 //
-//	allora_loadtest_produce_count
+//	allora_emissions_produce_count
 func IncrProducerEventCount(msgType string) {
 	telemetry.IncrCounterWithLabels(
-		[]string{"allora", "loadtest", "produce", "count"},
+		[]string{"allora", "emissions", "event", "count"},
```
amimart:
```suggestion
		[]string{"allora", "emissions", "event", "total"},
```

A counter is usually suffixed by `_total` to differentiate it from the counter of an histogram

amimart (commented):
I've let a few remarks, I think for the requests metric we could provide only one metric and put the module as a label instead of putting it in the metric's name, and our `metrics` package could then be shared among modules

spooktheducks (dismissed):

felipead (dismissed):

x/emissions/metrics/metrics_util.go
xmariachi: For what I've checked we would be missing some features and cleanliness. I think I favour keeping a clean success/failure counter here and keep the duration metric as such.

amimart (approved):

guilherme-brandao (approved):

PR #827, opened 2025-05-27
docs: Fix typo in ReadFromClientConfig comment

gap-editor:
Fixed a typo in the comment preceding ReadFromClientConfig in cmd/allorad/cmd/root.go, changing "ets" to "sets" for better clarity.

Before:
// This needs to go after ReadFromClientConfig, as that function ets the RPC client...

After:
// This needs to go after ReadFromClientConfig, as that function sets the RPC client...


spooktheducks:
Incorporated in #914

PR #830, opened 2025-06-02
fix: refactor all x/emissions unit tests

zale144:

The unit tests in the `x/emissions` module needed refactoring to make them compatible with the way natural topic flow would happen, and to accommodate the changes to be done in https://github.com/allora-network/allora-chain/pull/812.

- The helper function `getRewardsDistribution` is removed
- A new set of configurable helper functions is added for simulating a full topic pass
- Large tests were reduced in size
- Code repetition is reduced
- `TestSuite` is centralized and shared by most testing packages
- Some bugs were addressed

This is only the first such iteration of refactoring, it does not address every issue, so there should be more fixes in the future.

## Still Left Todo

### DIFF
--- a/test/testutil/test_utils.go
+++ b/test/testutil/test_utils.go
@@ -68,6 +68,7 @@ type TestSuite struct {
 }
 
 func NewTestSuite(moduleName string) TestSuite {
+	//nolint:exhaustruct
 	return TestSuite{
 		ModuleName: moduleName,
 	}
@@ -753,6 +754,7 @@ type builderValueO interface {
 	SetValue(alloraMath.Dec)
 }
 
+//nolint:forcetypeassert
 func createValue[O builderValueO](worker string, value alloraMath.Dec) (out O) {
 	out = reflect.New(reflect.TypeOf(out).Elem()).Interface().(O)
 	out.SetWorker(worker)
@@ -834,7 +836,6 @@ func (s *TestSuite) CreateTopic(opts ...Option) uint64 {
 		s.Require().NoError(err)
 		maxGTL := prms.MaxUnfulfilledReputerRequests * uint64(newTopicMsg.EpochLength)
 		if uint64(newTopicMsg.GroundTruthLag) > maxGTL {
-			// TODO: on FullTopicPass they have to be the same. why?
 			newTopicMsg.GroundTruthLag = newTopicMsg.EpochLength
 		}
 	}
@@ -1031,6 +1032,18 @@ func (s *TestSuite) FullTopicSetup(workerIndexes, reputerIndexes []int, options 
 	return topic
 }
 
+// FullTopicPass runs a full pass for a topic. It:
+// 	- takes the option and sets up the topic
+// 	- in case topicID is provided, it assumes that topic exists so it skips creating it,
+// 	- finds the next churning block for the topic and uses it as the nonce,
+// 	- submits inferences for the unfulfilled worker nonce,
+// 	- forwards the chain to the end of the epoch,
+// 	- submits reputer loss bundles for the unfulfilled reputer nonce,
+// 	- forwards the chain to the rewards end blocker,
+// 	- runs the end blocker for closing the reputer nonce and calculating and distributing rewards (when appropriate),
+// 	- when running on a newly created topic, there will be no losses, so no rewards will be distributed.
+// Limitations:
+// 	- with the current design, `ground_truth_lag` needs to be the same as `epoch_length`.
 func (s *TestSuite) FullTopicPass(workerIndexes, reputerIndexes []int, options ...Option) (uint64, int64) {
 	p := new(customParams)
 	for _, opt := range options {
--- a/test/testutil/test_utils.go
+++ b/test/testutil/test_utils.go
@@ -1,0 +1,1094 @@
+package testutil
+
+import (
+	"errors"
+	"math/rand"
+	"reflect"
+	"slices"
+	"strconv"
+	"time"
+
+	"cosmossdk.io/core/header"
+	"cosmossdk.io/core/store"
+	"cosmossdk.io/log"
+	cosmosMath "cosmossdk.io/math"
+	storetypes "cosmossdk.io/store/types"
+	"github.com/cometbft/cometbft/crypto/secp256k1"
+	"github.com/cosmos/cosmos-sdk/codec"
+	codecAddress "github.com/cosmos/cosmos-sdk/codec/address"
+	"github.com/cosmos/cosmos-sdk/runtime"
+	"github.com/cosmos/cosmos-sdk/testutil"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	moduletestutil "github.com/cosmos/cosmos-sdk/types/module/testutil"
+	"github.com/cosmos/cosmos-sdk/x/auth"
+	authcodec "github.com/cosmos/cosmos-sdk/x/auth/codec"
+	authkeeper "github.com/cosmos/cosmos-sdk/x/auth/keeper"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	"github.com/cosmos/cosmos-sdk/x/bank"
+	bankkeeper "github.com/cosmos/cosmos-sdk/x/bank/keeper"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+	stakingkeeper "github.com/cosmos/cosmos-sdk/x/staking/keeper"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/allora-network/allora-chain/app/params"
+	alloralog "github.com/allora-network/allora-chain/log"
+	alloraMath "github.com/allora-network/allora-chain/math"
+	"github.com/allora-network/allora-chain/utils/fn"
+	"github.com/allora-network/allora-chain/x/emissions/keeper"
+	actorutils "github.com/allora-network/allora-chain/x/emissions/keeper/actor_utils"
+	inferencesynthesis "github.com/allora-network/allora-chain/x/emissions/keeper/inference_synthesis"
+	"github.com/allora-network/allora-chain/x/emissions/keeper/msgserver"
+	"github.com/allora-network/allora-chain/x/emissions/keeper/queryserver"
+	"github.com/allora-network/allora-chain/x/emissions/module"
+	"github.com/allora-network/allora-chain/x/emissions/types"
+	mintkeeper "github.com/allora-network/allora-chain/x/mint/keeper"
+	minttypes "github.com/allora-network/allora-chain/x/mint/types"
+)
+
+type TestSuite struct {
+	suite.Suite
+	ModuleName string
+
+	ctx                   sdk.Context
+	codec                 codec.Codec
+	storeServiceBank      store.KVStoreService
+	storeServiceEmissions store.KVStoreService
+	accountKeeper         authkeeper.AccountKeeper
+	bankKeeper            bankkeeper.BaseKeeper
+	emissionsKeeper       *keeper.Keeper
+	mintKeeper            minttypes.MintKeeper
+	stakingKeeper         minttypes.StakingKeeper
+	emissionsAppModule    module.AppModule
+	emissionsQueryServer  types.QueryServiceServer
+	emissionsMsgServer    types.MsgServiceServer
+	privKeys              []secp256k1.PrivKey
+	addrs                 []sdk.AccAddress
+	addrsStr              []string
+	pubKeyHexStr          []string
+}
+
+func NewTestSuite(moduleName string) TestSuite {
+	return TestSuite{
+		ModuleName: moduleName,
+	}
+}
+
+func (s *TestSuite) Ctx() sdk.Context {
+	return s.ctx
+}
+
+func (s *TestSuite) Addrs(idx int) sdk.AccAddress {
+	return s.addrs[idx]
+}
+
+func (s *TestSuite) AddrsStr(idx int) string {
+	return s.addrsStr[idx]
+}
+
+func (s *TestSuite) PubKeyHexStr(idx int) string {
+	return s.pubKeyHexStr[idx]
+}
+
+func (s *TestSuite) PrivKeys(idx int) secp256k1.PrivKey {
+	return s.privKeys[idx]
+}
+
+func (s *TestSuite) LenAccounts() int {
+	return len(s.addrs)
+}
+
+func (s *TestSuite) EmissionsKeeper() *keeper.Keeper {
+	return s.emissionsKeeper
+}
+
+func (s *TestSuite) AccountKeeper() authkeeper.AccountKeeper {
+	return s.accountKeeper
+}
+
+func (s *TestSuite) BankKeeper() keeper.BankKeeper {
+	return s.bankKeeper
+}
+
+func (s *TestSuite) MintKeeper() minttypes.MintKeeper {
+	return s.mintKeeper
+}
+
+func (s *TestSuite) StakingKeeper() minttypes.StakingKeeper {
+	return s.stakingKeeper
+}
+
+func (s *TestSuite) EmissionsAppModule() module.AppModule {
+	return s.emissionsAppModule
+}
+
+func (s *TestSuite) EmissionsMsgServer() types.MsgServiceServer {
+	return s.emissionsMsgServer
+}
+
+func (s *TestSuite) EmissionsQueryServer() types.QueryServiceServer {
+	return s.emissionsQueryServer
+}
+
+func (s *TestSuite) StoreServiceEmissions() store.KVStoreService {
+	return s.storeServiceEmissions
+}
+
+func (s *TestSuite) StoreServiceBank() store.KVStoreService {
+	return s.storeServiceBank
+}
+
+func (s *TestSuite) WithBlockHeight(height int64) {
+	s.ctx = s.ctx.WithBlockHeight(height)
+}
+
+const (
+	multiPerm  = "multiple permissions account"
+	randomPerm = "random permission"
+)
+
+type (
+	Option       func(s *customParams)
+	customParams struct {
+		block             int64
+		topicID           uint64
+		initialTopicStake cosmosMath.Int
+		alphaRegret       alloraMath.Dec
+		epochLength,
+		groundTruthLag,
+		workerSubmissionWindow,
+		epochLastEnded int64
+		lossMethod,
+		initialRegret string
+		workerValues          []TestWorkerValue
+		reputerValues         []TestReputerValue
+		skipNetworkInferences bool
+		reputerStake          *cosmosMath.Int
+	}
+)
+
+func WithBlock(block int64) Option {
+	return func(s *customParams) {
+		s.block = block
+	}
+}
+
+func WithTopicID(topicID uint64) Option {
+	return func(s *customParams) {
+		s.topicID = topicID
+	}
+}
+
+func WithInitialTopicStake(stake cosmosMath.Int) Option {
+	return func(s *customParams) {
+		s.initialTopicStake = stake
+	}
+}
+
+func WithAlphaRegret(alphaRegret alloraMath.Dec) Option {
+	return func(s *customParams) {
+		s.alphaRegret = alphaRegret
+	}
+}
+
+func WithEpochLength(epochLength int64) Option {
+	return func(s *customParams) {
+		s.epochLength = epochLength
+	}
+}
+
+func WithGroundTruthLag(lag int64) Option {
+	return func(s *customParams) {
+		s.groundTruthLag = lag
+	}
+}
+
+func WithWorkerSubmissionWindow(window int64) Option {
+	return func(s *customParams) {
+		s.workerSubmissionWindow = window
+	}
+}
+
+func WithWorkerValues(workerValues []TestWorkerValue) Option {
+	return func(s *customParams) {
+		s.workerValues = workerValues
+	}
+}
+
+func WithReputerValues(reputerValues []TestReputerValue) Option {
+	return func(s *customParams) {
+		s.reputerValues = reputerValues
+	}
+}
+
+func WithReputerStake(reputerStake *cosmosMath.Int) Option {
+	return func(s *customParams) {
+		s.reputerStake = reputerStake
+	}
+}
+
+func WithEpochLastEnded(epochLastEnded int64) Option {
+	return func(s *customParams) {
+		s.epochLastEnded = epochLastEnded
+	}
+}
+
+func WithLossMethod(lossMethod string) Option {
+	return func(s *customParams) {
+		s.lossMethod = lossMethod
+	}
+}
+
+func WithInitialRegret(initialRegret string) Option {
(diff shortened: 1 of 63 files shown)


x/emissions/keeper/msgserver/msg_server_registrations_test.go
```diff
@@ -204,34 +195,32 @@ func (s *MsgServerTestSuite) TestMsgRemoveRegistrationWorker() {
...
+	ctx, msgServer := s.Ctx(), s.EmissionsMsgServer()
 	require := s.Require()
 
 	// Mock setup for addresses
-	reputerAddr := s.addrs[0]
-	topic1 := s.CreateOneTopic()
-	err := s.emissionsKeeper.SetTopic(ctx, topic1.Id, topic1)
-	require.NoError(err)
-	err = s.emissionsKeeper.ActivateTopic(ctx, topic1.Id)
+	reputerAddr := sdk.AccAddress(secp256k1.GenPrivKey().PubKey().Address())
```
guilherme-brandao: why not getting the address from the suite here? `s.Addrs(0)`

x/emissions/keeper/msgserver/msg_server_registrations_test.go
```diff
@@ -240,31 +229,31 @@ func (s *MsgServerTestSuite) TestMsgRegisterReputerInsufficientBalance() {
...
 func (s *MsgServerTestSuite) TestMsgRegisterReputerInsufficientDenom() {
-	ctx, msgServer := s.ctx, s.msgServer
+	ctx, msgServer := s.Ctx(), s.EmissionsMsgServer()
 	require := s.Require()
-	topic1 := s.CreateOneTopic()
+	topic := uint64(1)
 
 	// Mock setup for addresses
-	reputerAddr := s.addrs[0]
+	reputerAddr := sdk.AccAddress(secp256k1.GenPrivKey().PubKey().Address())
```
guilherme-brandao: same

x/emissions/keeper/msgserver/msg_server_registrations_test.go
```diff
@@ -274,59 +263,37 @@ func (s *MsgServerTestSuite) TestMsgRegisterReputerInsufficientDenom() {
 
 func (s *MsgServerTestSuite) TestBlocklistedAddressUnableToRegister() {
 	// Reputer Addresses
-	reputer := s.addrs[2]
+	reputer := sdk.AccAddress(secp256k1.GenPrivKey().PubKey().Address())
```
guilherme-brandao: same

x/emissions/keeper/msgserver/msg_server_registrations_test.go
```diff
@@ -274,59 +263,37 @@ func (s *MsgServerTestSuite) TestMsgRegisterReputerInsufficientDenom() {
...
 	)
-	s.emissionsKeeper = keeper.NewKeeper(
-		s.codec,
-		s.addressCodec,
-		s.storeService,
-		s.accountKeeper,
-		s.bankKeeper,
+	emissionsKeeper := keeper.NewKeeper(
+		s.Codec(),
+		codecAddress.NewBech32Codec(params.Bech32PrefixAccAddr),
```
guilherme-brandao: why not getting from the suite here as well?

x/emissions/module/rewards/nonce_management_test.go
```diff
@@ -2,310 +2,153 @@ package rewards_test
...
 	}
 
 	err = actorutils.CloseReputerNonce(
-		&s.emissionsKeeper, s.ctx, topic,
-		*lossBundles.ReputerValueBundles[0].ValueBundle.ReputerRequestNonce.ReputerNonce,
+		s.EmissionsKeeper(), s.Ctx(), topic,
+		reputerNonce,
 	)
-	s.Require().Error(err)
+	// s.Require().Error(err) // TODO: ???
```
guilherme-brandao: This was already discussed in the last meeting. Just need to make sure the defer function in `CloseReputerNonce` is triggered

test/testutil/test_utils.go
```diff
@@ -0,0 +1,1019 @@
...
+	s.EndBlock()
+
+	reputerNonces, err := s.emissionsKeeper.GetUnfulfilledReputerNonces(s.Ctx(), p.topicID)
+	s.Require().NoError(err)
+
+	topic, err = s.emissionsKeeper.GetTopic(s.Ctx(), topic.Id)
+	s.Require().NoError(err)
+
+	// TODO: check this if statement
+	if len(reputerNonces.Nonces) > 0 && topic.EpochLastEnded > topic.EpochLength {
```
xmariachi: The second leg of this "if" would just not make reputer loss bundle insertion, only if this is the very first iteration of the topic. I don't think we need this.
We are already moving the block height in the next line to a valid one for insertion of this nonce.

test/testutil/test_utils.go
```diff
@@ -0,0 +1,1019 @@
...
+		s.SetupParticipants(topicId, reputerIndexes, true, options...)
+	}
+
+	topic, err := s.emissionsKeeper.GetTopic(s.Ctx(), topicId)
+	s.Require().NoError(err)
+
+	return topic
+}
+
+func (s *TestSuite) FullTopicPass(workerIndexes, reputerIndexes []int, options ...Option) (uint64, int64) {
```
xmariachi: Since the way to tackle this problem may be opinionated, I think it would be good to add a method comment on the overall intention of how we are approaching this, and what this method attempts to do - and any limitation, e.g. if sensitivity to scenarios like `ground_truth_lag > 2x epoch_length`. 
Same goes for FullTopicSetup.

x/emissions/module/rewards/rewards_test.go
```diff
@@ -1,620 +1,106 @@
...
-		s.MintTokensToAddress(s.addrs[index], stakes[i])
-		_, err := s.msgServer.AddStake(s.ctx, &types.AddStakeRequest{
-			Sender:  s.addrsStr[index],
-			Amount:  stakes[i],
-			TopicId: topicId2,
-		})
-		s.Require().NoError(err)
-	}
+	// TOPIC 2 - PRE PASS
+	topic2 := s.FullTopicSetup(workerIndexes2, reputerIndexes2)
```
xmariachi, line 63: The initial run, In topic2, is a `FullTopicSetup` call and in topic1 it is a `FullTopicPass`. Just to make sure if this should be aligned, or if it is intended, and if so, why.

x/emissions/module/rewards/rewards_test.go
```diff
@@ -634,2446 +120,559 @@ func (s *RewardsTestSuite) TestStandardRewardEmissionShouldRewardTopicsWithFulfi
...
-		topic.EpochLastEnded,
-		workerValues,
-		reputerValues,
-		workerZeroAddress,
-		workerZeroOneOutInfererValue,
-		workerZeroInfererValue,
-	)
+	// Check that phase 1 rewards are NOT equal between alpha values
+	require.False(areTaskRewardsEqualIgnoringTopicId(s, rewardsDistributions[0][1], rewardsDistributions[1][1]),
+		"Rewards in the second phase should differ between alphas")
```
xmariachi, line 230: This is not something fromthe test - but "inequality" is possibly not the best assertion. The method comment says: 
> "We increase TaskRewardAlpha between the trials to show that weighting current performance more heavily
means that the worker is rewarded more for their better performance in the 2nd epoch of the 2nd trial."

So, shouldn't it be checking that one is bigger than the other instead of plain inequality?

x/emissions/module/rewards/rewards_test.go
```diff
@@ -634,2446 +120,559 @@ func (s *RewardsTestSuite) TestStandardRewardEmissionShouldRewardTopicsWithFulfi
...
-		block,
-		*networkLossBundles)
-	s.Require().NoError(err)
-
-	s.Require().Equal(uint64(len(infererScores)), params.GetMaxTopInferersToReward(), "Only few Top inferers can get reward")
-	s.Require().Equal(uint64(len(forecasterScores)), params.GetMaxTopForecastersToReward(), "Only few Top forecasters can get reward")
-}
-
-/* to be rewritten in PROTO-3088
-func (s *RewardsTestSuite) TestTotalInferersRewardFractionGrowsWithMoreInferers() {
```
xmariachi, line 2549: These preexisting tests, have they been reformulated/moved, or plain removed? What is the reason?

xmariachi (commented):
Partial review. Most done but in the meantime publishing these comments.

test/testutil/test_utils.go
xmariachi: yeah outdated comment, resolving

x/emissions/keeper/keeper_test.go
```diff
@@ -2,252 +2,59 @@ package keeper_test
...
 }
 
 func TestKeeperTestSuite(t *testing.T) {
-	suite.Run(t, new(KeeperTestSuite))
+	suite.Run(t, &KeeperTestSuite{
+		alloratestutil.NewTestSuite("emissions_keeper"),
+	})
 }
 
 func (s *KeeperTestSuite) MintTokensToAddress(address sdk.AccAddress, amount cosmosMath.Int) {
```
guilherme-brandao: `KeeperTestSuite` already has this method. Check `test_utils.go`

x/emissions/keeper/keeper_test.go
```diff
@@ -1314,171 +1117,155 @@ func (s *KeeperTestSuite) TestGetForecastsAtBlock() {
...
 	require.Empty(result.ReputerValueBundles, "Result should be empty for non-existent data")
 }
 
 func (s *KeeperTestSuite) TestInsertNetworkLossBundleAtBlock() {
-	ctx := s.ctx
+	ctx := s.Ctx()
 	require := s.Require()
 	topicId := uint64(1)
 	block := types.BlockHeight(100)
+	//nolint:exhaustruct
```
guilherme-brandao, line 1176: We can think about loosening this `exhaustruct` rule in tests cc @xmariachi

x/emissions/keeper/queryserver/query_server_test.go
```diff
@@ -1,262 +1,65 @@
...
 func (s *QueryServerTestSuite) MintTokensToAddress(address sdk.AccAddress, amount cosmosMath.Int) {
 	creatorInitialBalanceCoins := sdk.NewCoins(sdk.NewCoin(params.DefaultBondDenom, amount))
 
-	err := s.bankKeeper.MintCoins(s.ctx, types.AlloraStakingAccountName, creatorInitialBalanceCoins)
+	err := s.BankKeeper().MintCoins(s.Ctx(), types.AlloraStakingAccountName, creatorInitialBalanceCoins)
 	s.Require().NoError(err)
-	err = s.bankKeeper.SendCoinsFromModuleToAccount(s.ctx, types.AlloraStakingAccountName, address, creatorInitialBalanceCoins)
+	err = s.BankKeeper().SendCoinsFromModuleToAccount(s.Ctx(), types.AlloraStakingAccountName, address, creatorInitialBalanceCoins)
 	s.Require().NoError(err)
 }
```
guilherme-brandao: Same here

x/emissions/types/reputer.go
```diff
@@ -0,0 +1,11 @@
...
+
+import alloramath "github.com/allora-network/allora-chain/math"
+
+func (workerValue *WorkerAttributedValue) GetValue() alloramath.Dec {
+	return workerValue.Value
+}
+
+func (withheldWorkerValue *WithheldWorkerAttributedValue) GetValue() alloramath.Dec {
+	return withheldWorkerValue.Value
+}
```
guilherme-brandao, line 11: I think this can be moved somewhere else since this structure can be used for network inferences and loss, it is not specific to reputer payload structure

x/emissions/keeper/queryserver/query_server_inferences_test.go
```diff
@@ -379,157 +292,67 @@ func (s *QueryServerTestSuite) TestGetLatestNetworkInferences() {
...
-		},
+	// Set forecaster regrets
+	for i, forecaster := range forecasters {
+		err = keeper.SetForecasterNetworkRegret(s.Ctx(), topicId, forecaster, getWorkerRegretValue(fmt.Sprintf("0.%d", i+1)))
+		require.NoError(err)
 	}
 
-	err = keeper.InsertActiveForecasts(s.ctx, topicId, inferenceNonce.BlockHeight, forecasts)
+	// Set inferences and forecasts
+	inferernces := getInferencesForBlockHeight(workers, inferenceBlockHeight, topicId)
```
guilherme-brandao: typo => inferences

x/emissions/keeper/queryserver/query_server_inferences_test.go
```diff
@@ -729,207 +552,163 @@ func (s *QueryServerTestSuite) TestGetLatestTopicInferences() {
...
-	inferenceBlockHeight := epochLastEnded + epochLength
-	inferenceBlockHeight2 := inferenceBlockHeight + epochLength
-
-	lossNonce := types.Nonce{BlockHeight: lossBlockHeight}
-	// inferenceNonce := types.Nonce{BlockHeight: inferenceBlockHeight}
-	inferenceNonce2 := types.Nonce{BlockHeight: inferenceBlockHeight2}
+	nonces := map[string]*types.Nonce{
+		"loss":       {BlockHeight: blockHeights["loss"]},
+		"inference2": {BlockHeight: blockHeights["inference2"]},
+	}
```
guilherme-brandao, line 574: I preferred this part the way it was, tbh, but it’s not a big deal to stick with this way

guilherme-brandao (commented):
Nice work! So comments...

x/emissions/keeper/keeper_test.go
xmariachi: I know it's annoying but I think it forces you to think about whether you should set all fields in the cases, or whether that could be a source of issues in your test... so I lean towards keeping it for now.

x/emissions/types/reputer.go
xmariachi: Yeah, that makes sense to me as well.

x/emissions/module/rewards/rewards_test.go
zale144: Good point. In the original test, the bundles for topic2 are never sent, the revenue for topic2 was supposed to decrease due to the rewards distribution for topic1, because all topics rewards are shared. The refactored test is slightly different, now the topic2 full pass will trigger this mutual revenue decrease. I've updated the test so that it follows the original flow (topic1 full pass triggers the mutual revenue decrease).

x/emissions/module/rewards/rewards_test.go
zale144: I agree. But since we are comparing rewards for multiple addresses, `areTaskRewardsEqualIgnoringTopicId` is used, which only returns a boolean (using a delta to determine the difference). This function could perhaps be refactored in order to return the information about which is higher (if difference is more than delta). Wdyt?

x/emissions/module/rewards/rewards_test.go
zale144: `TestTotalInferersRewardFractionGrowsWithMoreInferers` was originally commented out. I should have probably asked if this test is still needed/relevant before removing it. So, do we need it?

x/emissions/keeper/msgserver/msg_server_registrations_test.go
zale144: all the s.Addr addresses are with some funds, we need an address here without funds

x/emissions/keeper/msgserver/msg_server_registrations_test.go
zale144: same here

x/emissions/keeper/msgserver/msg_server_registrations_test.go
zale144: here we need an address without funds and not whitelisted

x/emissions/keeper/msgserver/msg_server_registrations_test.go
zale144: this is now outdated

x/emissions/module/rewards/nonce_management_test.go
zale144: this is also outdated

x/emissions/keeper/keeper_test.go
zale144: this one was already addressed, however I just saw there were still some leftovers in the query server tests. Thanks for pointing it out

x/emissions/keeper/queryserver/query_server_test.go
zale144: nice catch!

x/emissions/types/reputer.go
zale144: the reason I put it here was because the type `WorkerAttributedValue` is in the file `reputer.pb.go`, so I created a file with the matching name `reputer.go`, to extend the methods of the declared type. I'm not super opinionated about this, however I think it's reasonable. Same goes for `inputreputer.go`.

x/emissions/keeper/queryserver/query_server_inferences_test.go
zale144: agreed, will update

x/emissions/module/rewards/rewards_test.go
xmariachi: makes sense!

x/emissions/keeper/whitelist_test.go
```diff
@@ -1,724 +1,506 @@
...
-	err = keeper.AddToGlobalWhitelist(ctx, testAddr)
-	s.Require().NoError(err)
 
-	enabled, err = keeper.IsEnabledGlobalActor(ctx, testAddr)
-	s.Require().NoError(err)
-	s.Require().True(enabled)
+	testCases := []struct {
+		name         string
+		setup        func(context.Context, string) error
+		setup1       func(context.Context, uint64, string) error
```
xmariachi: Not immediately clear what is the intention of setup/setup1 (same for check, remove). Could use more clear naming here (and/or comment)

x/emissions/module/rewards/rewards_test.go
xmariachi: No need - we have a ticket about it. I think it's ok, resolving conversation.

xmariachi (commented):
in general lgtm, let's finish up the review conversations and fix the linter issues

test/testutil/test_utils.go
```diff
@@ -0,0 +1,1107 @@
...
+	inferencesynthesis "github.com/allora-network/allora-chain/x/emissions/keeper/inference_synthesis"
+	"github.com/allora-network/allora-chain/x/emissions/keeper/msgserver"
+	"github.com/allora-network/allora-chain/x/emissions/keeper/queryserver"
+	"github.com/allora-network/allora-chain/x/emissions/module"
+	"github.com/allora-network/allora-chain/x/emissions/types"
+	mintkeeper "github.com/allora-network/allora-chain/x/mint/keeper"
+	minttypes "github.com/allora-network/allora-chain/x/mint/types"
+)
+
+type TestSuite struct {
```
amimart: This is exclusively used in the `x/emissions` module so maybe it should rely in it instead of this general package

amimart (commented):
This is a really great thing to have a single test suite, it'll definitely help!

As a side note for potential further improvements, I feel like we rely too much on the emissions keeper true implementation when testing parts that are only client of it, for example `queryserver`, `msgserver`, `rewards`, etc.. should be tested against the keeper's expected behaviour, not its actual implementation. And we could use a simple mock instead of setting up an entire application, I think it'd make tests in these parts even easier and especially error cases.

I'd be interested in your point of view on that

zale144:
I mean, yeah mocking would be great for those clients, where applicable. I was thinking about doing it, but it might be too many changes for this PR. From this point on, I think it would be much easier to achieve it since it's more centralized now.

x/emissions/types/reputer.go
xmariachi: That is reasonable @zale144 
I suppose a case can be made to refactor the following types: 
ValueBundle
WithheldWorkerAttributedValue
WorkerAttributedValue

into its own `valuebundle.proto` file  (and separately, same for `inputvaluebundle.proto` ). So then, ValueBundle can be used in several other places too as expressed by @guilherme-brandao .
what do you guys think ?

test/testutil/test_utils.go
zale144: My idea was to also use it in x/mint, but it was agreed to skip refactoring mint for this PR

amimart: I think the `x/mint` tests should rely on a mocked `x/emissions` keeper and not its actual implementation

zale144: I believe it already uses a mocked version of it. So, my idea was to use interfaces in the test suite and inject the mock/actual type when appropriate. But this is a problem for the future me

x/emissions/types/reputer.go
amimart: I agree, we could also remove the `WithheldWorkerAttributedValue` which is a duplication of  `WorkerAttributedValue`.

But it may be out of the scope of this PR

test/testutil/test_utils.go
amimart: I don't see how we'll manage to inject mint's specific mocks in the test suite. I'm just afraid making a generic test suite for everything may be too much and counterproductive.

btw the current `x/mint` [test suite](https://github.com/allora-network/allora-chain/blob/33e1aeafaf7765c10bf5c89fff9f9dfe2fc9d673/x/mint/keeper/keeper_test.go#L27) is very good, it instantiates the mint keeper using mocked keepers which should be the standard way imho

zale144: I haven't looked at x/mint closely, so you might be right. I don't mind moving the test util to be inside x/emissions for now

x/emissions/types/reputer.go
guilherme-brandao: @xmariachi Yeah, I agree with the changes! But I think since we are making this change, we should double-check if it’s not better to have specific bundles. But either way, if we want to change the type of the value bundle, I would do it as a fast follow. 

@amimart This change is a little bit trickier because it’s the same structure, but it means different things. I’m not sure if deleting one of them would be a good idea in terms of UX.

xmariachi: Yeah, can be done as a ff. Pls create a ticket with definition. 
Pls @guilherme-brandao resolve convo if all good

x/emissions/keeper/whitelist_test.go
xmariachi: Suggestion: 
`setup` to be renamed to `setupWhitelist`
`setup1` to be renamed to `setupTopicWhitelist`
?

zale144: I did add comments, but sure, I can rename them

xmariachi (approved):
lgtm

amimart (approved):
LGTM 🚀

x/emissions/types/reputer.go
xmariachi: Created ticket 3907, resolving this convo

PR #831, opened 2025-06-12
Guilherme/engn 3811 modifiable topics

guilherme-brandao:
This PR introduces the ability for topic creators to update existing topics

- Add `UpdateTopic` message/handler so topic creators can update metadata and selected tuning parameters.
- Emit a `TopicUpdated` event whenever a topic change is applied.

https://linear.app/alloralabs/issue/ENGN-3811/modifiable-topics


x/emissions/types/errors.go
```diff
@@ -97,4 +97,6 @@ var (
 	ErrMaxWhitelistInputArrayLengthExceeded      = errors.Register(ModuleName, 92, "max whitelist input array length exceeded")
 	ErrTopicAlreadyActive                        = errors.Register(ModuleName, 93, "topic already active")
 	ErrTopicCannotBeActivated                    = errors.Register(ModuleName, 94, "topic cannot be activated")
+	ErrNotTopicCreator                           = errors.Register(ModuleName, 95, "not topic creator")
```
xmariachi: Would modify it to sth like "ErrNotPermittedToModifyTopic" or similar, more aligned with other error names, and allows us to define what that means.

x/emissions/module/rewards/topic_rewards.go
```diff
@@ -154,6 +154,26 @@ func GetAndUpdateActiveTopicWeights(
 		if err != nil {
 			continue
 		}
+
+		// Apply pending topic updates at epoch end, before calculating weight
+		hasPendingUpdate, err := k.HasPendingTopicUpdate(ctx, topicId)
+		if err != nil {
+			Logger(ctx).Warn("Error checking pending topic update", "topicId", topicId, "error", err)
```
spooktheducks: Depending on the nature of the error, it seems that this could cause some non-determinism (i.e. problems with consensus re: app hash etc.).  Example: let's say one validator has a hardware error (disk full, whatever), and another does not.  They will come to different conclusions about the action to take here.

x/emissions/keeper/topic_weight.go
```diff
@@ -102,11 +102,7 @@ func (k *Keeper) GetCurrentTopicWeight(
 	return alloraMath.ZeroDec(), topicFeeRevenue, nil
 }
 
-func (k *Keeper) GetTopicWeightFromTopicId(ctx context.Context, topicId types.TopicId) (alloraMath.Dec, error) {
-	params, err := k.GetParams(ctx)
-	if err != nil {
-		return alloraMath.ZeroDec(), err
-	}
+func (k *Keeper) GetTopicWeightFromTopicId(ctx context.Context, topicId types.TopicId, params types.Params) (alloraMath.Dec, error) {
```
spooktheducks: Is `params` used?

spooktheducks (changes requested):

x/emissions/keeper/topic_weight.go
guilherme-brandao: Yes, we use them inside when we call `GetCurrentTopicWeight`

x/emissions/module/rewards/topic_rewards.go
guilherme-brandao: Yeah, I agree! It’s something we need to keep in mind for other parts of the code as well. Sometimes we use `continue` in loops and just log the error; sometimes we return the error; sometimes we panic. In cases where we need to iterate through topics, I think the old rule of thumb was: “if a topic has a problem, this shouldn’t affect other topics' processing.” I think we need to revisit that.

guilherme-brandao: I think it's reasonable to return the errors in this case btw, I will change this.

x/emissions/keeper/msgserver/msg_server_topics.go
```diff
@@ -97,3 +97,89 @@ func (ms msgServer) CreateNewTopic(ctx context.Context, msg *types.CreateNewTopi
...
+	if !hasChanges {
+		return nil, errorsmod.Wrap(types.ErrNoUpdateFields, "No fields to update")
+	}
+
+	// Validate the updated topic
+	if err := updatedTopic.Validate(params); err != nil {
+		return nil, errorsmod.Wrap(err, "Updated topic validation failed")
+	}
+
+	// If topic is inactive, apply immediately; for active topics, defer to epoch end.
```
xmariachi: Sending events: 
We need to send an event when topic gets updated. So, reputers can listen to the event of the change.
Also, the event needs to be triggered when the change comes into effect. So it would be sent within this tx, while the scheduled one will be triggered at end of epoch.

x/emissions/keeper/msgserver/msg_server_topics.go
```diff
@@ -97,3 +97,89 @@ func (ms msgServer) CreateNewTopic(ctx context.Context, msg *types.CreateNewTopi
 	types.EmitNewCreateNewTopicEvent(ctx, &topic)
 	return &types.CreateNewTopicResponse{TopicId: topicId}, err
 }
+
+func (ms msgServer) UpdateTopic(ctx context.Context, msg *types.UpdateTopicRequest) (_ *types.UpdateTopicResponse, err error) {
```
xmariachi, line 102: I think this logic (which parameters get changed, how do they change, etc) should be in the keeper, not in the msgServer. And, the msgServer would just relay the call to the keeper.

xmariachi (commented):
More discussion to be had about the store, for now these two comments

CHANGELOG.md
xmariachi, line 97: We should add the API changes (new tx) here, right? 
I'm thinking also that we should just state "API Changes" not "API Breaking Changes". wdyt? 
@spooktheducks @amimart

x/emissions/keeper/keeper.go
```diff
@@ -3253,6 +3253,105 @@ func (k *Keeper) TopicExists(ctx context.Context, topicId TopicId) (bool, error)
...
+		if err != nil {
+			return types.Topic{}, errorsmod.Wrap(err, "failed to parse p_norm")
+		}
+		updatedTopic.PNorm = pNorm
+		hasChanges = true
+	}
+
+	if !hasChanges {
+		return types.Topic{}, errorsmod.Wrap(types.ErrNoUpdateFields, "no fields to update")
+	}
```
amimart: Why erroring? We could simplify by letting it noop

x/emissions/keeper/keeper.go
```diff
@@ -3253,6 +3253,105 @@ func (k *Keeper) TopicExists(ctx context.Context, topicId TopicId) (bool, error)
...
+		return types.Topic{}, errorsmod.Wrapf(err, "error getting topic: %v", msg.TopicId)
+	}
+
+	if topic.Creator != msg.Sender {
+		return types.Topic{}, types.ErrNotPermittedToModifyTopic
+	}
+
+	if err := msg.Validate(params.MaxStringLength); err != nil {
+		return types.Topic{}, err
+	}
```
amimart: I'd rather remove that validation as the topic is validated before being stored

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -147,6 +147,10 @@ message EventCreateNewTopic {
   emissions.v3.Topic topic = 1;
 }
 
+message EventTopicUpdated {
+  uint64 topic_id = 1;
```
amimart, line 151: Should we add the topic as well? So that off chain nodes could directly use what's inside the event

amimart (commented):
Nice thx! Just left some suggestions

x/emissions/keeper/keeper.go
guilherme-brandao: good catch!

x/emissions/keeper/keeper.go
guilherme-brandao: I'm not sure about this behavior... It makes the experience ambiguous for the user. The transaction returns OK, but nothing is changed, the state stays the same. If the user accidentally sends an empty or invalid payload, it’s not obvious that nothing was actually applied.

xmariachi: I think noop is the right call. The update is successful in that it leaves the topic in the state that the request is stating, so in that sense, it complies with the contract.
I don't think we should treat an empty payload differently. (They may accidentally do that, but what they requested has been "done": they requested noop, they got noop )

CHANGELOG.md
amimart: I'm not sure it really brings a value to have a global section for all API changes, but it seems important to me to have a section dedicated to breaking changes. I'd add this change in the higher `### Added` section, and I guess the breaking changes `Added` section could be removed

guilherme-brandao: I'd at least keep the API breaking changes separately. I think it's important have it.

x/emissions/keeper/keeper.go
amimart: > it leaves the topic in the state that the request is stating

Yes I agree, it successfully changes nothing, in the same way as a field present in the update msg that is the same as the value already present

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -147,6 +147,10 @@ message EventCreateNewTopic {
   emissions.v3.Topic topic = 1;
 }
 
+message EventTopicUpdated {
+  emissions.v3.Topic topic = 1;
+}
```
amimart, line 153: When subscribing to the websocket or searching txs by events, in order to be able to filter on an event's field it must be a root field. I think it'd be great to have at least the topic id as root field to allow filtering on it

CHANGELOG.md
xmariachi: Yeah @amimart  @guilherme-brandao   `API Breaking Changes/Added` would never be API breaking because it would just be additive, so then it doesn't really makes sense - only `Changed` or `Removed` do.
So we could remove the `Added` subsection here.

x/emissions/keeper/keeper.go
```diff
@@ -3253,6 +3253,66 @@ func (k *Keeper) TopicExists(ctx context.Context, topicId TopicId) (bool, error)
 	return k.topics.Has(ctx, topicId)
 }
 
+// UpdateTopic applies allowed changes to a topic.
+func (k *Keeper) UpdateTopic(ctx context.Context, msg *types.UpdateTopicRequest) (types.Topic, error) {
```
amimart: The message should be the responsibility of the msg server layer, the Keeper should stays independent from the message. Maybe we should pass the topic with its fields already updated directly

xmariachi: Yep, `Request`s should not be in the keeper.

The `updatedTopic` can be built on the msg server, and being passed in to the `UpdateTopic(ctx context.Context, Topic updatedTopic)`  right?

x/emissions/keeper/keeper.go
```diff
@@ -3253,6 +3253,66 @@ func (k *Keeper) TopicExists(ctx context.Context, topicId TopicId) (bool, error)
...
+		return types.Topic{}, errorsmod.Wrapf(err, "error getting params for sender: %v", &msg.Sender)
+	}
+
+	topic, err := k.GetTopic(ctx, msg.TopicId)
+	if err != nil {
+		return types.Topic{}, errorsmod.Wrapf(err, "error getting topic: %v", msg.TopicId)
+	}
+
+	if topic.Creator != msg.Sender {
+		return types.Topic{}, types.ErrNotPermittedToModifyTopic
```
amimart: Maybe we can use the generic unauthorized error here:

```suggestion
		return types.Topic{}, sdkerrors.ErrUnauthorized
```

It's easier for clients to identify the underlying error when the code for a same error is consistent across modules

x/emissions/keeper/keeper.go
```diff
@@ -3253,6 +3254,62 @@ func (k *Keeper) TopicExists(ctx context.Context, topicId TopicId) (bool, error)
...
 }
 
+// UpdateTopic applies allowed changes to a topic.
+func (k *Keeper) UpdateTopic(ctx context.Context, sender ActorId, topic types.Topic, updatedTopic types.Topic) (types.Topic, error) {
+	params, err := k.GetParams(ctx)
+	if err != nil {
+		return types.Topic{}, errorsmod.Wrap(err, "error getting params")
+	}
+
+	if topic.Creator != sender {
```
xmariachi: This authorization layer should be better done in the msg server, prior to calling the keeper.

x/emissions/keeper/keeper.go
```diff
@@ -3253,6 +3254,62 @@ func (k *Keeper) TopicExists(ctx context.Context, topicId TopicId) (bool, error)
...
+		return types.Topic{}, errorsmod.Wrap(sdkerrors.ErrUnauthorized, "not permitted to modify topic")
+	}
+
+	// Only allow updates to specific fields, and keep all other topic fields unchanged.
+	newTopic := topic
```
xmariachi: isn't `updatedTopic` already the topic to update? So, I think there wouldn't be a need to create a `newTopic`, no?.

xmariachi (changes requested):
Small changes requested on `UpdateTopic` fn

x/emissions/proto/emissions/v9/tx.proto
```diff
@@ -319,6 +321,31 @@ message CreateNewTopicResponse {
...
+    (gogoproto.nullable) = false
+  ];
+  string merit_sortition_alpha = 6 [
+    (gogoproto.customtype) = "github.com/allora-network/allora-chain/math.Dec",
+    (gogoproto.nullable) = false
+  ];
+  string p_norm = 7 [
+    (gogoproto.customtype) = "github.com/allora-network/allora-chain/math.Dec",
+    (gogoproto.nullable) = false
+  ];
```
zale144, line 352: maybe leave a comment on why only these fields are updated

xmariachi: My 2ct here, I'm ot sure it is needed here in the proto. Each one has its own justification, so that could get too wordy. And also, justifications could be more meaningful into those topic fields that are _not_ modifiable - which are not defined here... so I'm ok not adding a comment here too.

xmariachi: Sounds like it could be added to proper documentation, eg the `emissions` README. Currently it is pretty thin, could be used for rationales like these.

x/emissions/keeper/keeper.go
```diff
@@ -3253,6 +3254,62 @@ func (k *Keeper) TopicExists(ctx context.Context, topicId TopicId) (bool, error)
...
+				return types.Topic{}, errorsmod.Wrap(err, "failed to get unfulfilled worker nonces")
+			}
+			if len(nonces.Nonces) > 0 {
+				lastNonce := nonces.Nonces[0]
+				withinWindow, err := BlockWithinWorkerSubmissionWindowOfNonce(topic, *lastNonce, blockHeight)
+				if err != nil {
+					return types.Topic{}, errorsmod.Wrap(err, "failed to check worker submission window")
+				}
+				if withinWindow {
+					return types.Topic{}, errorsmod.Wrap(types.ErrWorkerNonceWindowNotAvailable, "cannot update merit_sortition_alpha while worker window is open")
```
zale144, line 3287: `loss_method`, `p_norm`, and `alpha_regret` are updated immediately. Since EMA-based regrets/weights carry across epochs, changing these mid-topic can mix losses with different semantics. should these not be epoch-boundary gated or versioned like `merit_sortition_alpha`?

zale144: @xmariachi any thoughts on this?

xmariachi: They were considered to be safe, because they are being changed "for all actors at once". That is, the change (or not change) will impact every actor in an epoch - so it is fair. 
`merit_sortition_alpha` change could affect at tx time, so then this is why it is an exception.

amimart (dismissed):

x/emissions/keeper/msgserver/msg_server_params.go
```diff
@@ -126,9 +126,10 @@ func (ms msgServer) UpdateParams(ctx context.Context, msg *types.UpdateParamsReq
 	if len(newParams.CRewardForecast) == 1 {
 		existingParams.CRewardForecast = newParams.CRewardForecast[0]
 	}
-	if len(newParams.CNorm) == 1 {
-		existingParams.CNorm = newParams.CNorm[0]
-	}
+	// NOTE: `c_norm` is deprecated as a global param and is now per-topic (Topic.CNorm).
+	// We keep the field in protobuf types for backwards compatibility, but ignore it here.
+	// If a client attempts to set it via UpdateParams, it will be a no-op.
+	_ = newParams.CNorm
```
xmariachi: if this is done just to pass linter: can we add a //nolint directive instead somehow?

guilherme-brandao: It wasn’t for the linter, it’s just to make it explicit that c_norm is ignored. Just removed it!

xmariachi (approved):

amimart (approved):

spooktheducks (approved):

PR #834, opened 2025-06-23
feat: add event on outlier resistant network inferences

amimart:

Emit new `EventOutlierResistantNetworkInferences` event at the same moment we emit `EventNetworkInferences` event, but for outlier resistant inferences.


fstecker:
Does this really require a new event? Couldn't we just add two fields "OR naive" and "OR combined" to the existing network inference event?

xmariachi:
> Does this really require a new event? Couldn't we just add two fields "OR naive" and "OR combined" to the existing network inference event?

I think it makes it easier if we want to remove one of both in the future and stick to using one only, without needing schema modification here or in the indexer tables. Not a strong opinion here, though.

xmariachi (changes requested):
lgtm but needs rebasing

xmariachi (approved):

guilherme-brandao (approved):

PR #835, opened 2025-06-25
feat: design scheduler

amimart:

This PR introduces the new `x/scheduler` module with its first elements of design, especially regarding state management and exposed interfaces, along with its implementation.

The details about internals and usage of this module are present in the `./x/scheduler/README.md`

## TODO

- [x] Proper error management
- [x] Tests
- [x] ABCI BeginBlocker
- [x] Genesis export/import
- [x] App wiring
- [x] Implement missing Keeper methods
- [x] Register upgrade


x/scheduler/module.go
```diff
@@ -0,0 +1,100 @@
...
+	gs := am.keeper.ExportGenesis(ctx)
+	return cdc.MustMarshalJSON(&gs)
+}
+
+// ConsensusVersion implements HasConsensusVersion
+func (AppModule) ConsensusVersion() uint64 { return ConsensusVersion }
+
+// BeginBlock executes all ABCI BeginBlock logic respective to the scheduler module.
+func (am AppModule) BeginBlock(ctx context.Context) error {
+	defer telemetry.ModuleMeasureSince(types.ModuleName, telemetry.Now(), telemetry.MetricKeyEndBlocker)
```
guilherme-brandao: It should be `telemetry.MetricKeyBeginBlocker`

x/scheduler/proto/scheduler/v1/genesis.proto
```diff
@@ -0,0 +1,12 @@
...
+package scheduler.v1;
+
+import "gogoproto/gogo.proto";
+import "scheduler/v1/task.proto";
+
+option go_package = "github.com/allora-network/allora-chain/x/scheduler/types";
+
+// GenesisState defines the scheduler module's genesis state.
+message GenesisState {
+  repeated Task tasks = 1 [(gogoproto.nullable) = false];
```
xmariachi, line 11: here - don't we need to add the additional fields in the Scheduler Keeper? 
(`taskSchedule`, etc)

amimart: The task itself carries its scheduled run, the taskSchedule store is here only to index it, so when importing from genesis it'll be filled as well

x/scheduler/types/task.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// ScheduleEvery sets the task's interval between executions, if nil the task will execute once, otherwise the task is periodic.
+func ScheduleEvery(interval *time.Duration) SchedulingOption {
+	return func(_ sdk.Context, t *Task) {
+		t.Interval = interval
+	}
+}
+
+// Unschedule sets the task the next execution unscheduled, the task is kept for future execution but is considered paused.
```
xmariachi:
```suggestion
// Unschedule sets the task’s next execution to unscheduled. The task remains stored for future use but is considered paused.
```

x/scheduler/module.go
amimart: Oh good catch!

x/scheduler/keeper/genesis.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/allora-network/allora-chain/x/scheduler/types"
+)
+
+func (k *Keeper) InitGenesis(ctx context.Context, state types.GenesisState) {
+	for _, task := range state.Tasks {
+		if _, ok := k.handlersByTypename[task.Typename]; !ok {
+			panic(errors.Wrapf(types.ErrInvalidTask, "unknown task typename '%s'", task.Typename))
+		}
+
+		if err := k.tasks.Set(ctx, task.Id, task); err != nil {
```
zale144, line 16: if we're already passing the whole task, why bother passing the id separately?

x/scheduler/keeper/abci.go
```diff
@@ -0,0 +1,107 @@
...
+	case types.ArbitrageActionReschedule:
+		err = k.RescheduleTask(ctx, task, decision.RescheduleOpts...)
+	}
+	return
+}
+
+func (k *Keeper) runTask(ctx context.Context, task types.Task, handler types.TaskHandler) error {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+
+	if err := handler.Run(ctx, k.cdc, task.Id, task.Args, task.RunCount+1); err != nil {
```
zale144: could we not have just passed the whole task here as an argument?

xmariachi (dismissed):

x/scheduler/keeper/keeper.go
```diff
@@ -0,0 +1,250 @@
...
+			return errors.Wrapf(types.ErrInvalidTaskHandler, "task handler circular dependency over '%s'", handler.Typename())
+		}
+		if _, ok := added[typename]; ok {
+			return nil
+		}
+		visited[typename] = struct{}{}
+
+		for _, dep := range handler.DependsOn() {
+			if _, ok := k.handlersByTypename[dep]; !ok {
+				return errors.Wrapf(types.ErrInvalidTaskHandler, "unexisting dependency '%s' on task handler '%s'", dep, handler.Typename())
```
zale144:
```suggestion
				return errors.Wrapf(types.ErrInvalidTaskHandler, "nonexistent dependency '%s' on task handler '%s'", dep, handler.Typename())
```

x/scheduler/keeper/keeper.go
```diff
@@ -0,0 +1,250 @@
...
+			}
+
+			if err := addRec(dep); err != nil {
+				return err
+			}
+		}
+
+		delete(visited, typename)
+		added[typename] = struct{}{}
+		k.handlersOrder = append(k.handlersOrder, typename)
```
zale144, line 122: if we call `RegisterTaskHandlers` more than once, could we not end up with duplicates in `k.handlersOrder`? Why not first append to a local slice and set it to `k.handlersOrder`:

```
var newOrder []string
// append to newOrder instead of k.handlersOrder
k.handlersOrder = newOrder
```

xmariachi (dismissed):

x/scheduler/keeper/keeper.go
```diff
@@ -0,0 +1,250 @@
...
+
+		k.handlersByTypename[typename] = taskHandler
+		typenames = append(typenames, typename)
+	}
+
+	added := make(map[string]struct{}, len(typenames))
+	visited := make(map[string]struct{}, len(typenames))
+	var addRec func(string) error
+	addRec = func(typename string) error {
+		handler := k.handlersByTypename[typename]
```
zale144, line 101:
```suggestion
		handler, ok := k.handlersByTypename[typename]
        if !ok { 
            return errors.Errorf("unknown task handler: %q", typename) 
        }
```

x/scheduler/keeper/keeper.go
```diff
@@ -0,0 +1,250 @@
...
+			return nil
+		}
+		visited[typename] = struct{}{}
+
+		for _, dep := range handler.DependsOn() {
+			if _, ok := k.handlersByTypename[dep]; !ok {
+				return errors.Wrapf(types.ErrInvalidTaskHandler, "unexisting dependency '%s' on task handler '%s'", dep, handler.Typename())
+			}
+
+			if err := addRec(dep); err != nil {
```
zale144, line 115: are we putting a cap on the chain depth, to avoid stack overflow?

zale144 (changes requested):
nice work, left some comments

x/scheduler/keeper/genesis.go
amimart: Need to provide both key and value on the store

x/scheduler/keeper/abci.go
amimart: Yeah that would make sense, it initially felt inaccurate to me because typically the field `NextRunAt` refers to the current execution and not the next..

But maybe we can pass the whole task and solve this _inaccuracy_ by being more precise in the field naming, what do you think about renaming `NextRunAt` in `ScheduledAt`? So that it clearly refers to the current execution.

x/scheduler/keeper/keeper.go
amimart: Yes you're right, there's a comment on the method to indicate that it must be called once but I'll add the reset of both `k.handlersOrder` and `k.handlersByTypename` at the beginning

x/scheduler/keeper/keeper.go
amimart: Not sure it's needed it's something static once the binary has been compiled, it's not like it can be called through a tx, I consider it the responsibility of the dev using it

x/scheduler/keeper/abci.go
zale144: NextRunAt and ScheduledAt mean slightly different things. Scheduled at would be when the scheduling took place, rather than for when

x/scheduler/keeper/keeper.go
zale144: That makes sense, I didn't look at where it's being called. So, probably not an attack vector

x/scheduler/keeper/abci.go
zale144: Maybe it could be RunAt or ScheduledFor?

xmariachi: Agree that `ScheduledAt` is ambiguous since it seems to mean "when it was scheduled". 
So yep `RunAt`/ `ExecuteAt` sounds less ambiguous. If it's only about execution, I would keep reference to execution/run instead of to schedule.

amimart: Yeah agree this is ambiguous, I'd rather go for `ScheduledFor` to keep the _schedule_ verb, this would be consistent with the func used to set it: `func ScheduleAt(at time.Time) SchedulingOption`

xmariachi (approved):

x/scheduler/keeper/genesis.go
zale144: my bad, didn't realize it was the store

zale144 (approved):

PR #838, opened 2025-07-01
Remove recover from Begin/EndBlockers, align mint/emissions

xmariachi:

* Tested on a "poisoned" snapshot, runs as expected (showing the expected issue). 
* Ran the fuzzer tests with the changes on mint module and new invariants patch.
A fix, no need to document


amimart (commented):
I think the original issue was more related to the inter block cache enabled on rpc nodes, I'd rather let it panic as recovering and silencing it would make investigation more difficult

xmariachi:
> I think the original issue was more related to the inter block cache enabled on rpc nodes, I'd rather let it panic as recovering and silencing it would make investigation more difficult

Since it is a named return, it will return the error anyway. `return` = `return err` in that case.

guilherme-brandao:
I agree with @amimart, I’d rather let it panic here.

x/emissions/module/module.go
```diff
@@ -189,19 +189,11 @@ func (am AppModule) ExportGenesis(ctx sdk.Context, cdc codec.JSONCodec) json.Raw
...
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
-	defer func() {
-		if r := recover(); r != nil {
-			err := fmt.Errorf("error: %v", r)
-			sdkCtx.Logger().Error("Recover panic in EndBlocker", err)
-		}
-	}()
-
-	err := EndBlocker(ctx, am)
+	err = EndBlocker(ctx, am)
```
zale144: not sure why we need the named return, but no issue

x/mint/module/module.go
```diff
@@ -164,12 +164,13 @@ func (am AppModule) ExportGenesis(ctx sdk.Context, cdc codec.JSONCodec) json.Raw
 func (AppModule) ConsensusVersion() uint64 { return ConsensusVersion }
 
 // BeginBlock returns the begin blocker for the mint module.
-func (am AppModule) BeginBlock(ctx context.Context) error {
-	err := BeginBlocker(ctx, am.keeper)
+func (am AppModule) BeginBlock(ctx context.Context) (err error) {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	err = BeginBlocker(ctx, am.keeper)
```
zale144: same as above

zale144 (dismissed):

zale144 (approved):

guilherme-brandao (approved):

PR #839, opened 2025-07-03
Correct error handling in EndBlocker

spooktheducks:

## Still Left Todo

### DIFF
--- a/x/emissions/keeper/actor_utils/worker.go
+++ b/x/emissions/keeper/actor_utils/worker.go
@@ -3,7 +3,7 @@ package actorutils
 import (
 	"sort"
 
-	errorsmod "cosmossdk.io/errors"
+	"github.com/allora-network/allora-chain/errors"
 	keeper "github.com/allora-network/allora-chain/x/emissions/keeper"
 	synth "github.com/allora-network/allora-chain/x/emissions/keeper/inference_synthesis"
 	"github.com/allora-network/allora-chain/x/emissions/types"
@@ -19,7 +19,7 @@ func CloseWorkerNonce(k *keeper.Keeper, ctx sdk.Context, topic types.Topic, nonc
 	// Check if the nonce is unfulfilled
 	nonceUnfulfilled, err := k.IsWorkerNonceUnfulfilled(ctx, topic.Id, &nonce)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "failed: fetch is worker nonce unfulfilled")
 	}
 	// If the nonce is already fulfilled, return an error
 	if !nonceUnfulfilled {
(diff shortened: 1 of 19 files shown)


x/emissions/module/module.go
```diff
@@ -189,19 +190,21 @@ func (am AppModule) ExportGenesis(ctx sdk.Context, cdc codec.JSONCodec) json.Raw
...
-			sdkCtx.Logger().Error("Recover panic in EndBlocker", err)
-		}
-	}()
-
-	err := EndBlocker(ctx, am)
+	logger := sdkCtx.Logger().With(
+		"abci_step", "EndBlock",
+		"height", sdkCtx.BlockHeight(),
+	)
+	sdkCtx = sdkCtx.WithLogger(logger)
```
amimart, line 201: This is great, I think we should also systematically set the `module` logger field to allow proper filtering through configuration as it is done [here](https://github.com/cosmos/cosmos-sdk/blob/f5e0eeff6321f4a2a399aaffd0b149c51a892e72/x/distribution/keeper/keeper.go#L117)

x/emissions/keeper/actor_utils/worker.go
```diff
@@ -32,54 +32,12 @@ func CloseWorkerNonce(k *keeper.Keeper, ctx sdk.Context, topic types.Topic, nonc
...
 	// Get all active inferers for this topic
 	activeInfererAddresses, err := k.GetActiveInferersForTopic(ctx, topic.Id)
 	if err != nil {
-		return err
-	}
-	if len(activeInfererAddresses) == 0 {
-		return types.ErrNoQualifiedInferers
+		return errors.Wrap(err, "failed: fetch active inferrers for topic")
+	} else if len(activeInfererAddresses) == 0 {
+		return errors.Wrap(types.ErrNoQualifiedInferers, "no qualified inferrers")
```
amimart: I think this should not be an error as it seems to be a situation expected to happen, by returning an error for a normal behaviour we'll halt the chain

amimart (commented):
This seems like a good standard to adopt

However I think through this initiative we also need to carefully hunt down all the false-errors that are expected situations and non state breaking, as returning an error in an ABCI method will halt the chain.

Also as a side not, I think we should have less [error types](https://github.com/allora-network/allora-chain/blob/bd711259f2731693b4202c25f7b956dd0fa4cd39/x/emissions/types/errors.go#L9) as most of them denotes the same type but in different context

x/emissions/keeper/actor_utils/worker.go
```diff
@@ -112,31 +70,48 @@ func CloseWorkerNonce(k *keeper.Keeper, ctx sdk.Context, topic types.Topic, nonc
...
 	}
 
 	// Computes and stores both regular and outlier-resistant network inferences
 	err = ProcessAndStoreNetworkInferences(k, ctx, topic.Id, nonce.BlockHeight, activeInferences, activeForecasts)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "failed: process and store network inferences")
+	}
+
+	_, err = k.FulfillWorkerNonce(ctx, topic.Id, &nonce)
```
xmariachi, line 83: Moving this `defer`red block back to this place will not execute those methods if on error, which is not desired. We want them to run on any case.

x/emissions/module/abci.go
```diff
@@ -74,42 +74,41 @@ func EndBlocker(ctx context.Context, am AppModule) error {
...
-					continue
-				}
-				for _, nonce := range nonces.Nonces {
-					// No need to validate blockHeight boundaries - we accept submissions until this block.
-					sdkCtx.Logger().Debug("ABCI EndBlocker", "blockHeight", blockHeight, "closing worker window for topic", "topicId", topicId, "nonce", nonce)
-					err = allorautils.CloseWorkerNonce(&am.keeper, sdkCtx, topic, *nonce)
-					if err != nil {
-						sdkCtx.Logger().Info("Error closing worker nonce", "error", err)
-					}
+					return errors.WrapWithFields(err, "failed: close worker nonce", "topic_id", topicId)
```
xmariachi: If we want to return the error here, on any case of any error, then we need to carefully review all of the errors we currently have, and either classify them as unexpected (then halt) or recoverable, or just do not return and propagate such recoverable errors on those functions if they are expected to be handled.

This change would need careful review of our error handling within all Begin/EndBlocker code, so it'd be better done in a separate effort than the current fixes on BeginBlocker/EndBlocker.

xmariachi: And I think we would need to do that, for safety; because we are treating equally (= ignoring) both a business-as-usual error like "no inferers in this epoch" or a critical one like "no access to store".

x/emissions/keeper/actor_utils/worker.go
amimart: If we return the error on the end blocker it doesn't matter to execute these or not, the node will halt anyway

x/emissions/module/module.go
xmariachi: We actually [do it in mint module](https://github.com/allora-network/allora-chain/blob/bd711259f2731693b4202c25f7b956dd0fa4cd39/x/mint/keeper/keeper.go#L82) :D 

and [in the tests](https://github.com/allora-network/allora-chain/blob/bd711259f2731693b4202c25f7b956dd0fa4cd39/x/emissions/testutil/test_utils.go#L279), although here we use "submodules"

x/emissions/module/module.go
```diff
@@ -189,19 +190,21 @@ func (am AppModule) ExportGenesis(ctx sdk.Context, cdc codec.JSONCodec) json.Raw
 }
 
 // EndBlock returns the end blocker for the emissions module.
-func (am AppModule) EndBlock(ctx context.Context) error {
+func (am AppModule) EndBlock(ctx context.Context) (err error) {
```
xmariachi, line 193: Good one, I'll move this pattern to our current Begin/EndBlocker.

x/emissions/keeper/actor_utils/worker.go
xmariachi: right, i went from top-to-bottom :_D 
I put a comment below on that in that case any error we return means a halting error, so we need to be careful about what we return as error and what we handle within any function called from the BB/EB

PR #840, opened 2025-07-05
fix error in  testnet.go

aso20455:
Hey team, found and fixed simple error in `cmd/allorad/cmd/testnet.go`

`valdiators` - `validators`


spooktheducks:
Incorporated in #914

PR #850, opened 2025-08-13
Add missing events

zale144:

- [x] If tested, please describe how. If not, why tests are not needed. - probably not needed to test event emissions
- [x] If documented, please describe where. If not, describe why docs are not needed. - not needed


x/emissions/proto/emissions/v9/events.proto
```diff
@@ -62,6 +67,322 @@ message EventNetworkInferences {
...
+message EventAddReputerStake {
+  uint64 topic_id = 1;
+  string reputer = 2;
+  string amount = 3 [
+    (cosmos_proto.scalar) = "cosmos.Int",
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.nullable) = false,
+    (amino.dont_omitempty) = true
+  ];
+  string topic_stake = 4 [
```
amimart: I'd remove that as we should be able to compute it by replaying stake related events

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -3,7 +3,12 @@ package emissions.v9;
 
 import "emissions/v3/nonce.proto";
 import "emissions/v3/reputer.proto";
+import "emissions/v3/worker.proto";
+import "emissions/v3/topic.proto";
+import "emissions/v8/params.proto";
 import "gogoproto/gogo.proto";
+import "amino/amino.proto";
```
xmariachi: Why these new protos are used for? 
And if they are used here, should they be used in other parts too for alignment?

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -62,6 +67,322 @@ message EventNetworkInferences {
...
+
+message EventCreateNewTopic {
+  emissions.v3.Topic topic = 1;
+}
+
+message EventAddTopicFeeRevenue {
+  uint64 topic_id = 1;
+  string amount = 2 [
+    (cosmos_proto.scalar) = "cosmos.Int",
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
```
xmariachi: 1. Why these new protos are used? If they are used here, should they be used in other parts too for alignment?

2. Since this is math.Int, I fear this could overflow - shouldn't our Dec be used instead?

xmariachi (commented):
Lgtm in general, just two comments on protos

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -62,6 +67,322 @@ message EventNetworkInferences {
...
+  uint64 topic_id = 1;
+  string reputer = 2;
+  string amount = 3 [
+    (cosmos_proto.scalar) = "cosmos.Int",
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.nullable) = false,
+    (amino.dont_omitempty) = true
+  ];
+  int64 block_removal_started = 4;
+  int64 block_removal_completed = 5;
```
amimart: I think this should be deduced

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -62,6 +67,322 @@ message EventNetworkInferences {
...
+message EventInsertWorkerPayload {
+  uint64 topic_id = 1;
+  emissions.v3.WorkerDataBundle worker_data_bundle = 2;
+}
+
+message EventCreateNewTopic {
+  emissions.v3.Topic topic = 1;
+}
+
+message EventAddTopicFeeRevenue {
```
amimart: If we already have events for topic funding, registration and payload submission I don't see the value of this one

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -62,6 +67,322 @@ message EventNetworkInferences {
   emissions.v3.ValueBundle value_bundle = 3;
 }
 
+message EventInsertWorkerPayload {
+  uint64 topic_id = 1;
+  emissions.v3.WorkerDataBundle worker_data_bundle = 2;
```
amimart: Do we have an idea of how big data bundles can be? I feel like this could bring some overhead on the network as I've seen in the past some blocks with 5000 submission txs.

I'm always in favour of putting everything needed in events to recompose the state only relying on events, but this could be an exception.

I'm just putting a warning, not saying we shouldn't go with it, but running some benchmark on that could be justified

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -62,6 +67,322 @@ message EventNetworkInferences {
...
+    (cosmos_proto.scalar) = "cosmos.Int",
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.nullable) = false,
+    (amino.dont_omitempty) = true
+  ];
+  int64 block_removal_started = 4;
+  int64 block_removal_completed = 5;
+}
+
+message EventCancelRemoveReputerStake {
```
amimart: Do we need events for when a wallet ask for stake removal or to cancel stake removal? To observe stake evolution I think we only need two events: when stake is added and when stake is effectively removed, I don't see the value tracking internal stake removal scheduling

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -62,6 +67,322 @@ message EventNetworkInferences {
...
+    (cosmos_proto.scalar) = "cosmos.Int",
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.nullable) = false,
+    (amino.dont_omitempty) = true
+  ];
+  int64 block_removal_started = 4;
+  int64 block_removal_completed = 5;
+}
+
+message EventAddDelegateStake {
```
amimart: Currently we have two staking msg on reputers, one for the reputer itself and one for others but it's the same thing, I think we should have only one set of events for both. It's even easier to manage on the indexer side.

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -62,6 +67,322 @@ message EventNetworkInferences {
...
+  uint64 topic_id = 1;
+  string address = 2;
+}
+
+message EventTopicAdminWhitelistRemoved {
+  uint64 topic_id = 1;
+  string address = 2;
+}
+
+message EventTopicAdminWhitelistBulkAdded {
```
amimart: I think we should have a single event associated to a specific state mutation, bulk operations just run the same mutation multiple times which should emit multiple events, not a different one imho

amimart (commented):
Thanks!

I've made a few comments related to mainly two concerns:

1. We should dissociate state mutation and tx msg: different msgs can lead to the same mutation, but we shouldn't have an event for each one of them.

2. Events are not free: we should avoid unnecessary data, duplication and very large payloads.

These are only my opinion, I'd be happy to have yours and discuss that

x/emissions/proto/emissions/v9/events.proto
zale144: not sure if you mean you see or you don't see the value of this one?

x/emissions/proto/emissions/v9/events.proto
zale144: we already have NetworkInferences and NetworkLoss, which are I think even bigger 🤷‍♂️

x/emissions/proto/emissions/v9/events.proto
zale144: so you would rather emit an event for each address added or removed, instead of one event per bulk? It might be a lot of events, not sure

x/emissions/proto/emissions/v9/events.proto
zale144: they are used in other places as well. `amino` is used for when `(amino.dont_omitempty) = true` is needed

x/emissions/proto/emissions/v9/events.proto
zale144: it's the same type as the value being passed. check `(k *Keeper) AddTopicFeeRevenue`. I'd link it here, but the file is too big 😁

x/emissions/proto/emissions/v9/events.proto
amimart: To be honest I don't think having bulk msg is a good practice as we can just put multiple message in a tx.
Maybe a compromise can be found with a single event type having a repeated field.

x/emissions/proto/emissions/v9/events.proto
amimart: I don't see the value of this one, I think we already have the needed info through other events, to be checked though

x/emissions/proto/emissions/v9/events.proto
amimart: Yeah but these are emitted only once per epoch.

I think having the whole data bundle is not necessary, for example we don't need the signature and the public key in the event as at this point it is already verified. We only need the sending address, topic id/nonce and the inference value or forecast elements

x/emissions/proto/emissions/v9/events.proto
amimart: Yeah and the cosmos Int type is 256bit so I'd expect the Dec to overflow before it

x/emissions/proto/emissions/v9/events.proto
zale144: removed

x/emissions/proto/emissions/v9/events.proto
zale144: removed

x/emissions/proto/emissions/v9/events.proto
zale144: removed

x/emissions/proto/emissions/v9/events.proto
zale144: reworked

x/emissions/proto/emissions/v9/events.proto
zale144: removed

x/emissions/proto/emissions/v9/events.proto
zale144: removed

x/emissions/proto/emissions/v9/events.proto
zale144: removed, done with events per address

x/emissions/proto/emissions/v9/events.proto
zale144: will be addressed in @guilherme-brandao's  PR

x/emissions/proto/emissions/v9/events.proto
xmariachi: ok, the same type should be used, I assumed it was receiving a Dec. 
(However Dec would not overflow before it, it has a `MaxExponent = 100000`... although our input layer's Dec only 40 iirc)

amimart (dismissed):

x/emissions/proto/emissions/v9/query.proto
```diff
@@ -555,6 +540,18 @@ service QueryService {
...
+  rpc GetWorkerSubmissionWindowStatus(GetWorkerSubmissionWindowStatusRequest) returns (GetWorkerSubmissionWindowStatusResponse) {
+    option (cosmos.query.v1.module_query_safe) = true;
+    option (google.api.http).get = "/emissions/v9/worker_submission_window_status/{topic_id}";
+  }
+
+  // Get reputer submission window status and timing
+  rpc GetReputerSubmissionWindowStatus(GetReputerSubmissionWindowStatusRequest) returns (GetReputerSubmissionWindowStatusResponse) {
+    option (cosmos.query.v1.module_query_safe) = true;
+    option (google.api.http).get = "/emissions/v9/reputer_submission_window_status/{topic_id}";
+  }
```
amimart, line 554: We can have multiple nonce open at the same time so I think we should query the window status for a nonce, not only a topic.

Also maybe we could have a single query to provide both worker and reputer window information

xmariachi: Discussed already, but for the record. 
1. Yes - the purpose of this query is tailored to actor usage, which usually wants to know the latest nonce to action upon, and any subsequent or next nonce. 
2. "a single query to provide both worker and reputer window information" = this makes complete sense, a query informing the lifecycle of a particular nonce (at which point the nonce is, or at which point broke - eg if there worker nonce closed but there were no inferences, then the reputer does not get created - this could be informed as the lifecycle for the nonce, provided the nonce is available). Created ticket about it.

xmariachi (approved):
all my concerns are tackled :)

amimart (approved):
🚀

PR #851, opened 2025-08-14
feat: enforce consensus settings

amimart:

Define a bunch of recommended settings for both `app.toml` and `config.toml` and enforce them upon start-up.

These settings are automatically set as default when using the `init` command, and if changed they are overridden before loading the application.

In order to still allow their modification a `--unsafe-override-recommended-settings` flag has been added to the `start` command to bypass the enforcing logic and let the values provided in the `config.toml` and `app.toml` files.

The recommended settings to enforce that are set through this PR are the following:
- `consensus.timeout_propose`: 5s
- `consensus.timeout_propose_delta`: 500ms
- `consensus.timeout_prevote`: 3s
- `consensus.timeout_prevote_delta`:  500ms
- `consensus.timeout_precommit`: 3s
- `consensus.timeout_precommit_delta`: 500ms
- `consensus.timeout_commit`: 5s

### DIFF
--- a/cmd/allorad/cmd/config.go
+++ b/cmd/allorad/cmd/config.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"errors"
 	"time"
 
 	cmtcfg "github.com/cometbft/cometbft/config"
@@ -46,10 +47,13 @@ var (
 
 type settings []setting
 
-func (s settings) setInMap(cfg map[string]interface{}) {
+func (s settings) setInMap(cfg map[string]interface{}) error {
 	for _, setting := range s {
-		setting.setInMap(cfg)
+		if err := setting.setInMap(cfg); err != nil {
+			return err
+		}
 	}
+	return nil
 }
 
 func (s settings) setInViper(v *viper.Viper) {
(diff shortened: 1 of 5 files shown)


cmd/allorad/cmd/config.go
```diff
@@ -0,0 +1,134 @@
...
+	return nil
+}
+
+func (s setting) setInViper(v *viper.Viper) {
+	v.Set(s.section+"."+s.key, s.value)
+}
+
+func mustGetDefaultConfigs() (*serverconfig.Config, *cmtcfg.Config) {
+	srvCfg := serverconfig.DefaultConfig()
+	srvCfg.MinGasPrices = "10uallo"
```
xmariachi: Possibly this does not have an effect anymore, since we set it through feemarket; however it will force a min if eg in testing the feemarket value is set to 0.

xmariachi (commented):
left a comment. in general lgtm. Initially I agree on the set of parameters to fix.

cmd/allorad/cmd/config.go
amimart: Yeah just changed the default value to 0 that makes sense

xmariachi (dismissed):

guilherme-brandao (dismissed):

xmariachi (approved):

guilherme-brandao (approved):

PR #853, opened 2025-08-18
Added more ABCI related events

guilherme-brandao:
Added these events:

### **1. Stake Removal Events**
1. **`EmitReputerStakeRemovalCompletedEvent`**
   - Tracks when reputer stake removal is completed
   - Parameters: `topicId`, `blockHeight`, `reputer`, `amount`

2. **`EmitDelegateStakeRemovalCompletedEvent`** 
   - Tracks when delegate stake removal is completed
   - Parameters: `topicId`, `blockHeight`, `delegator`, `reputer`, `amount`

### **2. Delegate Reward Events**
3. **`EmitDelegateRewardShareUpdatedEvent`**
   - Tracks when delegate reward share percentage is updated
   - Parameters: `topicId`, `reputer`, `rewardPerShare`, `blockHeight`

4. **`EmitDelegateRewardDistributedEvent`**
   - Tracks when delegate rewards are distributed to pending account
   - Parameters: `topicId`, `reputer`, `amount`, `blockHeight`

### **3. Active Participants Tracking Events**
5. **`EmitActiveReputersSetEvent`**
   - Tracks active reputers for a topic/nonce during loss calculation
   - Parameters: `topicId`, `nonceBlockHeight`, `addresses`

6. **`EmitActiveInferersSetEvent`**
   - Tracks active inferers for a topic/nonce during worker nonce closing
   - Parameters: `topicId`, `nonceBlockHeight`, `addresses`

7. **`EmitActiveForecastersSetEvent`**
   - Tracks active forecasters for a topic/nonce during worker nonce closing
   - Parameters: `topicId`, `nonceBlockHeight`, `addresses`

8. **`EmitActiveTopicsAtBlockSetEvent`**
   - Tracks which topics are scheduled to be active at a future block
   - Parameters: `targetBlockHeight`, `topicIds`

### **4. Network Inference Weight Events**
9. **`EmitNetworkInferenceInfererWeightSetEvent`**
   - Tracks inferer weights during network inference calculation
   - Parameters: `topicId`, `nonceBlockHeight`, `address`, `weight`

10. **`EmitNetworkInferenceForecasterWeightSetEvent`**
    - Tracks forecaster weights during network inference calculation
    - Parameters: `topicId`, `nonceBlockHeight`, `address`, `weight`

### **5. Network Inference Regret Events**
11. **`EmitNetworkInferenceInfererRegretUsedEvent`**
    - Tracks regret values used for inferers during network inference calculation
    - Parameters: `topicId`, `nonceBlockHeight`, `address`, `regret`

12. **`EmitNetworkInferenceForecasterRegretUsedEvent`**
    - Tracks regret values used for forecasters during network inference calculation
    - Parameters: `topicId`, `nonceBlockHeight`, `address`, `regret`

### **6. Topic Weight Events**
13. **`EmitTopicWeightUpdatedEvent`**
    - Tracks when topic weights are updated
    - Parameters: `topicId`, `newWeight`

### **7. Submission Window Events**
14. **`EmitWorkerSubmissionWindowOpenedEvent`**
    - Tracks when worker submission windows are opened
    - Parameters: `topicId`, `nonceBlockHeight`, `windowStartBlock`, `windowEndBlock`

15. **`EmitWorkerSubmissionWindowClosedEvent`**
    - Tracks when worker submission windows are closed
    - Parameters: `topicId`, `nonceBlockHeight`

16. **`EmitReputerSubmissionWindowOpenedEvent`**
    - Tracks when reputer submission windows are opened
    - Parameters: `topicId`, `nonceBlockHeight`, `windowStartBlock`, `windowEndBlock`

17. **`EmitReputerSubmissionWindowClosedEvent`**
    - Tracks when reputer submission windows are closed
    - Parameters: `topicId`, `nonceBlockHeight`


x/emissions/proto/emissions/v9/events.proto
```diff
@@ -524,3 +524,78 @@ message EventPruneRecords {
...
+message EventReputerStakeRemovalCompleted {
+  uint64 topic_id = 1;
+  string reputer = 2;
+  string amount = 3 [
+    (cosmos_proto.scalar) = "cosmos.Int",
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.nullable) = false,
+    (amino.dont_omitempty) = true
+  ];
+  int64 block_height = 4;
```
amimart:
```suggestion
```

We can remove it from every events, events are already linked to a single block

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -436,3 +434,146 @@ message EventPruneRecords {
   int64 block_height = 1;
   uint64 topic_id = 2;
 }
+
+// Stake removal processing events
+message EventReputerStakeRemovalCompleted {
```
amimart: The event related to stake added is `EventAddStake` and is the same whether it is the reputer or any other delegator, I think the same logic should be applied here to avoid duplicating this event.

For consistency sake I'd align the naming with `EventRemoveStake` for example

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -436,3 +434,146 @@ message EventPruneRecords {
...
+  string amount = 4 [
+    (cosmos_proto.scalar) = "cosmos.Int",
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.nullable) = false,
+    (amino.dont_omitempty) = true
+  ];
+  int64 block_height = 5;
+}
+
+message EventDelegateRewardShareUpdated {
```
amimart, line 451: Is it useful to track that? Being curious because I'd consider this as an internal thing only

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -436,3 +434,146 @@ message EventPruneRecords {
...
+  uint64 topic_id = 1;
+  string reputer = 2;
+  string reward_per_share = 3 [
+    (gogoproto.customtype) = "github.com/allora-network/allora-chain/math.Dec",
+    (gogoproto.nullable) = false
+  ];
+  int64 block_height = 4;
+}
+
+message EventDelegateRewardDistributed {
```
amimart, line 461: Actually these rewards are not distributed but sent to a pending module account, is that what we want to track?

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -436,3 +434,146 @@ message EventPruneRecords {
...
+message EventActiveForecastersSet {
+  uint64 topic_id = 1;
+  repeated string addresses = 2;
+  int64 nonce_block_height = 3;
+}
+
+message EventActiveTopicsAtBlockSet {
+  int64 target_block_height = 1;
+  repeated uint64 topic_ids = 2;
+}
```
amimart: What's the value of exposing that as an event? Capturing nonce related events seems enough to track epochs lifecycle

x/emissions/keeper/keeper.go
```diff
@@ -3273,7 +3284,14 @@ func (k *Keeper) SetBlockToActiveTopics(ctx context.Context, block BlockHeight,
...
 		}
 	}
-	return k.blockToActiveTopics.Set(ctx, block, topicIds)
+	err := k.blockToActiveTopics.Set(ctx, block, topicIds)
+	if err != nil {
+		return err
+	}
+
+	// Emit event for active topics at block set
+	types.EmitActiveTopicsAtBlockSetEvent(ctx, block, topicIds.TopicIds)
```
amimart: If this event really needs to be emitted some rework is needed be cause this function can be called multiple times for the same target block

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -436,3 +434,146 @@ message EventPruneRecords {
...
+  string new_weight = 2 [
+    (gogoproto.customtype) = "github.com/allora-network/allora-chain/math.Dec",
+    (gogoproto.nullable) = false
+  ];
+}
+
+message EventWorkerSubmissionWindowOpened {
+  uint64 topic_id = 1;
+  int64 nonce_block_height = 2;
+  int64 window_start_block = 3;
```
amimart: If we just opened the window the start height should be the current one right?

x/emissions/keeper/actor_utils/worker.go
```diff
@@ -116,11 +122,23 @@ func CloseWorkerNonce(k *keeper.Keeper, ctx sdk.Context, topic types.Topic, nonc
...
 	}
 
+	// Emit reputer submission window opened event
+	extraLag := topic.GroundTruthLag % topic.EpochLength
+	if extraLag != 0 {
+		extraLag = topic.EpochLength - extraLag
+	}
+	windowStartBlock := nonce.BlockHeight + topic.GroundTruthLag
+	windowEndBlock := nonce.BlockHeight + topic.GroundTruthLag + extraLag + topic.EpochLength
+	types.EmitReputerSubmissionWindowOpenedEvent(ctx, topic.Id, nonce.BlockHeight, windowStartBlock, windowEndBlock)
```
amimart: At this point the window is not opened yet right?

x/emissions/keeper/actor_utils/losses.go
```diff
@@ -100,6 +100,9 @@ func CloseReputerNonce(
 		}
 
 		ctx.Logger().Info("Closed reputer nonce", "topicId", topic.Id, "nonce", nonce)
+
+		// Emit reputer submission window closed event
+		types.EmitReputerSubmissionWindowClosedEvent(ctx, topic.Id, nonce.BlockHeight)
```
amimart, line 105: This is executed only when we close the reputer nonce but I guess the window could be closed without closing the reputer nonce if the topic goes inactive, is that right?

amimart (commented):
Made some comments, mostly questions :)

I feel there's some inconsistencies in events naming, maybe we should decide setup a naming rule

x/emissions/keeper/actor_utils/losses.go
xmariachi: That's a good question. However if the topic gets inactive, it's true that it is no longer possible to send submissions, but also that the submissions sent will not get "processed". 

I think the question is about: do we want this event to mark "the closing of a valid window that has been processed" or "the closing of any window, even if won't be processed". 
The current implementation will serve the former. The decision would depend on what is the usage we plan to have on clients. Event-based clients should also listen to in/activation events, so I think this is OK.
any thoughts? cc @spooktheducks 

(btw, not for this pr because it may break consensus, but shouldn't this `defer` block should be run on top of the method instead of in the middle?)

x/emissions/keeper/actor_utils/worker.go
```diff
@@ -72,6 +72,9 @@ func CloseWorkerNonce(k *keeper.Keeper, ctx sdk.Context, topic types.Topic, nonc
 		}
 
 		ctx.Logger().Info("Closed worker nonce", "topicId", topic.Id, "nonce", nonce)
+
+		// Emit worker submission window closed event
+		types.EmitWorkerSubmissionWindowClosedEvent(ctx, topic.Id, nonce.BlockHeight)
```
xmariachi, line 77: This is similar to the above case on close of reputer windows. 

(and also, similar comment on the `defer` block too)

x/emissions/keeper/actor_utils/worker.go
```diff
@@ -96,6 +99,9 @@ func CloseWorkerNonce(k *keeper.Keeper, ctx sdk.Context, topic types.Topic, nonc
 		return err
 	}
 
+	// Emit event for active inferers set for topic nonce
+	types.EmitActiveInferersSetEvent(ctx, topic.Id, nonce.BlockHeight, activeInfererAddresses)
```
xmariachi: For this and the forecasters below: 
we have been following a pattern of emitting the events at the end of the function, that is, they will be emitted only if the function is successful. 

Example: `EmitNewWorkerLastCommitSetEvent` at the end of this function.
I think we should keep that pattern here too.

x/emissions/keeper/actor_utils/worker.go
xmariachi: Agreed.
This is a bit tricky - we may need to add new iteration in the endBlocker, in the active topics due this block, to check for a reputer nonce that is due to be open this block. 
We have the "skimmed topics" iteration already which can potentially be reused, but I think for this one we would need the whole set of topics due this block - but this latter point I'm not totally sure.

x/emissions/keeper/ema_scores.go
```diff
@@ -150,7 +150,7 @@ func (k *Keeper) CalcAndSaveInfererScoreEmaWithLastSavedTopicQuantile(
 
 	emaScores := []types.Score{emaScore}
 	activeArr := map[string]bool{previousInfererScore.Address: false}
-	types.EmitNewActorEMAScoresSetEvent(ctx, types.ActorType_ACTOR_TYPE_INFERER_UNSPECIFIED, emaScores, activeArr)
+	types.EmitNewActorEMAScoresSetEvent(ctx, types.ActorType_ACTOR_TYPE_INFERER_UNSPECIFIED, block, emaScores, activeArr)
```
xmariachi, line 153: are these removed from current `dev` or from Alek's pr? I mean it due to the problem of nonadditive proto modifications

x/emissions/keeper/inference_synthesis/network_inferences.go
```diff
@@ -236,6 +236,9 @@ func GetCalcNetworkInferenceArgs(
 
 		logger.Debug("Inferer has regret", "inferer", inferer, "regret", regret.Value)
 		infererToRegret[inferer] = &regret.Value
+
+		// Emit event for inferer regret used in network inference calculation
+		emissions.EmitNetworkInferenceInfererRegretUsedEvent(ctx, topicId, inferenceBlockHeight, inferer, regret.Value)
```
xmariachi: Maybe we should emit them as a set as we do with other closed "sets"? Or is there any particular reason why not?

x/emissions/keeper/inference_synthesis/network_inferences.go
```diff
@@ -247,6 +250,9 @@ func GetCalcNetworkInferenceArgs(
 
 		logger.Debug("Forecaster has regret", "forecaster", forecaster, "regret", regret.Value)
 		forecasterToRegret[forecaster] = &regret.Value
+
+		// Emit event for forecaster regret used in network inference calculation
+		emissions.EmitNetworkInferenceForecasterRegretUsedEvent(ctx, topicId, inferenceBlockHeight, forecaster, regret.Value)
```
xmariachi: same here

x/emissions/keeper/keeper.go
xmariachi: True. What is the use case for this event?

x/emissions/proto/emissions/v9/events.proto
xmariachi: That sounds correct

xmariachi (commented):
Several comments

x/emissions/proto/emissions/v9/events.proto
guilherme-brandao: It is used to calculate user's outstanding rewards

x/emissions/proto/emissions/v9/events.proto
guilherme-brandao: It’s distributed to an account where each user has an allowance to claim their rewards. Not sure if it’s the best naming, though. But yeah, we want to track the exact number of rewards directed to the reputer’s delegators.

x/emissions/keeper/keeper.go
guilherme-brandao: With this event, I can track the exact active topics that will be active in a specific block. The idea to add it here is that since the active topics set can change by block (topics activations and deactivations), we can know in real time which are the eligible topics to be processed in a specific block in the future.

x/emissions/proto/emissions/v9/events.proto
guilherme-brandao: Answered in the question below

x/emissions/keeper/actor_utils/losses.go
guilherme-brandao: Agreed! But I think there’s value in emitting all the closing window events, even if the topic is not active at the moment.

Regarding the `defer`, we need to investigate if we can do that. We would be fulfilling reputer nonces automatically, even though the worker/reputer nonce has not yet been fulfilled.

x/emissions/keeper/actor_utils/worker.go
guilherme-brandao: agreed!

x/emissions/keeper/keeper.go
xmariachi: 1. Why exactly do we need to track exact active topics for a block? (what is the actual Studio use case, i mean - trying to think if we can service it in some other way)

2. can we get a similar information via topic activation and inactivation events, by using its block + epochLength? (and keep on multiples until new inactivation events happen).

x/emissions/keeper/actor_utils/losses.go
xmariachi: Thinking more about it - I think the current impl is the consistent one. 

Inactivation of a topic does not remove the open, or pending, nonces. Basically it does not interact with nonces. So those nonces are open - eg if a topic gets inactivated at block b, and reactivates at b +1 , all within the same epoch, those reputer nonces (open or pending) would still get serviced - so that "window close" event would have been premature.

So essentially it wouldn't be correct now to send closing window on inactivation, because the window is not really closed, because the nonce is not fulfilled/removed.

The apparent problem would be that if a topic gets inactivated and does not get activated back within the current epoch, then the "window close" event is never sent. But the inactivation of topic event is  - and the client should still listen to that event type (topic in/activation) to stop/resume its work.

Hope that makes sense.

x/emissions/keeper/ema_scores.go
xmariachi: i thought proto was modified - my bad, resolving.

x/emissions/keeper/keeper.go
guilherme-brandao: 1 - This one was suggested in a meeting with the team about the new topics but it doesn't have a specific use case atm. 

2 - I mean, the topic activation and inactivation will have the same effect but we would need to send multiple events, but probably it's a more user-friendly way of doing this.

fstecker:
Hey I have a few comments on the event stuff, but I'm not sure if they belong here or on Alek's PR (I didn't look at the diffs but at the resulting code after merging both PRs):

* Some of the functions are called `EmitXEvent` and some are `EmitNewXEvent`?
* `OutlierRestitantNetworkInferencesEvent` has `topicId` and `blockHeight` fields, but `NetworkInferencesEvent` doesn't?
* Maybe `ForecastTaskUtilityScoreSetEvent` should also contain a "nonce" field (like, which epoch has caused this update in the utility score). Not that we need it at the moment, but might be useful in the future. Only if it's cheap to include.
* `XWeightSetEvent`, `NetworkInferenceXWeightSetEvent` and `NetworkInferenceXRegretUsedEvent` seem to emit one event per worker, while `XScoresSetEvent`, `ListeningCoefficientsEvent`, `XNetworkRegretSetEvent` emit them in a "packed" form. I feel like we should always use the packed form unless there is some reason against it.
* In the `FundTopicEvent`, can we include the total "dripped topic revenue" or whatever it's called, after the new funding is added? If it's cheap to do.
* Similarly, in the `TopicWeightUpdatedEvent` it would be cool to also include the two constituents of the topic weight, reputer stakes and topic revenue, if that's available cheaply.
* Are we planning to replace `emission_info` in `query_results`, which is currently provided by the producer querying once a minute? There are a few events which seem to try to do this (`RewardCurrentBlockEmissionEvent`, `RecalculateTargetEmissionEvent`, `PreviousPercentageRewardToStakedReputerSetEvent`), but it's very incomplete. How do you guys feel about one event, emitted every block, which contains all the fields that are currently in `emission_info`?

fstecker:
I made this list of all events after this and Alek's PR are merged. Using the signature of the `Emit(New)?XEvent` functions. If the argument types are anything except simple strings or numbers, I also wrote the type. The categories are a little arbitrary. But anyway, I found it useful as an overview, so I'll share it here.

### Mint/tokenomics
1. `TokenomicsSetEvent`
    stakedTokenAmount, circulatingAmount, emissionsAmount
1. `EcosystemTokenMintSetEvent`
    blockHeight, amount
1. `RewardCurrentBlockEmissionEvent`
    blockHeight, amount
1. `RecalculateTargetEmissionEvent`
    emissionPerUnitStakedToken, emissionPerMonth, blockEmission
1. `PreviousPercentageRewardToStakedReputersSetEvent`
    blockHeight, percentage

### Whitelist
1. `WhitelistAdminAddedEvent`
    address
1. `WhitelistAdminRemovedEvent`
    address
1. `GlobalWhitelistAddedEvent`
    address
1. `GlobalWhitelistRemovedEvent`
    address
1. `GlobalWorkerWhitelistAddedEvent`
    address
1. `GlobalWorkerWhitelistRemovedEvent`
    address
1. `GlobalReputerWhitelistAddedEvent`
    address
1. `GlobalReputerWhitelistRemovedEvent`
    address
1. `GlobalAdminWhitelistAddedEvent`
    address
1. `GlobalAdminWhitelistRemovedEvent`
    address
1. `TopicWorkerWhitelistEnabledEvent`
    topicId
1. `TopicWorkerWhitelistDisabledEvent`
    topicId
1. `TopicReputerWhitelistEnabledEvent`
    topicId
1. `TopicReputerWhitelistDisabledEvent`
    topicId
1. `TopicCreatorWhitelistAddedEvent`
    address
1. `TopicCreatorWhitelistRemovedEvent`
    address
1. `TopicWorkerWhitelistAddedEvent`
    topicId, address
1. `TopicWorkerWhitelistRemovedEvent`
    topicId, address
1. `TopicReputerWhitelistAddedEvent`
    topicId, address
1. `TopicReputerWhitelistRemovedEvent`
    topicId, address

### Submission windows
1. `WorkerSubmissionWindowOpenedEvent`
    topicId, nonceBlockHeight, windowStartBlock, windowEndBlock
1. `WorkerSubmissionWindowClosedEvent`
    topicId, nonceBlockHeight
1. `ReputerSubmissionWindowOpenedEvent`
    topicId, nonceBlockHeight, windowStartBlock, windowEndBlock
1. `ReputerSubmissionWindowClosedEvent`
    topicId, nonceBlockHeight

### Network inference / loss
1. `NetworkInferencesEvent`
    networkInferences `ValueBundle`
1. `OutlierResistantNetworkInferencesEvent`
    topicId, blockHeight, networkInferences `ValueBundle`
1. `RegretStdNormSetEvent`
    topicId, blockHeight, stdNorm
1. `NetworkLossSetEvent`
    lossBundle `ValueBundle`
1. `ForecastTaskUtilityScoreSetEvent`
    topicId, score

### Worker scores/regrets/weights
1. `InfererScoresSetEvent`
    scores `[]Score`
1. `ForecasterScoresSetEvent`
    scores `[]Score`
1. `ReputerScoresSetEvent`
    scores `[]Score`
1. `InfererWeightSetEvent`
    topicId, blockHeight, address, weight
1. `ForecasterWeightSetEvent`
    topicId, blockHeight, address, weight
1. `NetworkInferenceInfererWeightSetEvent`
    topicId, blockHeight, address, weight
1. `NetworkInferenceForecasterWeightSetEvent`
    topicId, blockHeight, address, weight
1. `NetworkInferenceInfererRegretUsedEvent`
    topicId, blockHeight, address, regret
1. `NetworkInferenceForecasterRegretUsedEvent`
    topicId, blockHeight, address, regret
1. `ListeningCoefficientsSetEvent`
    actorType, topicId, blockHeight, addresses `[]string`, coefficients `[]Dec`
1. `InfererNetworkRegretSetEvent`
    topicId, blockHeight, addresses `[]string`, regrets `[]Dec`
1. `ForecasterNetworkRegretSetEvent`
    topicId, blockHeight, addresses `[]string`, regrets `[]Dec`
1. `NaiveInfererNetworkRegretSetEvent`
    topicId, blockHeight, addresses `[]string`, regrets `[]Dec`

### Sortition
1. `ActorEMAScoresSetEvent`
    actorType, nonceBlockHeight, scores `[]Score`, activations `map[string]bool`
1. `ActiveReputersSetEvent`
    topicId, nonceBlockHeight, addresses `[]string`
1. `ActiveInferersSetEvent`
    topicId, nonceBlockHeight, addresses `[]string`
1. `ActiveForecastersSetEvent`
    topicId, nonceBlockHeight, addresses `[]string`

### Workers
1. `InsertInfererPayloadEvent`
    bundle `WorkerDataBundle`
1. `InsertForecasterPayloadEvent`
    bundle `WorkerDataBundle`
1. `InsertReputerPayloadEvent`
    bundle `ReputerValueBundle`
1. `ReputerRegisteredEvent`
    topicId, reputer, owner
1. `WorkerRegisteredEvent`
    topicId, worker, owner
1. `ReputerUnregisteredEvent`
    topicId, reputer
1. `WorkerUnregisteredEvent`
    topicId, worker
1. `WorkerLastCommitSetEvent`
    topicId, height, nonce
1. `ReputerLastCommitSetEvent`
    topicId, height, nonce

### Rewards settled
1. `InfererRewardsSettledEvent`
    blockHeight, blockHeightTx, rewards `[]TaskReward`
1. `ForecasterRewardsSettledEvent`
    blockHeight, blockHeightTx, rewards `[]TaskReward`
1. `ReputerAndDelegatorRewardsSettledEvent`
    blockHeight, blockHeightTx, rewards `[]TaskReward`

### Stake & delegation
1. `AddStakeEvent`
    topicId, reputer, delegator, amount
1. `RewardDelegateStakeEvent`
    topicId, reputer, delegator, amount
1. `ReputerStakeRemovalCompletedEvent`
    topicId, blockHeight, reputer, amount
1. `DelegateStakeRemovalCompletedEvent`
    topicId, blockHeight, delegator, reputer, amount
1. `DelegateRewardShareUpdatedEvent`
    topicId, reputer, rewardPerShare
1. `DelegateRewardDistributedEvent`
    topicId, reputer, amount

### Topics
1. `CreateNewTopicEvent`
    topic `Topic`
1. `FundTopicEvent`
    topicId, funder, amount
1. `TopicWeightUpdatedEvent`
    topicId, newWeight
1. `ActiveTopicsAtBlockSetEvent`
    targetBlockHeight, topicIds `[]TopicId`
1. `TopicInitialRegretSetEvent`
    topicId, blockHeight, regret
1. `TopicInitialEmaScoreSetEvent`
    actorType, topicId, blockHeight, score
1. `TopicRewardSetEvent`
    topicRewards `map[uint64]*Dec`

### Other
One of the `ParamsSetEvent`s is for `mint`, the other for `emissions`.
1. `ParamsSetEvent`
    params Params, blocksPerMonth
1. `ParamsSetEvent`
    params Params
1. `PruneRecordsEvent`
    blockHeight, topicId

x/emissions/keeper/actor_utils/worker.go
guilherme-brandao: Moved this event to `UpdateReputerNonce`

x/emissions/keeper/keeper.go
guilherme-brandao: Created a new event to be emitted every topic activation and inactivation

xmariachi (approved):

PR #855, opened 2025-09-02
Add new actor window queries

guilherme-brandao:

## Still Left Todo


x/emissions/keeper/queryserver/query_server_topics.go
```diff
@@ -164,3 +165,181 @@ func (qs queryServer) GetNextChurningBlockByTopicId(ctx context.Context, req *ty
...
+				response.IsOpen = true
+				response.CurrentNonceBlockHeight = nonce.BlockHeight
+				response.WindowStartBlock = windowStart
+				response.WindowEndBlock = windowEnd
+			}
+		}
+	}
+
+	// Calculate next window timing based on topic epoch
+	nextChurningBlock, isActive, err := qs.k.GetNextPossibleChurningBlockByTopicId(ctx, req.TopicId)
```
xmariachi, line 237: Just a note: I first thought that we may also include an `is_active` flag, however then I noticed that this is covered by having a `Next...` values informed (which is the way we check topic activity in `isTopicActive` ), so it would be a bit redundant.  If there is no "next" , then it means also that the topic is inactive... but may not be crystal clear to the user.

wdyt? Should we add `is_active` explicitly for a full picture?

x/emissions/keeper/queryserver/query_server_topics.go
```diff
@@ -164,3 +165,181 @@ func (qs queryServer) GetNextChurningBlockByTopicId(ctx context.Context, req *ty
...
+
+		if currentBlockHeight >= windowStart && currentBlockHeight <= windowEnd {
+			// If this is the first active nonce or it's more recent than the current latest
+			if latestActiveNonce == nil || nonce.ReputerNonce.BlockHeight > latestActiveNonce.ReputerNonce.BlockHeight {
+				latestActiveNonce = nonce
+				response.IsOpen = true
+				response.CurrentNonceBlockHeight = nonce.ReputerNonce.BlockHeight
+				response.WindowStartBlock = windowStart
+				response.WindowEndBlock = windowEnd
+			}
```
xmariachi, line 322: once we found the nonce that applies, we can exit the loop once we find the nonce for efficiency, right?

x/emissions/keeper/queryserver/query_server_topics.go
```diff
@@ -164,3 +165,181 @@ func (qs queryServer) GetNextChurningBlockByTopicId(ctx context.Context, req *ty
...
+		}
+	}
+
+	// For next window timing, we need to check unfulfilled worker nonces
```
xmariachi: Nope - They open on next nonce + GroundTruthLag. 
~I think in practice the most economical way to get this, is just `windowStart + EpochLength`, and no reference to worker nonces. 
This assumes topic is active - and still, this is an estimation (topic may be inactivated between now and next window), so it's fine.~
We can get it from the loop before: first iteration get the nonce, second iteration get the next-nonce, then exit - that would be the real info right there).

(Note for current solution as is, that worker nonces do not remain there - so if gtlag is big it won't work , note that there would be a big decalage between current worker nonce, and the projected reputer nonce)

xmariachi (commented):
Partial review

x/emissions/keeper/queryserver/query_server_topics.go
guilherme-brandao: I think it's a good idea to add `isTopicActive` property, but it would be just for workers, right?

x/emissions/keeper/queryserver/query_server_topics.go
guilherme-brandao: in this case, we should go until the end to find the most recent one.

x/emissions/module/autocli.go
```diff
@@ -673,30 +673,6 @@ func (am AppModule) AutoCLIOptions() *autocliv1.ModuleOptions {
...
-				},
-				{
-					RpcMethod: "GetActiveReputersForTopic",
-					Use:       "active-reputers [topic_id]",
-					Short:     "Get active reputers for a topic",
-					PositionalArgs: []*autocliv1.PositionalArgDescriptor{
-						{ProtoField: "topic_id"},
-					},
-				},
 				// add inclusion methods
```
xmariachi, line 676: The new queries should also be added into the CLI ( `GetReputerSubmissionWindowStatus`, and worker )

xmariachi (changes requested):
Second leg of the review

x/emissions/module/autocli.go
guilherme-brandao: added!

x/emissions/keeper/queryserver/query_server_topics.go
xmariachi: Why?
I think it could be for both. 
Currently our way of determining whether a topic is active, is to check whether this topic has a next churning block, right?
But topic activity is "the same" for both workers and reputers.

x/emissions/keeper/queryserver/query_server_topics.go
xmariachi: I think they are sorted the other way: most recent first.
However a bit below, I make a comment about not just getting the first one - also the next so we can use it for the `next`.

```
 allorad q emissions  unfulfilled-reputer-nonces 63  --node https://allora-rpc.testnet.allora.network   
nonces:
  nonces:
  - reputer_nonce:
      block_height: "5469384"
  - reputer_nonce:
      block_height: "5469324"
  - reputer_nonce:
      block_height: "5469264"
  - reputer_nonce:
      block_height: "5469204"
...
```

x/emissions/keeper/queryserver/query_server_topics.go
```diff
@@ -164,3 +165,183 @@ func (qs queryServer) GetNextChurningBlockByTopicId(ctx context.Context, req *ty
...
+			return nil, errors.Wrapf(err, "error checking reputer whitelist status")
+		}
+		response.IsWhitelisted = isWhitelisted
+	}
+
+	var latestActiveNonce *types.ReputerRequestNonce
+	var earliestFutureNonce *types.ReputerRequestNonce
+
+	for _, nonce := range nonces.Nonces {
+		extraLag := topic.GroundTruthLag % topic.EpochLength
```
xmariachi: these four lines of the extraLag calculation can go outside the loop - they are not dependent on the nonce but on the topic.

x/emissions/keeper/queryserver/query_server_topics.go
xmariachi: using for both. Adding linter fixes [here](https://github.com/allora-network/allora-chain/pull/856)

x/emissions/keeper/queryserver/query_server_topics.go
xmariachi: refactored [here](https://github.com/allora-network/allora-chain/pull/856)

xmariachi (approved):

x/emissions/proto/emissions/v9/query.proto
```diff
@@ -555,6 +540,18 @@ service QueryService {
...
     option (google.api.http).get = "/emissions/v9/latest_forecaster_weight/{topic_id}/{actor_id}";
   }
+
+  // Get worker submission window status and timing
+  rpc GetWorkerSubmissionWindowStatus(GetWorkerSubmissionWindowStatusRequest) returns (GetWorkerSubmissionWindowStatusResponse) {
+    option (cosmos.query.v1.module_query_safe) = true;
+    option (google.api.http).get = "/emissions/v9/worker_submission_window_status/{topic_id}";
+  }
+
+  // Get reputer submission window status and timing
```
amimart, line 550: We can have multiple nonce open at the same time so I think we should query the window status for a nonce, not only a topic.

Also maybe we could have a single query to provide both worker and reputer window information

PR #859, opened 2025-09-12
Fix OR-NIs event emission + ValueBundle optimized types + custom proto

xmariachi:
* Emit Outlier Resistant network inferences event correctly (using unwrapped context)
* Fix the format, aligning with Network Inferences.

- [x] If tested, please describe how. If not, why tests are not needed. -- regular unit testing
- [x] If documented, please describe where. If not, describe why docs are not needed. -- no need 


x/emissions/proto/emissions/v9/events.proto
```diff
@@ -62,6 +61,11 @@ message EventNetworkInferences {
   EventValueBundle value_bundle = 2;
 }
 
+message EventOutlierResistantNetworkInferences {
+  reserved 1, 2, 3;
```
zale144: why reserve these?

xmariachi: A good practice on evolving protos is to reserve the fields that are no longer used.
Prevents accidental reuse and also makes the intent explicit.

xmariachi: In fact - these ones are evolutions as well, right? They should also reserve the unused field 1 - I can add it to this PR maybe?

```
message EventNetworkLossSet {
  EventValueBundle value_bundle = 2;
}

message EventNetworkInferences {
  EventValueBundle value_bundle = 2;
}
```

zale144: these should be removed, and in the event emission, we could omit the `v9` bit, since it probably won't make sense to have it anymore

x/emissions/types/events_emitters.go
```diff
@@ -70,6 +68,21 @@ func EmitNewNetworkInferencesEvent(ctx context.Context, networkInferences ValueB
...
+func EmitNewOutlierResistantNetworkInferencesEvent(ctx context.Context, networkInferences ValueBundle) {
+	metrics.IncrProducerEventCount(metrics.OUTLIER_RESISTANT_NETWORK_INFERENCES_EVENT)
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	jb, err := valueBundleToEventValueBundleJSON(&networkInferences)
+	if err != nil {
+		sdkCtx.Logger().Warn("Error emitting NewOutlierResistantNetworkInferencesEvent", "error", err)
+		return
+	}
+	sdkCtx.EventManager().EmitEvent(
+		sdk.NewEvent("emissions.v9.EventOutlierResistantNetworkInferences",
```
zale144: here I mean, I think it should be just `"emissions.EventOutlierResistantNetworkInferences"`

guilherme-brandao (commented):
We need to also check `NewInsertReputerPayloadEventBase`.

x/emissions/types/events_emitters.go
```diff
@@ -70,6 +68,21 @@ func EmitNewNetworkInferencesEvent(ctx context.Context, networkInferences ValueB
...
 
+func EmitNewOutlierResistantNetworkInferencesEvent(ctx context.Context, networkInferences ValueBundle) {
+	metrics.IncrProducerEventCount(metrics.OUTLIER_RESISTANT_NETWORK_INFERENCES_EVENT)
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	jb, err := valueBundleToEventValueBundleJSON(&networkInferences)
+	if err != nil {
+		sdkCtx.Logger().Warn("Error emitting NewOutlierResistantNetworkInferencesEvent", "error", err)
+		return
+	}
+	sdkCtx.EventManager().EmitEvent(
```
amimart: Do we know why the `EventManager#EmitTypedEvent` func doesn't work?

x/emissions/proto/emissions/v9/events.proto
amimart: Maybe it'd make sense to put the topic id at the highest level in the event message instead of the `EventValueBundle`, otherwise we won't be able to filter events on it

x/emissions/types/events_emitters.go
xmariachi: I think we should still keep the versioning. Future releases may introduce changes. And it makes sense to use the same version as the rest of the events protos.

amimart: I tend to disagree here, the API contract is defined by the proto spec, if not it is not part of that API. It seems very dangerous to me as it can easily be forgotten during a release of a new API version. And it's also misleading, a client could think an associated proto generated type exists and thus use the `cosmostypes.ParseTypedEvent()` func to unmarshal it

x/emissions/types/events_utils.go
```diff
@@ -149,6 +150,18 @@ func valueBundleToEventValueBundleBase(bundle *ValueBundle) *EventValueBundle {
 	}
 }
 
+// valueBundleToEventValueBundleJSON returns a JSON representation of the ValueBundle.
+// NOTE: the ValueBundle contains an array of array containing messages, which the protobuf serializes as such,
+// but in order to emit the event with the actual two-dimensional array which we want, we must use json.Marshal.
+func valueBundleToEventValueBundleJSON(bundle *ValueBundle) ([]byte, error) {
+	evBundle := valueBundleToEventValueBundleBase(bundle)
+	jsn, err := json.Marshal(evBundle)
```
amimart: We should avoid using this marshaller as it follows the json rfc 8259 which is not fully interoperable regarding numbers: we may have issues with integers using more than 53bits, which should be encoded as strings.

For the sake of consistency, I think we should use the proto json codec

amimart (changes requested):
I'm not comfortable with the idea of bypassing the cosmos proto json marshalling, here's what I suggests to still use the default marshalling but having the control only on that double dimensional array:

Let's introduce a custom type for that, implementing the needed proto marshalling funcs:

```go
type DecArray []Dec

var _ sdk.CustomProtobufType = DecArray{}

func (d DecArray) MarshalJSON() ([]byte, error) {
	b := bytes.Buffer{}
	b.WriteByte('[')
	for i, dec := range d {
		if i > 0 {
			b.WriteByte(',')
		}
		decJson, err := dec.MarshalJSON()
		if err != nil {
			return nil, err
		}
		if _, err := b.Write(decJson); err != nil {
			return nil, err
		}
	}
	b.WriteByte(']')
	return b.Bytes(), nil
}

// ...
```

Then in the `EventValueBundle` we can use that custom type:

```proto
repeated string one_out_inferer_forecaster_values = 14 [
    (gogoproto.customtype) = "github.com/allora-network/allora-chain/math.DecArray",
    (gogoproto.nullable) = false
  ];
```

And in the event emission we can keep using the `EventManager#EmitTypedEvent` func

x/emissions/types/events_emitters.go
xmariachi: But in that case, how would we evolve those types? 
However, I agree on the other concerns - I don't like this "rogue" proto generation if we can avoid it - and it seems like with [your suggestion here](https://github.com/allora-network/allora-chain/pull/859#pullrequestreview-3233635082), it would still be consistent and we can unify and emit typed events in every case, right?

amimart: In that case I would detach the compatibility from the emissions proto API versioning and rely on block height upgrade to identify which event version to use, which is far from ideal...

And yes with that suggestion we can keep all events under the proto API

xmariachi: resolving since with new protos the problem disappears - it will emit typed events only.

xmariachi:
> We need to also check `NewInsertReputerPayloadEventBase`.

Hey, I think this remark made sense before, but now that we are still emitting typed events, I think it is covered already (it gets the same treatment as the other ones using [valueBundleToEventValueBundleBase](https://github.com/allora-network/allora-chain/blob/3b9ca23a98ac8d644d29ee0f8acd4303312b07a3/x/emissions/types/events_utils.go#L106) e.g. `NewNetworkLossSetEventBase`, `NewNetworkInferencesEventBase`, etc.

cc @amimart

math/dec_array_test.go
```diff
@@ -0,0 +1,173 @@
...
+		name      string
+		data      []byte
+		expectErr bool
+		expected  math.DecArray
+	}{
+		{
+			name:      "empty array",
+			data:      []byte(`[]`),
+			expectErr: false,
+			expected:  nil,
```
xmariachi, line 119: is empty array expected to be `nil` instead of empty array?

math/dec_array_test.go
```diff
@@ -0,0 +1,173 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			require.Equal(t, tc.expected, tc.arr1.Equal(tc.arr2))
+		})
+	}
+}
+
```
xmariachi, line 55: `Marshal()` is tested but `MarshalTo()` is not. Even if it is a simple piece of code which does use `Marshal()` , no harm on completing the unit tests for it too, right?

math/dec_array_test.go
xmariachi: Produced [another PR](https://github.com/allora-network/allora-chain/pull/862) to this branch with additional suggested changes. Let's discuss there

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -62,6 +61,11 @@ message EventNetworkInferences {
   EventValueBundle value_bundle = 2;
 }
```
xmariachi, line 71: In v0.12.1, we had these:
```
message EventNetworkLossSet {
  uint64 topic_id = 1;
  int64 block_height = 2;
  emissions.v3.ValueBundle value_bundle = 3;
}

message EventNetworkInferences {
  uint64 topic_id = 1;
  int64 block_height = 2;
  emissions.v3.ValueBundle value_bundle = 3;
}
```
now we are changing it to this:
```
message EventNetworkLossSet {
  EventValueBundle value_bundle = 2;
}

message EventNetworkInferences {
  EventValueBundle value_bundle = 2;
}
```
I don't think this is right: reuses "2", and uses the same name, losing any backwards-compatibility. Also adding your suggestion about keeping topic_id, and to keep compatibility with earlier v0.12.0 and v0.12.1 versions, I would suggest:
```
message EventNetworkLossSet {
  uint64 topic_id = 1;
  int64 block_height = 2; // even if not used now, but it will filled by the prior versions, and can be reserved in v0.13.x
  emissions.v3.ValueBundle value_bundle = 3;  // same here
  EventValueBundle event_value_bundle = 4; // change name, otherwise it may conflict.
}
// ... same for network inferences. 
```
wdyt?

math/dec_array_test.go
xmariachi: added on [PR](https://github.com/allora-network/allora-chain/pull/862) to this one

math/dec_array_test.go
xmariachi: In particular this test has this behaviour, is that expected? 
```
{
			name:      "extra content after array",
			data:      []byte(`["1.0"]extra`),
			expectErr: false,
			expected:  math.DecArray{math.MustNewDecFromString("1.0")},
		},
```

xmariachi (commented):
Some comments, including a PR on unit tests

math/dec_array_test.go
amimart: > is empty array expected to be nil instead of empty array?

Yeah in Go an empty slice should be preferred `nil`, on the encoding side a `nil` slice will always be encoded as `[]`, and `nil` vs empty is the same on the usage, for example:
- `append([]byte(nil), 'x')` == `append([]byte{}, 'x')`
- `len([]byte(nil)) == len([]byte{})`

The test you added should fail, I'll fix that

x/emissions/proto/emissions/v9/events.proto
amimart: Agree, I just pushed the related changes marking the old fields as deprecated

guilherme-brandao (approved):

amimart (approved):

PR #863, opened 2025-09-19
Emit event when reputer stake is updated

amimart:

Add the emission of a `EventReputerStakeUpdated` event whenever a reputer stake changes, it's a derived event coming to complement the primitive one: `EventAddStake` & `EventRemoveStake`.

In addition, the `EventRequestStakeRemoval` & `EventCancelStakeRemoval` events are also emitted, denoting the intention to start or cancel the undelegation process.

- https://linear.app/alloralabs/issue/ENGN-4212/event-emit-eventupdatedreputerstake


x/emissions/types/events_emitters.go
```diff
@@ -89,13 +89,30 @@ func EmitNewCreateNewTopicEvent(ctx context.Context, topic *Topic) {
 	}
 }
 
-func EmitNewAddStakeEvent(ctx context.Context, topicId TopicId, reputer, delegator string, amount cosmosMath.Int) {
+func EmitNewAddStakeEvent(ctx context.Context, topicId TopicId, reputer, delegator string, amount, totalStake cosmosMath.Int) {
 	metrics.IncrProducerEventCount(metrics.ADD_STAKE_EVENT)
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
-	err := sdkCtx.EventManager().EmitTypedEvent(NewAddStakeEventBase(topicId, reputer, delegator, amount))
-	if err != nil {
+	if err := sdkCtx.EventManager().EmitTypedEvent(NewAddStakeEventBase(topicId, reputer, delegator, amount)); err != nil {
```
guilherme-brandao, line 95: I think we should create a metric in `IncrProducerEventCount` for delegate stake add event, like we do in the removal.

x/emissions/keeper/keeper.go
```diff
@@ -2359,6 +2359,15 @@ func (k *Keeper) RemoveReputerStake(
 		return errorsmod.Wrapf(err, "Deleting stake removal from queue failed")
 	}
 
+	types.EmitNewRemoveStakeEvent(
+		sdk.UnwrapSDKContext(ctx),
+		topicId,
+		reputer,
+		"",
+		stakeToRemove,
+		topicStakeNew,
```
guilherme-brandao: It should be `reputerStakeNew` not `topicStakeNew`

x/emissions/keeper/keeper.go
```diff
@@ -2515,6 +2524,15 @@ func (k *Keeper) RemoveDelegateStake(
...
 
+	types.EmitNewRemoveStakeEvent(
+		sdk.UnwrapSDKContext(ctx),
+		topicId,
+		reputer,
+		delegator,
+		stakeToRemove,
+		topicStakeNew,
+	)
+
```
guilherme-brandao, line 2535: same

guilherme-brandao (changes requested):

guilherme-brandao (commented):
The `RemoveStake` event used to be emitted when stake removal started. I’m wondering whether this will cause problems interpreting old events that have the same type but were emitted at the beginning of the staking process and could have been canceled partway through.

x/emissions/proto/emissions/v9/events.proto
```diff
@@ -159,6 +159,33 @@ message EventAddStake {
...
+  string delegator = 3;
+  string amount = 4 [
+    (cosmos_proto.scalar) = "cosmos.Int",
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.nullable) = false,
+    (amino.dont_omitempty) = true
+  ];
+}
+
+// EventReputerStakeUpdated is an event derived from the primitives ones EventAddStake and EventRemoveStake.
```
xmariachi: Good to add this comment here. Suggestion on wording/grammar: 
"the primitives ones" => "the primitive ones" , or "the primitive events". 
Also below I think "in an event sourcing approach" is more correct.

x/emissions/types/events_emitters.go
xmariachi: yep, it would be consistent with the pattern.

x/emissions/module/stake_removals.go
```diff
@@ -103,15 +103,6 @@ func RemoveStakes(
...
-
-		// Emit event indicating a reputer stake removal has been processed
-		emissionstypes.EmitNewStakeRemovalCompletedEvent(
-			sdkCtx,
-			stakeRemoval.TopicId,
-			stakeRemoval.Reputer,
-			"",
-			stakeRemoval.Amount,
-		)
 	}
```
xmariachi, line 106: Apart from the event name change (which make sense), this event emission has been moved into `RemoveReputerStake`.
However I wonder, if it would be better to still emit it after the `write()` as it was, which would emit the event _only if_ everything went fine, otherwise an event would have been emitted but ultimately there was no state modification. 
Is there a reason for moving this one up inside the function?

amimart: The reason is that events should only be emitted by the Keeper, if an event is linked to a particular state change, this is this only way to ensure it by design.
If we are operating on a cached context, the events will only be propagated to the parent's context on the call to the write func, so I think we're good

x/emissions/types/events_emitters.go
amimart: Shouldn't provide only one event metric on the removal instead? we should have one event => one metric

I don't think we should differentiate delegations from the reputer itself with other delegators, I'd even add the reputer addr in the `delegator` field instead of empty. This would open the way to a future mutualization of stake management txs

guilherme-brandao (dismissed):

x/emissions/module/stake_removals.go
xmariachi: got it, ty!

x/emissions/types/events_emitters.go
xmariachi: I see. Unifying delegation and self-delegation makes sense to me as a goal, even if not reached yet. Are there other parts that should be changed for unification?

(re the delegator being empty - it signals that it is a self-delegation. I don't have a strong opinion there, because we'd need to analyse the use cases to see what's best; but it should not be a blocker, right? It allows to quickly determine self-delegation, in a similar way that if we add the delegator to be the reputer there, we would determine self-delegation when delegator = reputer. Meaning: same information is there, only defined differently).

amimart: There's also the `EventRewardDelegateStake` event but it can't be used for self-delegation as their rewards goes directly to their stake.

The `delegator` field being empty or not is just semantic and could works either way, my opinion is that in the case of self-delegation there is actually a delegator so I'd tend to fill it, an empty field could let think the related stake isn't related to a delegator.

xmariachi: OK, lgtm to fill it but pls check with downstream effects of changing this (indexers, consumers), in case a "delegator is empty` check is done at some place.

guilherme-brandao (approved):

x/emissions/types/events_emitters.go
xmariachi: confirmed ok

xmariachi (approved):

PR #868, opened 2025-10-02
scores take 2 gtlags to appear on topic start

xmariachi:
* Generation of oneOutInfererValues in the network inference generation, in the case where there are no network losses (in which case a median of the combined value is applied). 
This in turn makes the scoring process to be activated on that epoch, otherwise there is no scoring on those epochs until there are actually network losses.
This will reduce the warm-up time until rewards can be given, from 2 GTLags, to 1 GTLag + 1 EpochLength.   
* Added `Average` function over a vector of `Dec`

- [x] If tested, please describe how. If not, why tests are not needed. Added unit tests over the `Average`function and 
- [x] If documented, please describe where. If not, describe why docs are not needed. -- a bugfix, affects starting time of a topic, but it does not affect documentation.

### DIFF
--- a/x/emissions/keeper/inference_synthesis/network_inferences.go
+++ b/x/emissions/keeper/inference_synthesis/network_inferences.go
@@ -78,7 +78,7 @@ func calcNetworkInferencesMultipleByMedian(
 		return nil, errorsmod.Wrap(err, "while calculating median")
 	}
 
-	// Compute one-out median for each inferer
+	// Compute one-out average for each inferer
 	oneOutInfererValues := make([]*emissions.WithheldWorkerAttributedValue, 0, len(inferences.Inferences))
 	for i, inf := range inferences.Inferences {
 		// Build slice without the current inference
@@ -89,15 +89,15 @@ func calcNetworkInferencesMultipleByMedian(
 			}
 		}
 
-		// Calculate one-out median
-		oneOutMedian, err := alloraMath.Median(without)
+		// Calculate one-out average
+		oneOutAverage, err := alloraMath.Average(without)
 		if err != nil {
 			return nil, errorsmod.Wrapf(err, "while calculating one-out median for inferer %s", inf.Inferer)
 		}
 
 		oneOutInfererValues = append(oneOutInfererValues, &emissions.WithheldWorkerAttributedValue{
 			Worker: inf.Inferer,
-			Value:  oneOutMedian,
+			Value:  oneOutAverage,
 		})
 	}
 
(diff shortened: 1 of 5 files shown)


fstecker (changes requested):
1. What's the reason for using averages? A median of the (n-1) inferers seems more appropriate, as that's also how the not-one-out inference is computed.
2. Here's a suggestion for a completely different solution lol: 
    1. move retrieving the network losses via `k.GetLatestNetworkLossBundle` from `GetNetworkInferences` into `GetCalcNetworkInferenceArgs`. That's the only place where it's used, and the only field that's used is `networkLosses.CombinedValue`
    2. if the network loss can't be found, just use 0 instead of `networkLosses.CombinedValue`
    3. delete the `calcNetworkInferencesMultipleByMedian` function

The advantage of the alternative solution would be that we get rid of this terrible median codepath entirely. Disadvantage is I'm only like 80% convinced that 0 as a placeholder value doesn't break anything. I'll think about it some more until tomorrow.

xmariachi:
> What's the reason for using averages? 
Initially I used the median but checked the whitepaper, and I also thought that this would be more meaningful - we don't have weights for a weighted-average, but lmk If you think median is better.

> a completely different solution
Nice, I would also like to get rid of this codepath. Please let me know about that 20% :)

x/emissions/keeper/inference_synthesis/network_inferences.go
```diff
@@ -78,6 +78,29 @@ func calcNetworkInferencesMultipleByMedian(
...
+	// Compute one-out average for each inferer
+	oneOutInfererValues := make([]*emissions.WithheldWorkerAttributedValue, 0, len(inferences.Inferences))
+	for i, inf := range inferences.Inferences {
+		// Build slice without the current inference
+		without := make([]alloraMath.Dec, 0, len(inferenceValues)-1)
+		for j, val := range inferenceValues {
+			if i != j {
+				without = append(without, val)
+			}
+		}
```
amimart, line 90:
```suggestion
        without := append([]alloraMath.Dec(nil), inferenceValues[:i]...)
		without = append(without, inferenceValues[i+1:]...)
```

This can be simplified

amimart (approved):
Looks good!
Just let a slight suggestion :)

x/emissions/keeper/inference_synthesis/network_inferences.go
xmariachi: Thanks! This PR is abandoned though - please review this other one instead: https://github.com/allora-network/allora-chain/pull/871

xmariachi:
Abandoned - #871 instead

PR #870, opened 2025-10-07
feat: expose events on x/scheduler module

amimart:

Add the emissions of the events below in the `x/scheduler` module:
- `TaskScheduledEvent`: Emitted when a task is scheduled or rescheduled at a specific time (i.e. a task can be created without being scheduled, in that case the event is not emitted);
- `TaskUnscheduledEvent`: Emitted when a task that was scheduled, is cancelled or unscheduled;
- `TaskExecutedEvent`: Emitted when a task is executed.;


xmariachi (dismissed):

xmariachi (dismissed):
lgtm!

x/scheduler/keeper/keeper.go
```diff
@@ -163,14 +163,30 @@ func (k *Keeper) ScheduleTask(ctx context.Context, typename string, id types.Tas
...
+	if key := getTaskScheduleKey(task); key != nil {
+		if err := sdkCtx.EventManager().EmitTypedEvent(&types.TaskScheduledEvent{
+			Id:       task.Id,
+			Typename: task.Typename,
+			At:       task.ScheduledFor,
+		}); err != nil {
+			return err
+		}
+
 		return k.tasksSchedule.Set(ctx, *key)
```
guilherme-brandao: I think the event should be triggered after and it doesn't block setting the task if there's an error.

x/scheduler/keeper/keeper.go
```diff
@@ -181,6 +197,14 @@ func (k *Keeper) CancelTask(ctx context.Context, taskID types.TaskID) error {
...
 	if key := getTaskScheduleKey(task); key != nil {
+		if err := sdkCtx.EventManager().EmitTypedEvent(&types.TaskUnscheduledEvent{
+			Id:       task.Id,
+			Typename: task.Typename,
+			At:       task.ScheduledFor,
+		}); err != nil {
+			return err
+		}
+
 		return k.tasksSchedule.Remove(ctx, *key)
```
guilherme-brandao: same here

x/scheduler/keeper/abci.go
```diff
@@ -78,6 +78,14 @@ func (k *Keeper) executeTask(ctx context.Context, task types.Task, handler types
 		return errors.Wrapf(types.ErrTaskExecution, "run func failed for task '%s' of type '%s': %s", task.Id, handler.Typename(), err)
 	}
 
+	if err := sdkCtx.EventManager().EmitTypedEvent(&types.TaskExecutedEvent{
+		Id:       task.Id,
+		Typename: task.Typename,
+		At:       task.ScheduledFor,
+	}); err != nil {
+		return err
+	}
```
guilherme-brandao: It would be good to follow the pattern from the `event_emitters.go` file in the emissions module, where we created all the events there, and we just log the errors.

x/scheduler/keeper/abci.go
```diff
@@ -78,6 +78,14 @@ func (k *Keeper) executeTask(ctx context.Context, task types.Task, handler types
...
+
 	oldScheduleKey := getTaskScheduleKey(task)
```
guilherme-brandao, line 89: Usually we return the struct + err on the keeper's get methods

x/scheduler/keeper/keeper.go
amimart: Yeah I'll move it you're right it's always a good practice to emit the event after, but in reality we are operating on a cached context so if either fails the whole context is dropped

guilherme-brandao: yeah but probably we need to just log this error to avoid this scenario, no?

x/scheduler/keeper/abci.go
amimart: I tend to disagree, events must reflect the state so both events and state updates should succeed, btw the only way for the event emission to fail at this point is a marshalling error so it's very unlikely to happen

x/scheduler/keeper/abci.go
amimart: Oh but the task schedule key can be `nil` it's not an error, but it's an impossible/forbidden case when we execute a task

x/scheduler/keeper/keeper.go
amimart: If we allow the event emission to fail by silencing the error we accept inconsistent block events, I prefer to ensure that the events are aligned with state changes

guilherme-brandao: yeah, makes sense! I was wondering if it would be good or not to block state transactions / txs from happening because of failed events, since the cached context is dropped if the error happen.

x/scheduler/keeper/abci.go
zale144: but then we should change it everywhere else. Also, let's not forget the deprecation of `block_height` in favor of `nonce`. Both of these should be tackled in a separate PR me thinks

x/scheduler/keeper/abci_test.go
```diff
@@ -424,6 +469,21 @@ func TestRunTask(t *testing.T) {
...
 }
+
+func parseEvents(t *testing.T, events []abci.Event) []proto.Message {
+	t.Helper()
+
+	var parsed []proto.Message
+	for _, e := range events {
+		evt, err := cosmostypes.ParseTypedEvent(e)
+		require.NoError(t, err)
+		parsed = append(parsed, evt)
```
zale144, line 486: maybe consider extracting this for parsing single events, in case we might use it. maybe in a util? just a thought

zale144 (dismissed):

xmariachi (dismissed):

x/scheduler/keeper/keeper.go
```diff
@@ -181,7 +197,15 @@ func (k *Keeper) CancelTask(ctx context.Context, taskID types.TaskID) error {
 	}
 
 	if key := getTaskScheduleKey(task); key != nil {
-		return k.tasksSchedule.Remove(ctx, *key)
+		if err := k.tasksSchedule.Remove(ctx, *key); err != nil {
+			return err
```
zale144, line 201: would it make sense to wrap such errors with a message?

x/scheduler/proto/scheduler/v1/events.proto
```diff
@@ -0,0 +1,56 @@
...
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/duration.proto";
+
+option go_package = "github.com/allora-network/allora-chain/x/scheduler/types";
+
+// TaskScheduledEvent is emitted when a task is scheduled or rescheduled at a specific time.
+// If a task is created with no scheduling (i.e. scheduled_for is empty), this event is not emitted until the task is scheduled.
+// If an arbitrage decision is made to reschedule a task, this event is emitted.
+// For a periodic task, this event is emitted at each execution.
+message TaskScheduledEvent {
```
zale144: I think `EventTaskScheduled` would be more consistent with general event type naming. Same for the others

amimart: Yep that's true I'll change that thanks 🙏

x/scheduler/keeper/keeper.go
amimart: generic collections errors looks clear enough to me so I'd rather let it untouched, and if an error occurs here this is abnormal and independent from inputs

x/scheduler/keeper/abci_test.go
amimart: It is currently only used here, but the day we need it elsewhere yes it'd be worth putting it as a global test utility func

xmariachi (approved):

PR #871, opened 2025-10-08
Scores take 2 gtlags remove median

xmariachi:
* Remove "median" codepath when there were no previous losses - the same code is run, only without forecasting layer (which requires previous losses)
* Fixes the "cold start" problem where a topic needs to wait 2 GroundTruthLags to get scored, because that "median" codepath did not generate `OneOutInfererValues`.  
 

- [x] If tested, please describe how. If not, why tests are not needed. -- with unit tests, modified to comply and adjust to the changes.
- [x] If documented, please describe where. If not, describe why docs are not needed. -- no need, just a bugfix and improvement.


fstecker (dismissed):
Looks great, though I really hoped this would have net negative lines of code! I guess it still does if you ignore tests :)

x/emissions/keeper/inference_synthesis/forecast_implied.go
```diff
@@ -38,6 +38,12 @@ func CalcForecastImpliedInferences(args CalcForecastImpliedInferencesArgs) (
 	err error,
 ) {
 	args.Logger.Debug("Calculating forecast-implied inferences", "topicId", args.TopicId)
+
+	// If NetworkCombinedLoss is nil, return empty maps immediately
+	if args.NetworkCombinedLoss == nil {
+		args.Logger.Debug("NetworkCombinedLoss is nil, returning empty forecast-implied inferences", "topicId", args.TopicId)
+		return make(map[Forecaster]*emissionstypes.Inference), args.InfererToRegret, args.ForecasterToRegret, nil
```
amimart:
```suggestion
		return nil, args.InfererToRegret, args.ForecasterToRegret, nil
```

May be an unnecessary allocation, a nil map has same behaviour than an empty one on read operations

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -967,28 +975,33 @@ func CalcNetworkInferences(
...
-			EpsilonTopic:                         args.EpsilonTopic,
-			EpsilonSafeDiv:                       args.EpsilonSafeDiv,
-			PNorm:                                args.PNorm,
-			CNorm:                                args.CNorm,
-			StdDevPlusEpsilon:                    args.StdDevPlusEpsilon,
-		})
-	if err != nil {
-		return &emissions.ValueBundle{}, RegretInformedWeights{}, errorsmod.Wrap(err, "CalcNetworkInferences() error calculating one-out forecaster inferences")
+	var oneOutForecasterInferences []*emissions.WithheldWorkerAttributedValue
+	if args.NetworkCombinedLoss != nil {
```
amimart, line 981: Instead of adding this check on each step onward we should early return:

```go
if args.NetworkCombinedLoss == nil {
		return &emissions.ValueBundle{
			TopicId: args.TopicId,
			ReputerRequestNonce: &emissions.ReputerRequestNonce{
				ReputerNonce: &emissions.Nonce{BlockHeight: args.InferenceBlockHeight},
			},
			Reputer:                       "allo1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqas6usy",
			ExtraData:                     nil,
			CombinedValue:                 combinedInference,
			InfererValues:                 inferences,
			ForecasterValues:              forecastImpliedInferences,
			NaiveValue:                    naiveInference,
			OneOutInfererValues:           oneOutInfererInferences,
			OneOutForecasterValues:        nil,
			OneInForecasterValues:         nil,
			OneOutInfererForecasterValues: nil,
		}, weights, err
	}
```

xmariachi: applied - but instead of sending the bundle twice, just the other code will happen only if. I think it keeps intention clear too (it can be clearly seen that if no previous losses, no forecasting layer).

amimart (approved):
🚀

guilherme-brandao (approved):

fstecker (approved):

PR #872, opened 2025-10-09
refactor(keeper): optimize validation and remove redundant checks

lukema95:
This PR refactors the `x/emissions/keeper/keeper.go` file to remove redundant code and improve performance. The changes eliminate unnecessary validation checks, remove impossible conditions, and optimize repeated function calls.

1. Remove Duplicate Validation in `SetDelegateRewardPerShare`
The `share` parameter was being validated 3 times in the same function:
```go
if err := types.ValidateDec(share); err != nil {
    return errorsmod.Wrap(err, "share is not valid")
}
// ... 
if err := types.ValidateDec(share); err != nil {  // Duplicate 
    return errorsmod.Wrapf(err, "SetDelegateRewardPerShare: invalid share")
}
```
Removed the duplicate validation calls, keeping only the first one.

2. Remove Impossible Negative Length Check
Code was checking if `len()` returns a negative value, which is impossible in Go:
```go
keysLen := len(keys)
if keysLen == 0 {
    return types.StakeRemovalInfo{...}, false, nil
}
if keysLen < 0 {  // This condition can NEVER be true
    return types.StakeRemovalInfo{}, false, 
        errorsmod.Wrapf(types.ErrInvariantFailure, 
            "Why is golang len function returning negative values?")
}
```
The Go language specification and runtime guarantee that `len()` will never return a negative value. 

3. Optimize Repeated `len()` Calls and Remove Redundant Conditions
Code was calling `len()` multiple times on the same slice and had redundant conditions:
```go
lenNonces := uint64(len(nonces.Nonces))  // First call
if lenNonces > maxUnfulfilledRequests {
    diff := uint64(len(nonces.Nonces)) - maxUnfulfilledRequests  // Second call!
    if diff > 0 {  // Always true when we reach here!
        nonces.Nonces = nonces.Nonces[:maxUnfulfilledRequests]
    }
}
```

Use the already computed length value and remove redundant conditions:

```go
lenNonces := uint64(len(nonces.Nonces))
if lenNonces > maxUnfulfilledRequests {
    nonces.Nonces = nonces.Nonces[:maxUnfulfilledRequests]
}
```

## Still Left Todo

### DIFF
--- a/x/emissions/keeper/keeper.go
+++ b/x/emissions/keeper/keeper.go
@@ -726,10 +726,7 @@ func (k *Keeper) AddWorkerNonce(ctx context.Context, topicId TopicId, nonce *typ
 
 	lenNonces := uint64(len(nonces.Nonces))
 	if lenNonces > maxUnfulfilledRequests {
-		diff := uint64(len(nonces.Nonces)) - maxUnfulfilledRequests
-		if diff > 0 {
-			nonces.Nonces = nonces.Nonces[:maxUnfulfilledRequests]
-		}
+		nonces.Nonces = nonces.Nonces[:maxUnfulfilledRequests]
 	}
 
 	if err := nonces.Validate(); err != nil {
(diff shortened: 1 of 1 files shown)


xmariachi:
Findings look correct. No need for entry in the CHANGELOG.

xmariachi (approved):
Findings are correct.

amimart (approved):
LGTM, thx for helping sanitising this.. 🙏

PR #874, opened 2025-10-10
feat: exposes rpc queries on xscheduler

amimart:

The PR brings RPC queries allowing to inspect the `x/scheduler` module state:
- `Tasks`: Retrieve all the tasks in the state, scheduled or not;
- `Task`: Retrieve the details of a task given its id;
- `ScheduledTasks`: Retrieve the currently scheduled tasks for a specific type, not the one created with no execution scheduled;
- `Handlers`: Retrieve the currently registered task handlers, in their execution order;

⚠️ The `Handlers` query is marked as **unsafe** as it is not based on any state but on the currently registered task handlers.

There are some pagination and filtering options on the `Tasks` and `ScheduledTasks` queries.

For each RPC query a related CLI query command has been added.

## Still Left Todo

- [x] Tests
- [x] CLI


x/scheduler/proto/scheduler/v1/query.proto
```diff
@@ -0,0 +1,107 @@
...
+message QueryScheduledTasksRequest {
+  // Typename is the type of tasks to query.
+  string typename = 1;
+
+  // From is an optional filter to only return tasks scheduled from a given time (i.e. after or before depending on the order, inclusive)
+  // With this filter pagination by offset is not supported, pagination by key must be used instead.
+  google.protobuf.Timestamp from = 2 [(gogoproto.stdtime) = true];
+
+  // Pagination defines an optional pagination for the request.
+  cosmos.base.query.v1beta1.PageRequest pagination = 4;
```
xmariachi: any reason why 4 is used here instead of 3 ?

amimart: Yeah that's a mistake I'll fix it

PR #875, opened 2025-10-10
fix(keeper): prevent clearing all active topics when removing non-existent topic

lukema95:

Fix a bug in `inactivateTopicWithoutMinWeightReset` function that could cause **all active topics to be inadvertently cleared** when attempting to remove a topic that doesn't exist in the active topics list.

```go
// Remove the topic from the active topics at the block
// If the topic is not found in the active topics at the block, no op
newActiveTopicIds := []TopicId{}  // Empty slice initialization
for i, id := range topicIdsActiveAtBlock.TopicIds {
    if id == topicId {
        newActiveTopicIds = append(topicIdsActiveAtBlock.TopicIds[:i], topicIdsActiveAtBlock.TopicIds[i+1:]...)
        break
    }
}
err = k.SetBlockToActiveTopics(ctx, block, types.TopicIds{TopicIds: newActiveTopicIds})
```

If `topicId` is not found in the loop (due to state inconsistency), `newActiveTopicIds` remains empty, and `SetBlockToActiveTopics` will clear all active topics for that block.

## Still Left Todo

### DIFF
--- a/x/emissions/keeper/topic_activation.go
+++ b/x/emissions/keeper/topic_activation.go
@@ -133,7 +133,7 @@ func (k *Keeper) inactivateTopicWithoutMinWeightReset(ctx context.Context, topic
 	}
 	// Remove the topic from the active topics at the block
 	// If the topic is not found in the active topics at the block, no op
-	newActiveTopicIds := []TopicId{}
+	newActiveTopicIds := topicIdsActiveAtBlock.TopicIds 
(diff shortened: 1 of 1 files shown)


xmariachi:
Thanks! We'll take a closer look for a more integrated solution, but your finding looks correct.

xmariachi:
Thanks, this has been done in a different PR involving more changes than this. Thanks for your help, appreciated!

PR #876, opened 2025-10-13
fix: preserve full set of inferers in oneOutInfererForecasterValues event field

zale144:
Handles the case where the `EventNetworkInferences`, `EventNetworkInferencesOutlierResistant` and `EventNetworkLoss` has a jagged matrix for the `OneOutInfererForecasterValues` field. When indexing these events in the Producer, the jagged 2D array causes a panic when inserting to postgres DB.
The solution is to add padding to the matrix until all forecasters contain the inner `OneOutInfererValues` with the complete set of values that matches the length of the active inferers. Where values are missing, a `NaN` is added.

- [x] If documented, please describe where. If not, describe why docs are not needed. - event update

## Still Left Todo


xmariachi (changes requested):
Solution implementation lgtm, though I'd add a table unit tests to ensure consistency on corner cases (eg 0/1 actor, N actors, N=max actors).

x/emissions/types/events_utils.go
```diff
@@ -147,57 +147,65 @@ func NewInsertReputerPayloadEventBase(bundle *ReputerValueBundle) proto.Message
...
-		forecasterValues = append(forecasterValues, forcVal.Value)
+
+	evb.InfererValues = convertArray(bundle.InfererValues, func(i *WorkerAttributedValue) alloraMath.Dec { return i.GetValue() })
+	evb.InfererAddresses = convertArray(bundle.InfererValues, func(i *WorkerAttributedValue) string { return i.GetWorker() })
+	evb.ForecasterValues = convertArray(bundle.ForecasterValues, func(i *WorkerAttributedValue) alloraMath.Dec { return i.GetValue() })
+	evb.ForecasterAddresses = convertArray(bundle.ForecasterValues, func(i *WorkerAttributedValue) string { return i.GetWorker() })
+	evb.OneOutInfererValues = convertArray(bundle.OneOutInfererValues, func(i *WithheldWorkerAttributedValue) alloraMath.Dec { return i.GetValue() })
+	evb.OneInForecasterValues = convertArray(bundle.OneInForecasterValues, func(i *WorkerAttributedValue) alloraMath.Dec { return i.GetValue() })
+	evb.OneOutForecasterValues = convertArray(bundle.OneOutForecasterValues, func(i *WithheldWorkerAttributedValue) alloraMath.Dec { return i.GetValue() })
+
```
fstecker, line 169: Doesn't this assume that the addresses in `InfererValues` and `OneOutInfererValues` come in the same order? Is that always true? Same for `ForecasterValues`, `OneInForecasterValues` and `OneOutForecasterValues`.

xmariachi: Good point. 
In the case that comes from the chain, this is guaranteed, but not on reputer submissions. I think we could validate / normalize at submission time, so all bundles are guaranteed to be ordered the same way.
(ie using `GetSortedKeys`, etc)

zale144: do we all agree that it can be addressed in this PR: https://github.com/allora-network/allora-chain/pull/812
instead of the current one?

xmariachi: This PR already fixes an issue so I think it can be merged.
Noting the reputer submission for separate investigation.

xmariachi (approved):

amimart (approved):

PR #878, opened 2025-10-16
Enabling emissions (WIP)

xmariachi:
* Adds `emissions.startingBlock` store
* Defines the logic when enabling emissions
* Changes the way scheduling comes into effect
* auto-wiring contained in the same module 

## Still Left Todo


app/tasks.go
```diff
@@ -0,0 +1,32 @@
...
+	schedulertypes "github.com/allora-network/allora-chain/x/scheduler/types"
+)
+
+// InitializeRecurringTasks schedules recurring tasks after the application state is loaded.
+func (app *AlloraApp) InitializeRecurringTasks() {
+	ctx := app.BaseApp.NewUncachedContext(true, tmproto.Header{ //nolint:exhaustruct
+		Height: app.LastBlockHeight(),
+	})
+
+	mintTaskHandlers := app.MintKeeper.TaskHandlers()
```
amimart: The task handlers should be exposed as a module output, it will directly be passed by depinject to the scheduler module that will call the register func automatically

app/tasks.go
```diff
@@ -0,0 +1,32 @@
...
+	if err := app.SchedulerKeeper.RegisterTaskHandlers(mintTaskHandlers); err != nil {
+		if !errors.Is(err, schedulertypes.ErrInvalidTaskHandler) || !strings.Contains(err.Error(), "duplicate task handler") {
+			app.Logger().Error("failed to register mint task handlers", "err", err)
+			return
+		}
+	}
+
+	if err := app.MintKeeper.ScheduleEmissionRecalculationTask(ctx, app.SchedulerKeeper, 0); err != nil {
```
amimart: This may be done in the mint module when receiving the scheduler keeper as module input

xmariachi: moved there - also made it happen only on migration, or on emissions enablement.

PR #883, opened 2025-10-21
Tokenomics Adjustments

xmariachi:

- [x] If tested, please describe how. If not, why tests are not needed. -- locally for now


x/mint/keeper/emissions.go
```diff
@@ -19,45 +24,75 @@ func GetLockedVestingTokens(
...
-	thirtySix := math.NewInt(36)
-	// calculate whether the number of months unlocked should be allowed to increase
-	calcMonthsUnlocked := blockHeight.Quo(math.NewIntFromUint64(blocksPerMonth))
-	if calcMonthsUnlocked.GTE(thirtySix) {
-		calcMonthsUnlocked = thirtySix
+
+	// Locked sub-amounts
+	lockedFoundationTotalAmount := lockedFoundationPercentage.MulTruncate(maxSupply).TruncateInt()
+
+	// ---- 3. Determine months since genesis ----
```
amimart:
```suggestion
	// ---- 3. Determine months since TGE ----
```

x/mint/keeper/emissions.go
```diff
@@ -19,45 +24,75 @@ func GetLockedVestingTokens(
 	blockHeight math.Int,
```
amimart, line 24: AFAIU this parameter is actually used as a block count since TGE so it requires every path leading to this call to compute that count (i.e. currentHeight - tgeHeight).

I see a lot of paths calling this func so I'd rather set two arguments: `currentHeight` and `tgeHeight` and do the count computation in here instead

x/mint/keeper/keeper.go
```diff
@@ -315,25 +321,33 @@ func (k Keeper) GetEmissionInfo(ctx context.Context) (*types.Params, *types.Even
...
+	if err != nil {
+		return nil, nil, errorsmod.Wrap(err, "failed to get starting emissions block height")
+	}
 
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
 	blockHeight := uint64(sdkCtx.BlockHeight())
-	numberOfRecalcs := blockHeight / blocksPerMonth
-	blockHeightTarget_e_i_LastCalculated := numberOfRecalcs*blocksPerMonth + 1          //nolint:revive // var-naming: don't use underscores in Go names
-	blockHeightTarget_e_i_Next := blockHeightTarget_e_i_LastCalculated + blocksPerMonth //nolint:revive // var-naming: don't use underscores in Go names
+	blockHeightTGEAdjusted := blockHeight - (uint64(startingEmissionsBlockHeight))
```
amimart:
```suggestion
	blockCountSinceTGE := blockHeight - (uint64(startingEmissionsBlockHeight))
```

The naming seems a bit misleading to me, this is not a height

x/mint/keeper/keeper.go
```diff
@@ -315,25 +321,33 @@ func (k Keeper) GetEmissionInfo(ctx context.Context) (*types.Params, *types.Even
...
 	networkStakedTokens, err := GetNumStakedTokens(ctx, k)
 	if err != nil {
 		return nil, nil, errorsmod.Wrap(err, "failed to get number of staked tokens")
 	}
 	monthsAlreadyUnlocked := k.GetMonthsAlreadyUnlocked(ctx)
 	_, lockedVestingTokensPreseed,
-		lockedVestingTokensSeed, lockedVestingTokensTeam, _ := GetLockedVestingTokens(
+		lockedVestingTokensSeed, lockedVestingTokensTeam, _, _, _, err := GetLockedVestingTokens(
 		blocksPerMonth,
 		math.NewIntFromUint64(blockHeight),
```
amimart: This should be `blockHeightTGEAdjusted` I guess, but as I mentioned in another comment I think we should provide both current and TGE heights to be less error prone

x/mint/keeper/keeper_test.go
```diff
@@ -160,6 +166,9 @@ func (s *MintKeeperTestSuite) TestGetEmissionInfo() {
 	err = s.mintKeeper.EcosystemTokensMinted.Set(testCtx, math.ZeroInt())
 	s.Require().NoError(err)
 
+	err = s.mintKeeper.StartingEmissionsBlockHeight.Set(testCtx, 0)
```
amimart: On these tests I think we should test for non-zero values for the starting emissions height as well, the 0 value being a particular case where current height == block count since tge

amimart (changes requested):

x/mint/keeper/emissions.go
xmariachi: yeah this pattern then should be applied to other functions too like GetCirculatingSupply

x/mint/keeper/keeper.go
xmariachi: yep, changed.

x/mint/keeper/keeper.go
xmariachi: here and other occurrence in the function, yes. good catch

amimart (changes requested):
I think we need to increase the `x/mint` consensus version and provide a noop migration

x/mint/keeper/msg_server.go
```diff
@@ -60,10 +91,11 @@ func (ms msgServiceServer) UpdateParams(ctx context.Context, msg *types.UpdatePa
 		if err != nil {
 			return nil, errors.Wrap(err, "error getting args for recalculate")
 		}
+		blockCountSinceTGE := uint64(sdkCtx.BlockHeight() - startingEmissionsBlockHeight)
```
guilherme-brandao, line 96: local var `startingEmissionsBlockHeight` needs to be updated after `SetStartingEmissionsBlockHeight`

guilherme-brandao: Probably a good idea to add more unit tests for the emissions enablement msg/tx.

x/mint/keeper/msg_server.go
```diff
@@ -41,6 +42,36 @@ func (ms msgServiceServer) UpdateParams(ctx context.Context, msg *types.UpdatePa
...
 
+	moduleParams, err := ms.Keeper.GetParams(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting module params")
+	}
+
+	startingEmissionsBlockHeight, err := ms.Keeper.GetStartingEmissionsBlockHeight(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting starting emissions block height")
+	}
```
guilherme-brandao, line 53: It will `error` if not found, right?

guilherme-brandao (changes requested):

x/mint/keeper/msg_server.go
xmariachi: if not found, that's a 0. Should not error.

x/mint/keeper/msg_server.go
xmariachi: > local var `startingEmissionsBlockHeight` needs to be updated after `SetStartingEmissionsBlockHeight`

great catch

xmariachi: pushed changes

xmariachi: added

amimart (approved):
🚀

zale144 (approved):
lgtm

guilherme-brandao (approved):

PR #885, opened 2025-11-07
Add duplicate routes linter on protos

xmariachi:
* Add a linter to detect duplicate routes on protos 
* Added to CI

- [x] If tested, please describe how. If not, why tests are not needed. -- tested against v2 protos that we know that fail. Additional unit tests added.
- [x] If documented, please describe where. If not, describe why docs are not needed. -- no need, just a custom linter, no change in functionality
- [x] Added to `Unreleased` section of `CHANGELOG.md`? -- no need, no change in functionality


spooktheducks (dismissed):

linter/duplicate_routes/internal/duplicateroutes/duplicateroutes.go
```diff
@@ -0,0 +1,244 @@
...
+
+// Finding represents a duplicated route within a protobuf file.
+type Finding struct {
+	File  string
+	Route string
+	Lines []int
+}
+
+// FileSystem encapsulates the filesystem interactions required by the scanner.
+type FileSystem interface {
```
amimart: Instead of redefining that interface and have both `osFS` and `fsAdapter` structs, we can simply use `fs.FS` directly, then we can define:

```golang
func Scan(paths []string) ([]Finding, error) {
	return ScanFS(os.DirFS("/"), paths)
}

func ScanFS(fsys fs.FS, roots []string) ([]Finding, error) {
	return scanWithFilesystem(fsys, roots)
}
```

And when using the filesystem we should call the funcs from the `fs` packages, e.g. `fs.Stat(myFS, path)`, `fs.ReadFile(myFS, path)`, etc..

linter/duplicate_routes/internal/duplicateroutes/duplicateroutes.go
```diff
@@ -0,0 +1,244 @@
...
+func findDuplicates(fsys FileSystem, path string) (map[string][]int, error) {
+	data, err := fsys.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+
+	lines := strings.Split(string(data), "\n")
+	occur := make(map[string][]int)
+
+	for index := 0; index < len(lines); index++ {
```
amimart: To avoid loading the whole file in memory there's already the standard `Scanner` type allowing to read line by line, for example:


```golang
file, err := f.Open(path)
if err != nil {
	return nil, err
}
scanner := bufio.NewScanner(file)
for scanner.Scan() {
	line := scanner.Text()
	...
}
if err := scanner.Err(); err != nil {
	return nil, err
}
```

amimart (commented):
That's nice, just let some remarks I'll let you consider to make it more go idiomatic regarding file system management

linter/duplicate_routes/internal/duplicateroutes/duplicateroutes.go
xmariachi: ty! I'm not sure we need to optimize for memory usage here. Proto files are expected to be small. Keeping it like this also makes it easier to handle for multi-line statements.

linter/duplicate_routes/internal/duplicateroutes/duplicateroutes.go
xmariachi: Changed and added helper functions

amimart (dismissed):

amimart (approved):

spooktheducks (approved):

PR #886, opened 2025-11-08
Fix query via adjusting disabled emissions

xmariachi:
* Fix query emission-info on disabled emissions


x/mint/keeper/emissions.go
```diff
@@ -149,6 +236,45 @@ func GetCirculatingSupply(
 	return circulatingSupply, totalSupply, lockedVestingTokens, ecosystemLocked, updatedMonthsUnlocked, nil
 }
 
+// return the circulating supply of tokens
+func GetCirculatingSupplyDisabledAdjusted(
```
guilherme-brandao, line 240: I'd use something like`GetCirculatingSupplyForDisabledEmissions`, just for clarity. Same thing for the locked vesting tokens query.

guilherme-brandao (commented):
Looks good! A small test for this new query path  would be nice to have.

PR #888, opened 2025-11-13
merge from v0.13.0, v0.14.0 into dev

xmariachi:

Backport latest releases (v0.13.0, v0.14.0) to dev. 
dev already contained a migration (for adding scheduler module), so moving that from to-be-version 13 to 15. 
Tested migration from v0.14.0 to v0.15.0.

-- Ran upgrade and integration tests.
-- No need, just backporting. 
-- Fixed CHANGELOG including latest changes.


amimart (approved):
Looks good 🚀

x/mint/keeper/msg_server.go
```diff
@@ -95,10 +129,15 @@ func (ms msgServiceServer) RecalculateTargetEmission(ctx context.Context, msg *t
 		return nil, errors.Wrap(err, "error getting args for recalculate")
 	}
 
+	startingEmissionsBlockHeight, err := ms.Keeper.GetStartingEmissionsBlockHeight(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting starting emissions block height")
+	}
+	blockCountSinceTGE := uint64(sdkCtx.BlockHeight() - startingEmissionsBlockHeight)
```
zale144, line 136: `blockCountSinceTGE` can be calculated above, so it's done only once

x/mint/module/module.go
```diff
@@ -143,6 +143,13 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
 	if err != nil {
 		panic(fmt.Sprintf("failed to migrate x/%s from version 4 to 5: %v", types.ModuleName, err))
 	}
+	err = cfg.RegisterMigration(types.ModuleName, 5, func(ctx sdk.Context) error {
+		ctx.Logger().Info("MIGRATING %s MODULE FROM VERSION 5 TO VERSION 6", "module", types.ModuleName)
```
zale144, line 147: maybe use `ConsensusVersion-1` and  `ConsensusVersion` so we don't hardcode values here

guilherme-brandao (approved):

x/mint/keeper/msg_server.go
xmariachi: Discussed separately - this happens in two different txs so the comment is not applicable

x/mint/module/module.go
xmariachi: Yes, I think this whole set can be refactored and turned into an iteration (keeping custom actions like the `MigrateStore` for `4`). I think this can be done in a different PR though, and in both emissions and mint modules.

spooktheducks (approved):

PR #889, opened 2025-11-18
Regret boundaries removal + set fixed initial regret cnorm

xmariachi:
* Introduces an initial regret fixed cnorm value, fixed to 0.75
* Removes the regret boundaries
* Implements function `Exp1DivExp1` 
* Uses `Exp1DivExp1` function in `CalcWeightFromNormalizedRegret`

- [x] If tested, please describe how. If not, why tests are not needed. -- via unit tests, and additional tests
- [ ] If documented, please describe where. If not, describe why docs are not needed. -- no need, impl details, intention is not changed
- [ ] Added to `Unreleased` section of `CHANGELOG.md`?  -- not yet, current CHANGELOG in dev is pending update from v0.13.0, v0.14.0 releases.


math/dec.go
```diff
@@ -775,3 +775,187 @@ func decimalPlaces(d *apd.Decimal) uint32 {
...
+		if err != nil {
+			return Dec{}, errorsmod.Wrapf(err, "error computing final division")
+		}
+
+		return result, nil
+	}
+
+	// Case 2: a > 0 && b <= 0
+	// Formula: e^a * (e^{-a} + 1) / (e^b + 1) == (e^a + e^0) / (e^b + 1)
+	if (a.IsPositive()) && (b.Lte(zero)) {
```
fstecker: Is there a reason why you don't just use `if a.IsPositive() && !b.IsPositive()`? I think that would make the 4 cases more obvious.

fstecker (dismissed):
Looks great! I hope you had some AI help you implement this `Exp1DivExp1` function, it looks very repetitive otherwise!

math/dec.go
xmariachi: Yes - there is a catch: `!isPositive()` means  “b is either zero, negative, or non-finite”, so it wouldn't be exactly the same.
However, using `isNegative()` would leave the zero case out. 
So `Lte(zero)` seems to be the correct way to express it, right?

fstecker: Ok good point! I hope there is something which prevents us from getting infinities here (or more importantly, from the regret becoming infinite). But assuming they can appear, the correct way to handle them is `a > 0` includes +infinity and `a <= 0` includes -infinity. If I read this right, `isPositive` just checks the `Negative` flag of the `apd.Decimal`, and that's independent of the `Form` field which says whether the values is infinite. So -infinity has `Negative` set and infinity doesn't. If that's right, `if a.IsPositive() && !b.IsPositive()` should give the right behavior. `if a.Gt(zero) && b.Lte(zero)` is of course also good. I just found it strange to mix the two.

xmariachi:
Pending re-checking with re-generated data from simulation

math/dec.go
xmariachi: Applied uniformity via `if a.Gt(zero) && b.Lte(zero)` style

math/dec.go
```diff
@@ -775,3 +775,187 @@ func decimalPlaces(d *apd.Decimal) uint32 {
 func shouldBeNaN(d *apd.Decimal) bool {
 	return d.Form == apd.NaN || d.Form == apd.NaNSignaling
 }
+
+// Exp1DivExp1 computes (e^a + 1) / (e^b + 1) with numerical stability optimizations.
+// It uses different mathematical transformations based on the signs of a and b to avoid
+// overflow/underflow issues that can occur with large positive or negative values.
+func Exp1DivExp1(a, b Dec) (Dec, error) {
```
amimart: Maybe better to place it under `exp.go` with other exponential funcs.

For better accuracy we could use the `exp()` func taking the `apd.Decimal` type and round the result in the end, this may be helpful to reduce allocations as well.

As it replaces the `Gradient` func it's the bottleneck on performance issues when computing network inferences, and we have more exponential calculations so it may be worth to optimize using the following properties:

- `exp(-x) == 1 / exp(x)`
- `exp(a-b) == exp(a) / exp(b)`

This could look like:
```go
func Exp1DivExp12(a, b Dec) (Dec, error) {
	if a.IsNaN() || b.IsNaN() {
		return Dec{}, errorsmod.Wrapf(ErrNaN, "cannot compute Exp1DivExp1 with NaN arguments: a=%s, b=%s", a.String(), b.String())
	}

	var d apd.Decimal
	if err := exp1DiveExp1(&d, &a.dec, &b.dec); err != nil {
		return Dec{}, errorsmod.Wrap(err, "error computing exponential")
	}

	if _, err := fnRoundCtx.Round(&d, &d); err != nil {
		return Dec{}, errorsmod.Wrap(err, "error computing exponential")
	}

	return Dec{dec: d, isNaN: false}, nil
}

func exp1DiveExp1(d, a, b *apd.Decimal) error {
	var expA, expB, num, den apd.Decimal
	if err := exp(&expA, a); err != nil {
		return err
	}
	if err := exp(&expB, b); err != nil {
		return err
	}

	ed := apd.MakeErrDecimal(&fnCalcCtx)

	if a.Cmp(zeroDec) <= 0 && b.Cmp(zeroDec) <= 0 {
		// (exp(a) + 1) / (exp(b) + 1)
		ed.Add(&num, &expA, oneDec)
		ed.Add(&den, &expB, oneDec)
	} else if a.Cmp(zeroDec) > 0 && b.Cmp(zeroDec) <= 0 {
		// exp(a) * (exp(-a) + 1) / (exp(b) + 1)
		var expNegA apd.Decimal
		ed.Quo(&expNegA, oneDec, &expA)

		ed.Add(&num, &expNegA, oneDec)
		ed.Mul(&num, &expA, &num)
		ed.Add(&den, &expB, oneDec)
	} else if a.Cmp(zeroDec) <= 0 && b.Cmp(zeroDec) > 0 {
		// exp(-b) * (exp(a) + 1) / (exp(-b) + 1)
		var expNegB apd.Decimal
		ed.Quo(&expNegB, oneDec, &expB)

		ed.Add(&num, &expA, oneDec)
		ed.Mul(&num, &expNegB, &num)
		ed.Add(&den, &expNegB, oneDec)
	} else {
		// exp(a-b) * (exp(-a) + 1) / (exp(-b) + 1)
		var expAMinusB, expNegA, expNegB apd.Decimal
		ed.Quo(&expAMinusB, &expA, &expB)
		ed.Quo(&expNegB, oneDec, &expB)
		ed.Quo(&expNegA, oneDec, &expA)

		ed.Add(&num, &expNegA, oneDec)
		ed.Mul(&num, &expAMinusB, &num)
		ed.Add(&den, &expNegB, oneDec)
	}

	ed.Quo(d, &num, &den)
	return ed.Err()
}
```

amimart (commented):
Nice! I've let some potential optimization remarks, as this is a sensible part of the network inference computation

math/dec.go
fstecker: The last suggestion defeats the purpose of this function: it was meant to avoid infinities / very large intermediate results by (almost) never taking the exponent of a positive number. Yes, the cost is having to evaluate 3 instead of 2 exps in most cases, but the robustness is more important than performance here (this thing apparently caused real problems in the simulator ~1.5 years ago).

To be clear, I have nothing against the other ideas, rounding at the end, avoiding allocations, etc.

amimart: Oh ok I see thanks for the clarification!

xmariachi: Yep, moved, makes sense.

However re. rounding using the `apd.Decimal`: isn't that what the `Exp()` function already does [here](https://github.com/allora-network/allora-chain/blob/f8852723c4b0e692f54889a3b72e843237fc3c92/math/exp.go#L78-L96) ?

amimart: Yes I was thinking for better accuracy, as we are using Exp 2/3 times it may be better to round only once at the end. It shouldn't be a big gain though

test/testutil/inEpsilon.go
```diff
@@ -60,10 +60,18 @@ func InEpsilon3(t *testing.T, value alloraMath.Dec, target string) {
...
+	epsilonDec := alloraMath.MustNewDecFromString("0.0001")
+	targetDec := alloraMath.MustNewDecFromString(target)
+	InEpsilon(t, value, targetDec, epsilonDec)
+}
+
+// unused:
+// func InEpsilon4Dec(t *testing.T, value alloraMath.Dec, target alloraMath.Dec) {
+// 	t.Helper()
+// 	InEpsilon(t, value, target, alloraMath.MustNewDecFromString("0.0001"))
+// }
```
amimart:
```suggestion
```

math/dec_test.go
```diff
@@ -1317,3 +1317,439 @@ func TestUnmarshalJSONNaN(t *testing.T) {
 	require.NoError(t, unmarshaled.UnmarshalJSON(json))
 	require.Equal(t, nan, unmarshaled)
 }
+
+func TestExp1DivExp1(t *testing.T) {
```
amimart: This can go in `exp_test.go` as well :)

amimart (dismissed):
Looks good, made some small remarks but preemptively approving

amimart (approved):

x/emissions/keeper/inference_synthesis/weight.go
```diff
@@ -39,42 +39,42 @@ type CalcRegretStdDevFilteredByWeightsArgs struct {
...
 }
 
-// Gradient cache implementation
+// Math helper cache implementation
+// Used to memoize expensive helper computations like Gradient and Exp1DivExp1.
 var (
-	gradientCache = make(map[string]alloraMath.Dec)
-	cacheEnabled  = false
+	mathHelperCache  = make(map[string]alloraMath.Dec)
+	mathCacheEnabled = false
```
zale144, line 46: I guess I can only dream of these not being global

zale144 (approved):

x/emissions/keeper/inference_synthesis/weight.go
xmariachi: Good point. I'm taking a look at this but I don't think it's a blocker for this fix

xmariachi: To be done in a separate PR to limit the scope.

PR #893, opened 2025-12-09
Editable per topic c norm

guilherme-brandao:

- Move `c_norm` from global params to a per-topic field, updating proto definitions, tx messages, keeper/msgserver logic, and synthesis/regret calculations to read from `topic.CNorm`.
- Add migration v12 to drop global `c_norm` and copy the previous value into every existing topic using a store iterator to avoid gaps.

https://linear.app/alloralabs/issue/ENGN-4064/editable-per-topic-c-norm


x/emissions/proto/emissions/v4/topic.proto
```diff
@@ -0,0 +1,87 @@
+syntax = "proto3";
+package emissions.v4;
```
xmariachi: I think this `topic.proto` file should be defined in v9, since it's only existing from v9. It will also simplify previous vX which will use v3.Topic, until v9 that will use v9.

x/emissions/migrations/v7/migrate.go
```diff
@@ -5,6 +5,7 @@ import (
 	storetypes "cosmossdk.io/store/types"
 	"github.com/allora-network/allora-chain/x/emissions/keeper"
 	oldV6Types "github.com/allora-network/allora-chain/x/emissions/migrations/v7/oldtypes"
+	v8Types "github.com/allora-network/allora-chain/x/emissions/migrations/v8/oldtypes"
```
xmariachi: Good catch - though these are the "old types of v8", which effectively are `v7`, right? 
Then I think we should call them `oldV7Types` (or just `v7Types`) instead of `v8Types`.
(This applies to all of the `v*/migrate*.go`)

x/emissions/module/module.go
```diff
@@ -44,7 +45,7 @@ var (
 )
 
 // ConsensusVersion defines the current module consensus version.
-const ConsensusVersion = 11
+const ConsensusVersion = 12
```
xmariachi: FYI Consensus Version will be at 13... when we merge back #888  :P

xmariachi: ... so, this will be integrated in the consensus version and migration there, once merged

x/emissions/keeper/keeper.go
```diff
@@ -3333,6 +3333,15 @@ func (k *Keeper) UpdateTopic(ctx context.Context, msg *types.UpdateTopicRequest)
...
+		cNorm, err := alloraMath.NewDecFromString(msg.CNorm[0])
+		if err != nil {
+			return types.Topic{}, errorsmod.Wrap(err, "failed to parse c_norm")
+		}
+		updatedTopic.CNorm = cNorm
+		hasChanges = true
+	}
+
 	if !hasChanges {
 		return topic, nil
```
xmariachi: Maybe this could be done in the original  #831 instead

xmariachi (commented):
Some comments on protos and migration mostly

x/emissions/migrations/v14/migrate.go
```diff
@@ -0,0 +1,179 @@
+package v14
```
xmariachi: Why v14? 
This should be v12, right? (after current, latest one which is v11) 

Also, since here we have breaking changes in the API (due to the removal of `c_norm` from Params), shouldn't we push to a new version (ie moving the new proto files to `x/emissions/proto/emissions/v10`)? 

(either that, or only deprecate instead of removing the c_norm in the params)

x/emissions/migrations/v6/migrate.go
```diff
@@ -60,7 +61,7 @@ func MigrateParams(store storetypes.KVStore, cdc codec.BinaryCodec) error {
 	// DIFFERENCE BETWEEN OLD PARAMS AND NEW PARAMS:
 	// ADDED:
 	//      GlobalWhitelistEnabled, TopicCreatorWhitelistEnabled
-	newParams := emissionstypes.Params{ //nolint: exhaustruct
+	newParams := v7Types.Params{ //nolint: exhaustruct
```
xmariachi: Should be v6Types (ie "v7 old types")

x/emissions/migrations/v5/migrate.go
```diff
@@ -72,7 +73,7 @@ func MigrateParams(store storetypes.KVStore, cdc codec.BinaryCodec) error {
 	// DIFFERENCE BETWEEN OLD PARAMS AND NEW PARAMS:
 	// ADDED:
 	//      InitialRegretQuantile, PNormSafeDiv
-	newParams := emissionstypes.Params{ //nolint: exhaustruct
+	newParams := v6Types.Params{ //nolint: exhaustruct
```
xmariachi: Should be v5Types (ie v6 old types)

x/emissions/migrations/v3/migrate.go
```diff
@@ -73,7 +74,7 @@ func MigrateParams(store storetypes.KVStore, cdc codec.BinaryCodec) error {
 	//      MaxRetriesToFulfilNoncesWorker
 	//      MaxRetriesToFulfilNoncesReputer
 	//      MaxTopicsPerBlock
-	newParams := types.Params{ //nolint: exhaustruct // not sure if safe to fix, also this upgrade has already happened.
+	newParams := v4Types.Params{ //nolint: exhaustruct // not sure if safe to fix, also this upgrade has already happened.
```
xmariachi: Should be v3Types (ie v4 old types)

x/emissions/migrations/v8/migrate.go
```diff
@@ -53,7 +54,7 @@ func MigrateParams(ctx sdk.Context, store storetypes.KVStore, cdc codec.BinaryCo
 	// DIFFERENCE BETWEEN OLD PARAMS AND NEW PARAMS:
 	// ADDED:
 	//       MinWeightThresholdForStdnorm
-	newParams := emissionstypes.Params{ //nolint: exhaustruct
+	newParams := v9Types.Params{ //nolint: exhaustruct
```
xmariachi: Should be v8Types (ie v9 old types)

x/emissions/migrations/v9/migrate.go
```diff
@@ -48,10 +49,8 @@ func MigrateParams(ctx sdk.Context, store storetypes.KVStore, cdc codec.BinaryCo
 		return errorsmod.Wrapf(err, "failed to unmarshal old parameters")
 	}
 
-	// defaultParams := emissionstypes.DefaultParams()
-
 	// DIFFERENCE BETWEEN OLD PARAMS AND NEW PARAMS:
-	newParams := emissionstypes.Params{ //nolint: exhaustruct
+	newParams := v9Types.Params{ //nolint: exhaustruct
```
xmariachi: Should be v8Types (ie v9 old types)

x/emissions/migrations/v14/migrate.go
```diff
@@ -0,0 +1,179 @@
...
+	ctx sdk.Context,
+	store storetypes.KVStore,
+	cdc codec.BinaryCodec,
+	oldCNorm alloraMath.Dec,
+) error {
+	topicStore := prefix.NewStore(store, emissionstypes.TopicsKey)
+	iterator := topicStore.Iterator(nil, nil)
+	defer iterator.Close()
+
+	ctx.Logger().Info("MIGRATION V12: Migrating topics to add CNorm", "cNorm", oldCNorm)
```
zale144: v12?

x/emissions/migrations/v14/migrate.go
xmariachi: Following on this. The migrations are of the emissions module version, not the overall chain consensus version migration. This could be what is causing confusion maybe?

guilherme-brandao: We decided in the last protocol meeting to use v14 in the migration to be in line with the `ConsensusVersion`.

Also to deprecate `c_norm` property in params instead of creating a new version of the proto.

We can discuss in Slack any further questions @xmariachi

x/emissions/module/module.go
```diff
@@ -44,7 +45,7 @@ var (
 )
 
 // ConsensusVersion defines the current module consensus version.
-const ConsensusVersion = 13
+const ConsensusVersion = 14
```
xmariachi: In latest release v0.14.0, the `emissions` module version was at 12 (see [here](https://github.com/allora-network/allora-chain/blob/release-v0.14.0/x/emissions/module/module.go) ).

So this consensus version was already pushed in previous PRs, and should stay at 13.

x/emissions/module/module.go
```diff
@@ -160,6 +161,11 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
 	}); err != nil {
 		panic(fmt.Sprintf("failed to migrate x/%s from version 12 to 13: %v", types.ModuleName, err))
 	}
+	if err := cfg.RegisterMigration(types.ModuleName, 13, func(ctx sdk.Context) error {
```
xmariachi: ...and following the [previous comment](https://github.com/allora-network/allora-chain/pull/893/changes#r2641115021), this would not be needed.

x/emissions/migrations/v14/migrate.go
xmariachi: Since we deprecate then there's no need to push that version, right? 
However just to underline that is "13· not "14" :P but ig now it doesnt have an effect, right? If so please resolve

xmariachi (changes requested):
changes requested wrt version

xmariachi (approved):

PR #894, opened 2025-12-09
Fix potential active topic set issue

xmariachi:
* Fix potential active topic set with tests
* Improve logic 

- [x] If tested, please describe how. If not, why tests are not needed. -- added unit tests 
- [ ] If documented, please describe where. If not, describe why docs are not needed. -- not needed, no change in functionality


amimart (changes requested):
If this happens it means we have a big state inconsistency issue so I'd rather error and let the chain halt or the tx to fail

amimart (dismissed):

guilherme-brandao (dismissed):

amimart (approved):

guilherme-brandao (approved):

PR #897, opened 2025-12-23
Turn math cache into its own object

xmariachi:
Turning math cache into an object instead of using global vars.

- [x] If tested, please describe how. If not, why tests are not needed. -- Added extensive unit tests, in both the cache and the usage.
- [x] If documented, please describe where. If not, describe why docs are not needed. -- It is a performance change, not a functionality change, so it should not need its own documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,7 +68,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * [#851](https://github.com/allora-network/allora-chain/pull/851) Enforce recommended settings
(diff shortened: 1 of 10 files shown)


spooktheducks (changes requested):
Per comments in private messages with @jmdkastro and @xmariachi , I say we hold on approvals here pending:
- solid benchmarking showing that this cache is actually necessary (as opposed to just causing code bloat)
- a more performantly constructed key for the cache map (we can reduce `Decimal` into hashable bytes much more efficiently than converting to string with `fmt.Sprintf`)
- an understanding as to why we're fragmenting the cache, which somewhat negates the effects of caching math in the first place

PR #905, opened 2026-01-28
docs: fix positive boundary comments

Olexandr88:
Fix misleading comments for positive boundary helpers to match actual behavior and function names


xmariachi (dismissed):
good catch.

Olexandr88:
@amimart, could you please review this? This is just a documentation change that corrects misleading comments

PR #910, opened 2026-02-10
Regret norm to use MAD

xmariachi:
* Use of MAD instead of stddev to further reduce impact of outliers
* Removed side-effect of sorting on the input parameter array on `Median` and `GetQuantileOfDecs`. 
* Changed unit tests values where applicable.

- [x] If documented, please describe where. If not, describe why docs are not needed. -- not needed, an improvement on internal logic which keeps the original intention.


guilherme-brandao (approved):
nice work!

math/utils.go
```diff
@@ -228,19 +228,23 @@ func Median(data []Dec) (Dec, error) {
 		return ZeroDec(), nil
 	}
 
-	// Sort the data
-	slices.SortFunc(data, func(x, y Dec) int {
+	// Create a copy to avoid modifying the original slice
```
amimart, line 231:
```suggestion
	// Create a copy to avoid modifying the original array
```

Not really important, but more precise 😄. The slice only holds a pointer to the array, and it's the array we don't want to modify

x/emissions/keeper/actor_utils/losses.go
```diff
@@ -233,23 +233,23 @@ func CloseReputerNonce(
 		return err
 	}
 	// 2.b ... and store it.
-	err = k.SetLatestRegretStdNorm(ctx, topic.Id, stdDevPlusEpsilon)
+	err = k.SetLatestRegretStdNorm(ctx, topic.Id, regretScalePlusEpsilon)
```
amimart: I think this method and the underlying store should be renamed accordingly to avoid any confusion. The store key can be kept untouched to avoid a migration

x/emissions/keeper/inference_synthesis/weight.go
```diff
@@ -124,25 +129,31 @@ func CalcRegretStdDevFilteredByWeights(args CalcRegretStdDevFilteredByWeightsArg
...
-func CalcStdDevPlusEpsilon(regrets []alloraMath.Dec, epsilonTopic alloraMath.Dec) (alloraMath.Dec, error) {
-	// Calc std dev of regrets + epsilon
-	// σ(R_ijk) + ε
-	stdDevRegrets, err := alloraMath.StdDev(regrets)
+// CalcRegretScalePlusEpsilon calculates the MAD-based scale (scaled to match stddev under normality)
+// of the regrets provided plus epsilon.
+func CalcRegretScalePlusEpsilon(regrets []alloraMath.Dec, epsilonTopic alloraMath.Dec) (alloraMath.Dec, error) {
+	// Calc MAD of regrets, scaled to match stddev under normality, + epsilon
+	// madToStdDevFactor * MAD(R_ijk) + ε
+	regretsCopy := append([]alloraMath.Dec(nil), regrets...)
```
amimart: I don't think the copy is needed as we don't modify it

amimart (commented):
Nice, I've let some suggestions I'll let up to you

math/utils.go
xmariachi: Good point, though `data` here is a slice itself so I think the original is stil more precise. If data would define its size then it would be an array - but it is defined as a slice there.
 

> In Go terminology, this comment is more correct as written (using "slice").

> Here's why:
> 
> Slice vs Array distinction: In Go, slices and arrays are different types. A slice is a dynamic view into an array with a pointer, length, and capacity. Arrays have a fixed size. The data parameter in this function is a slice ([]Dec), not an array ([n]Dec).
> 
> The operation: When you do dataCopy := make([]Dec, n) and copy(dataCopy, data), you're creating a copy of the slice's underlying data, not modifying the original slice header itself. However, the semantic intent of the comment is about not modifying the underlying data that the original slice points to.
> 
> Go convention: Go developers typically refer to []Type as a "slice" in documentation and comments, which is the technically precise term.
> 
> So the current comment saying "slice" is the more idiomatic and technically correct terminology for Go code. While someone unfamiliar with Go might say "array," Go developers consistently use "slice" when referring to types created with make() or slice literals like []Dec.

x/emissions/keeper/actor_utils/losses.go
xmariachi: Yeah, maybe that is better. Ty, will do.

x/emissions/keeper/inference_synthesis/weight.go
xmariachi: great catch

x/emissions/keeper/actor_utils/losses.go
```diff
@@ -206,17 +206,17 @@ func CloseReputerNonce(
...
 
-	// 2. Calculate the regret_stdnorm to be used in
-	// 2.a Calculate the regret_stdnorm filtered by ∫the previous weights. If not, apply stddev.
-	stdDevPlusEpsilon, err := synth.CalcRegretStdDevFilteredByWeights(
-		synth.CalcRegretStdDevFilteredByWeightsArgs{
+	// 2. Calculate the regret scale (legacy "stdnorm") to be used.
+	// 2.a Calculate the regret scale filtered by the previous weights. If not, apply MAD-based scale.
+	regretScalePlusEpsilon, err := synth.CalcRegretScaleFilteredByWeights(
+		synth.CalcRegretScaleFilteredByWeightsArgs{
 			Ctx:                 ctx,
```
fstecker: I think we can (and should) replace `CalcRegretScaleFilteredByWeights` by `CalcRegretScalePlusEpsilon` here. The reason we did weight filtering in the first place was because outliers affected the standard deviation a lot. But the MAD is naturally resistant to outliers, so that doesn't apply anymore. There was a little discussion about this in the linear ticket ~3 weeks ago.

This is actually the only call to `CalcRegretScaleFilteredByWeights` so if we remove it here we can get rid of the entire thing. Which might also make things easier for the topic lifecycle changes as it removes a dependency between epochs via weights.

xmariachi: done, pls check

fstecker (approved):
Looks great, thanks Diego!

zale144 (approved):
lgtm

PR #911, opened 2026-02-11
refactor(utils): use fmt.Errorf in EnsureDirAndMaxPerms

Olexandr88:
Small cleanup to use idiomatic fmt.Errorf instead of errors.New(fmt.Sprintf(...)) and remove unused import.  
No behavior changes — just makes the code simpler and more Go-idiomatic.

### DIFF
--- a/utils/files.go
+++ b/utils/files.go
@@ -1,7 +1,6 @@
 package utils
 
 import (
-	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


amimart (approved):
Thx :)

spooktheducks:
Incorporated in #914

PR #927, opened 2026-03-04
fix one-in forecaster map wiring and remove dead forecast map code

xmariachi:

**Summary**
* Use `singleForecastImpliedInference` in `calcOneInValue` so one-in computation is scoped to the selected forecaster.
* Remove unused one-in plumbing (`singleForecasterToForecast` and unused internal arg field).
* Add a regression test for nil forecast-implied entries in one-in forecaster inference path.

**Validation**
go test ./x/emissions/keeper/inference_synthesis -count=1

Tests pass, added test
no need, no func change
no need, no func change

---
## Summary by cubic
Aligns one-in forecaster computation to the selected forecaster’s forecast-implied inference and remove dead forecast map wiring. Addresses Linear ENGN-5777 by fixing map wiring and hardening nil-entry handling.

- **Bug Fixes**
  - Pass singleForecastImpliedInference into calcWeightedInference for one-in.
  - Guard against nil entries when building the single-forecaster implied inference map.
  - Add regression test to confirm nil forecast-implied entries do not panic.

- **Refactors**
  - Remove singleForecasterToForecast and the ForecasterToForecast arg from calcOneInValueArgs.
  - Simplify one-in path inputs; existing epoch outputs remain unchanged.

<sup>Written for commit e34054d64814eb6a0fec17edfe1dd689a092b4db. Summary will update on new commits.</sup>

### DIFF
--- a/x/emissions/keeper/inference_synthesis/network_inference_builder.go
+++ b/x/emissions/keeper/inference_synthesis/network_inference_builder.go
@@ -693,7 +693,6 @@ type calcOneInValueArgs struct {
 	AllInferersAreNew                    bool
 	Inferers                             []Inferer
 	InfererToInference                   map[Inferer]*emissions.Inference
-	ForecasterToForecast                 map[Forecaster]*emissions.Forecast
 	ForecasterToForecastImpliedInference map[Forecaster]*emissions.Inference
 	EpsilonTopic                         alloraMath.Dec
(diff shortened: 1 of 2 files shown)


zale144:
I believe I already fixed this in the classification PR. let's leave this for later, we could still use the unit test

PR #929, opened 2026-03-09
classification #6 - vectorize inference synthesis

zale144:

https://linear.app/alloralabs/issue/ENGN-5155/6-vectorize-inference-synthesis
https://linear.app/alloralabs/issue/ENGN-5156/7-networkinferencebundle-plumbing

## Still Left Todo

---
## Summary by cubic
Vectorized inference synthesis for classification topics and finished the end-to-end move to `NetworkInferenceBundle` across storage, queries, events, and genesis. Removed scalar `Inference.value`, enforced per‑nonce label registries (incl. `SINGLE`=1 label), and tightened validation, meeting ENGN-5155 acceptance and ENGN-5156 plumbing.

- New Features
  - Vectorized combined, forecast‑implied, naive, and one‑out/one‑in inferences; regrets/weights stay scalar.
  - Keeper writes/reads bundles via `InsertNetworkInferenceBundle` with new bundle stores (incl. outlier‑resistant); query server returns bundles directly.
  - New `EventNetworkInferenceBundle` with labeled values; emitters and event utils use bundle format.
  - Per‑nonce epoch label registry required; `DecMatrix` is now `[]DecArray`; inference `values` validated against the registry.

- Migration
  - v14 migrates stored `network_inferences` to new `NetworkInferenceBundle` stores and converts legacy `ValueBundle` entries; genesis accepts both and persists bundles.
  - Register epoch labels for every nonce. For `SINGLE`, register one label and submit `values` with one element; `value` is no longer accepted.
  - If `RequireUnity` is true, use `OutputArity = MULTI` and submit values that sum to 1 within `UnityTolerance`.

<sup>Written for commit 9a4afd73333248330cea488e56cf7204b942c444. Summary will update on new commits.</sup>


x/emissions/keeper/actor_utils/worker_test.go
```diff
@@ -60,28 +60,47 @@ func (s *WorkerTestSuite) TestCloseWorkerNonce() {
...
-			},
-			{
-				Inferer:     worker1,
-				Value:       alloraMath.MustNewDecFromString("-0.07333303938740420"),
-				Values:      nil,
-				TopicId:     topicId,
-				BlockHeight: blockHeight,
-			},
+	// MULTI: ensure epoch label registry exists for this nonce (and has labels)
+	// (Pick whatever labels your tests commonly use)
```
xmariachi: (unnecessary agentic comment?)

x/emissions/keeper/inference_synthesis/forecast_implied.go
```diff
@@ -1,16 +1,22 @@
...
 	alloraMath "github.com/allora-network/allora-chain/math"
+	"github.com/allora-network/allora-chain/x/emissions/keeper"
 	emissionstypes "github.com/allora-network/allora-chain/x/emissions/types"
 )
 
 // CalcForecastImpliedInferencesArgs holds inputs for CalcForecastImpliedInferences.
 type CalcForecastImpliedInferencesArgs struct {
 	Logger                 log.Logger
+	Ctx                    context.Context
+	K                      keeper.Keeper
```
xmariachi: I don't think we want to pass the whole keeper here. 
The struct defines many "small" dependencies, but it does it this way so the dependencies for the process are clear, ie it is easier to see what the process r`CalcForecastImpliedInferences` really requires.

x/emissions/keeper/inference_synthesis/forecast_implied.go
```diff
@@ -45,165 +52,208 @@ func CalcForecastImpliedInferences(args CalcForecastImpliedInferencesArgs) (
...
+
+			for _, inferer := range sortedInferersInForecast {
+				inf := args.InfererToInference[inferer]
+
+				if blockHeight == 0 {
+					blockHeight = inf.BlockHeight
 				}
 
-				medianValue, err := alloraMath.Median(inferenceValues)
+				iv, err := keeper.InferenceValuesFromProto(topic, reg, inf)
```
xmariachi: I suggest "conversion" operations should contain "convert" (or similar wording) in the function name, to describe with more clarity that this is a conversion only function (no additional logic applied). 

Separately - is "keeper" the right package that which the "conversions" should belong to? If there are several conversions, maybe they should be on their own sub-package. Just a thought, not sure it does apply @zale144

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -208,6 +224,8 @@ func GetOneOutInfererForecastImpliedInferences(args GetOneOutInfererForecastImpl
 				CalcForecastImpliedInferencesArgs{
 					Logger:                 args.Logger,
 					TopicId:                args.TopicId,
+					Ctx:                    args.Ctx,
```
xmariachi: Likewise previous comment, ctx should be passed in function and keeper is too broad a dependency.

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -234,19 +253,17 @@ func GetOneOutInfererForecastImpliedInferences(args GetOneOutInfererForecastImpl
...
-				Value:  forecastImpliedInference.Value,
-			})
-		}
+			oneOutInfererValues, err = emissionskeeper.InferenceValuesToLabeledValues(forecastImpliedInference.Values, args.LabelRegistry)
+			if err != nil {
+				return nil, errorsmod.Wrap(err, "failed to convert forecast implied inference values")
+			}
 
-		// Only add this forecaster's results if we calculated values for at least one withheld inferer
-		if len(oneOutInfererValues) > 0 {
```
xmariachi, line 244: what is the reason for not keeping this guard here? Is this covered above?

x/emissions/keeper/inference_synthesis/network_inference_builder.go
xmariachi, line 286: OK, the K was passed here... but I find it inconsistent. 
One of the ideas of these args objects is to clearly define the dependencies needed for the function. 

Here, K is only used to get the 
```regret, _, err := args.K.GetNaiveInfererNetworkRegret(args.Ctx, args.TopicId, inferer)```
but this `infererToRegret := make(map[string]*alloraMath.Dec)` could be read before, and passed as an argument, so we don't need to pass the whole keeper.

File: x/emissions/keeper/inference_synthesis/network_inference_builder.go [hunk 6]
```diff
@@ -387,6 +413,8 @@ func calcOneOutInfererInference(args CalcOneOutInfererInferenceArgs) (
 			CalcForecastImpliedInferencesArgs{
 				Logger:                 args.Logger,
 				TopicId:                args.TopicId,
+				Ctx:                    args.Ctx,
+				K:                      args.K,
```
xmariachi: Also here, ctx can be passed separately, and is K really needed? We can get and pass these instead, right?

```
remainingInfererRegrets
forecasterToForecastImpliedInference 
```

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -670,88 +709,94 @@ func GetOneOutForecasterInferences(args GetOneOutForecasterInferencesArgs) (
...
-		regret, _, err := args.K.GetOneInForecasterNetworkRegret(args.Ctx, args.TopicId, args.OneInForecaster, inferer)
-		if err != nil {
-			return alloraMath.Dec{}, errorsmod.Wrapf(err, "CalcOneInValue() error getting one-in forecaster regret")
+		inf, ok := args.InfererToInference[inferer]
+		if !ok || inf == nil {
+			args.Logger.Debug("calcOneInValues: missing inferer inference, skipping", "inferer", inferer)
+			continue
 		}
-		infererToRegretForSingleForecaster[inferer] = &regret.Value
+		infererToInferenceForSingleForecaster[inferer] = inf
```
xmariachi, line 811: The order in which `infererToInferenceForSingleForecaster` and `infererToRegretForSingleForecaster` appear has been swapped, is there a reason for this? (don't really care either way, just want to understand)

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -316,13 +334,20 @@ func GetNaiveInference(args GetNaiveInferenceArgs) (naiveInference alloraMath.De
 		forecasterToForecastImpliedInference: args.ForecasterToForecastImpliedInference,
 		weights:                              weights,
 		epsilonSafeDiv:                       args.EpsilonSafeDiv,
+		numLabels:                            len(args.LabelRegistry.GetLabels()),
```
xmariachi: ¿Is it guaranteed that LabelRegistry is not nil? Maybe a check upstream but doing Defense In Depth, we should make sure it gets caught early.
 
Also: since the LabelRegistry is immutable throughout the computations, this len value should be computed once and passed in, so we're not having to compute this every time (this `len(args.LabelRegistry.GetLabels())` appears five times in this file)

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -75,35 +79,46 @@ func GetCombinedInference(args GetCombinedInferenceArgs) (
...
 	inferers []Worker,
 	infererToInference map[Worker]*emissions.Inference,
-) (infererValues []*emissions.WorkerAttributedValue) {
-	infererValues = make([]*emissions.WorkerAttributedValue, 0)
+	reg *emissions.EpochLabelRegistry,
+) (infererValues []*emissions.WorkerInference, _ error) {
+	infererValues = make([]*emissions.WorkerInference, 0)
 	for _, inferer := range inferers {
-		infererValues = append(infererValues, &emissions.WorkerAttributedValue{
+		inferenceValues, err := emissionskeeper.InferenceValuesToLabeledValues(infererToInference[inferer].Values, reg)
```
xmariachi: Can `infererToInference[inferer]` be nil?
Probably add a guard to emit an explicit error if this is reached, instead of panic when accessing `.Values`.

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -75,35 +79,46 @@ func GetCombinedInference(args GetCombinedInferenceArgs) (
...
 			logger.Warn("getForecastImpliedInferences() no forecast-implied inference", "forecaster", forecaster)
 			continue
 		}
-		forecastImpliedInferences = append(forecastImpliedInferences, &emissions.WorkerAttributedValue{
+		inferenceValues, err := emissionskeeper.InferenceValuesToLabeledValues(forecasterToForecastImpliedInference[forecaster].Values, reg)
+		if err != nil {
+			return []*emissions.WorkerInference{}, errorsmod.Wrap(err, "failed to convert inference values")
+		}
+		forecastImpliedInferences = append(forecastImpliedInferences, &emissions.WorkerInference{
 			Worker: forecasterToForecastImpliedInference[forecaster].Inferer,
```
xmariachi: I think here 
`forecasterToForecastImpliedInference[forecaster].Inferer` 
should be the same as `forecaster`, right? 
And it would be more clear to what the intention is -- seeing an "Inferer" there, while correct (it is the forecast-implied inference of this forecast, so in that sense the forecaster is the inferer here), it may add more confusion than simply using `forecaster`.

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -234,19 +253,17 @@ func GetOneOutInfererForecastImpliedInferences(args GetOneOutInfererForecastImpl
 			}
 
 			// Add to our results
-			oneOutInfererValues = append(oneOutInfererValues, &emissions.WithheldWorkerAttributedValue{
-				Worker: withheldInferer,
-				Value:  forecastImpliedInference.Value,
-			})
-		}
+			oneOutInfererValues, err = emissionskeeper.InferenceValuesToLabeledValues(forecastImpliedInference.Values, args.LabelRegistry)
```
xmariachi: This variable is created outside the loop, but gets reassigned here every time before being used below. 
So, it should not be created outside and reassigned, instead it should be just created here and then used in 261 right after, inside the scope of the loop.

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -670,88 +709,94 @@ func GetOneOutForecasterInferences(args GetOneOutForecasterInferencesArgs) (
...
 	args.Logger.Debug("Calculating one-in inference", "forecaster", args.OneInForecaster)
 
-	// In each loop, remove all forecast-implied inferences except one
-	singleForecastImpliedInference := make(map[Worker]*emissions.Inference, 1)
-	singleForecastImpliedInference[args.OneInForecaster] = args.ForecasterToForecastImpliedInference[args.OneInForecaster]
+	// Need the forecaster's forecast-implied inference to do "one-in".
+	fi, ok := args.ForecasterToForecastImpliedInference[args.OneInForecaster]
+	if !ok || fi == nil {
+		return alloraMath.DecArray{}, errorsmod.Wrapf(
+			types.ErrInvalidRequest,
```
xmariachi: Should be `sdkerrors.ErrInvalidRequest` for alignment with the rest of the code in `emissions` module.

xmariachi (commented):
I've done a thorough revision on the network inference synthesis changes on `network_inference_builder.go` along with other areas. 
The implementation ideas I think they are correct. I left several comments there for improvements.

x/emissions/keeper/inference_synthesis/network_inference_builder.go
zale144: old version grouped results per forecaster, so we needed the `len(...) > 0` check to avoid appending empty containers.

in the new version we emit one result per (forecaster, withheldInferer) only when we actually have a valid computed value, so there’s no way to append an empty entry anymore - the check isn’t needed.

x/emissions/keeper/inference_synthesis/network_inference_builder.go
zale144: some places it's straightforward, other places not so much. I started refactoring this and ended up in a deep rabbit hole. so I'd recommend not touching this for now

x/emissions/keeper/inference_synthesis/network_inference_builder.go
zale144: as commented above, this is a rabbit hole and not fit for this current effort. I'll remove the added Ctx and K fields but I don't want to refactor every other place where it's passed like that, I already tried doing that and ended up in a big mess... way too much for this PR

x/emissions/keeper/inference_synthesis/network_inference_builder.go
zale144: no functional reason - just refactored for readability/consistency. the order doesn’t affect behavior, since they’re independent maps and both get passed explicitly.

x/emissions/keeper/actor_utils/worker_test.go
zale144: addressed

x/emissions/keeper/inference_synthesis/forecast_implied.go
zale144: addressed

x/emissions/keeper/inference_synthesis/forecast_implied.go
zale144: addressed

x/emissions/keeper/inference_synthesis/network_inference_builder.go
zale144: addressed

x/emissions/keeper/inference_synthesis/network_inference_builder.go
zale144: addressed

x/emissions/keeper/inference_synthesis/network_inference_builder.go
zale144: addressed

x/emissions/keeper/inference_synthesis/network_inference_builder.go
zale144: addressed

x/emissions/keeper/inference_synthesis/network_inference_builder.go
zale144: addressed

x/emissions/keeper/inference_synthesis/network_inference_builder.go
zale144: addressed

zale144:
implemented and merged in https://github.com/allora-network/allora-chain/pull/942

PR #932, opened 2026-03-12
classification #3 - epoch label registry growth limits

xmariachi:
### Direction 

* maintain labels for current active inferers only during the worker window; when an inferer is expelled, labels no longer referenced by any active inferer are removed from the active label set. This keeps a bounded set of labels (otherwise we could be attacked by brute force of sending random labels)

###

* Decision: do not use first-seen epoch-wide registry growth; derive the final epoch label registry from the final active inferer set at CloseWorkerNonce.
** Why: first-seen is ordering-sensitive and vulnerable to timing/manipulation pressure; final-active-set derivation aligns labels with the actual contributors used for synthesis.
* Bounds: No MaxLabelsPerEpoch; retain MaxLabelsPerSubmission as explicit anti-abuse cap. Effective practical bound comes from active inferer cap and per-submission cap.
* Determinism: build final registry at close with explicit deterministic ordering policy (alternatives below).
* Behavior change: labels from inferers expelled before close are no longer part of final registry if no remaining active inferer references them.

Note: we may want to change this to per-topic settings, since different topics may have different label needs. There is a scalability concern + use case suitability.

## Still Left Todo

* Internal review going on right now, set as `Draft` in the meantime.

---
## Summary by cubic
Implements an epoch-scoped label registry frozen at close and materialized from each worker’s raw inference, with multi-label support, per-submission caps, and optional per-topic whitelists. Adds topic-level limits and module caps to meet ENGN-5152 while keeping growth bounded.

- **New Features**
  - Epoch label registry in `x/emissions`: built from the final active inferer set at `CloseWorkerNonce`, deterministic, immutable after freeze, pruned per-epoch. Satisfies ENGN-5152 acceptance.
  - Topic-level controls (`emissions.v10`): `max_labels_per_submission` and `label_whitelist`. SINGLE topics force `max_labels_per_submission = 1`; MULTI defaults to the module cap if unset; whitelist size capped and not editable during the worker window.
  - Module params (`emissions.v9`): `MaxLabelsPerSubmission` (cap for topic-level limit) and `MaxWhitelistLabelSize`; both updatable via `UpdateParams`.
  - Raw inference persistence: require and store the latest raw `Inference` in `InferenceForecastsBundle`; registry materialized at close; submissions exceeding caps or emptied by whitelist are rejected.
  - Benchmarks: opt-in multi-label network inference benchmarks and README notes; integration tests stabilized with transient RPC retries.

- **Migration**
  - Workers must include the raw `Inference` in `InferenceForecastsBundle`; reputers continue using `LossBundle`.
  - When creating/updating topics, set `max_labels_per_submission` and optionally `label_whitelist`; SINGLE topics always accept exactly one value.
  - Configure module caps via `UpdateParams` (`MaxLabelsPerSubmission`, `MaxWhitelistLabelSize`); `emissions.v10` query/tx/events remain the integration surface.
  - Behavior: the registry is empty until close; labels from inferers expelled before close are excluded unless still referenced. Submissions filtered to zero labels by a whitelist will fail.

<sup>Written for commit c03aef994261a49041d7d00d7f55aacfc6e5bc88. Summary will update on new commits.</sup>


xmariachi:
Abandoned, working on #947 with a simplified version.

PR #947, opened 2026-04-21
labelreg limits + consolidation + label whitelist

xmariachi:
# Label Registry Limits, Canonicalization, and Close-Time Consolidation

## Summary
This PR implements the epoch label registry flow for classification topics. It adds per-topic label limits and allowlists, canonicalizes worker-submitted labels before they reach registry state, and freezes a compact final epoch label registry from the active worker set at `CloseWorkerNonce`.

The main goal is to keep classification label growth bounded and deterministic while preserving flexible topic-defined label schemas.

## What Changed
- Extends `Topic`, `CreateNewTopicRequest`, and `UpdateTopicRequest` with:
  - `max_labels_per_submission`
  - `label_whitelist`
  - `label_default_value`
- Adds canonical label handling for worker inputs and topic whitelists:
  - valid UTF-8 only
  - NFC normalization
  - leading/trailing whitespace trimming
  - control/format rune rejection
  - 64-byte canonical label limit
  - duplicate detection after canonicalization
- Enforces per-topic label submission constraints in the worker payload path:
  - rejects too many labels in one submission
  - rejects labels outside a non-empty topic whitelist
  - rewrites accepted labels into canonical form before normalization/registry use
- Reworks epoch label registry construction for classification inferences:
  - worker submissions build a temporary first-seen registry during the worker submission window
  - only admitted active inferers are persisted in the live temporary inference store
  - `CloseWorkerNonce` materializes the final registry from active non-default labels
  - final label IDs are compacted to `1..N`
  - active inference vectors are remapped to the final compact registry
  - single-output topics use the canonical `y` label
- Adds `label_default_value` semantics for dense multi-label inference vectors:
  - missing labels are filled with the topic default
  - submitted values equal to the default are treated as absent for final registry inclusion
- Blocks label-sensitive topic updates while any unfulfilled worker nonce is inside its worker submission window:
  - `max_labels_per_submission`
  - `label_whitelist`
  - `label_default_value`
  - `merit_sortition_alpha`
- Emits `EventEpochLabelRegistryFrozen` after close-time registry materialization so off-chain consumers can track when an epoch registry is finalized.
- Removes the public `GetWorkerLatestInferenceByTopicId` query and AutoCLI command, since latest per-worker inference state is now temporary internal worker-submission-window state.

## Migration and Compatibility
- The v15 migration backfills `MaxLabelsPerSubmission` for existing topics with `DefaultMaxLabelsPerSubmission`.
[...]

PR #948, opened 2026-04-22
Diego/engn 7634 label normalization and limits

xmariachi:

Epoch Label Registry: label canonicalization, normalization and limits.

### Summary

This change layers **strict label canonicalization and byte limits** on top of the Epoch Label Registry work in the base branch. Labels are normalized for determinism and safety, then **restricted to an explicit ASCII character set** with a configurable per-label byte cap.

### What changed

- **Canonical form** — `CanonicalLabelName` / `CanonicalizeLabelList` apply **NFC**, **trim**, optional **lowercasing** (when the topic is not case-sensitive), **deduplication** after canonicalization, and an **ASCII-only** allowed charset (`a`–`z` / `A`–`Z` when `label_case_sensitive` is true, `0`–`9`, `_`, `-`, space, `/`, `.`). Invalid UTF-8 and any other rune are rejected with `ErrInvalidLabelName`.
- **New module param** — `max_canonical_label_byte_length` (default **64**; validated in `Params`), used whenever whitelist and submission-time canonicalization run.
- **New topic field** — `label_case_sensitive` on `Topic` and `CreateNewTopicRequest` (immutable after topic creation; wired through validation and the worker payload path).
- **Persistence** — `SetTopic` / `UpdateTopic` (and the worker path) use the new cap and the topic’s case-sensitivity so stored whitelists and in-flight labels stay consistent.
- **v15 migration** — **re-canonicalizes** existing per-topic `label_whitelist` under the new rules, drops entries that no longer canonicalize, collapses post-canonical duplicates, and rewrites topics when the stored list changes (idempotent re-runs when the canonical list is already stable).
- **CLI** — Autocli/flags extended for the new `create-topic` / `update-topic` arguments and related params.
- **Tests** — Unit and fuzz tests for the canonicalizer; keeper/msgserver/migration tests updated. **Integration** `label_registry_classification_test.go` exercises **happy-path** E2E only with **valid ASCII** labels; negative cases remain in unit tests.
- **Tooling** — `integration-test` Makefile target where applicable.

### Design note

Unicode letters and symbols are **intentionally not** allowed: canonical labels are bounded ASCII strings suitable for store keys, sorting, and byte equality, in line with “normalization and limits for chain safety” on the PR.

---
## Summary by cubic
Adds label normalization and limits across `x/emissions` to standardize labels and improve chain safety, implementing ENGN-7634. Enforces ASCII-only canonical labels with a configurable byte cap and a topic-level case-sensitivity flag across whitelists, submissions, epoch registries, and migrations.

- New Features
  - Canonicalization: trim; ASCII letters a–z/A–Z, digits 0–9, separators `_`, `-`, space, `/`, `.`; lowercased when `label_case_sensitive=false` (default); rejects invalid UTF-8 and non-ASCII. Applied to topic whitelists, worker payloads, and the epoch label registry using the topic’s case flag. Validation runs charset/UTF-8 checks before case-fold and length checks. SINGLE-arity topics hard-enforce a 1-label cap at submission and registry materialization.
  - Config/API: new `max_canonical_label_byte_length` param (default 64; bounded) used in whitelist and submission validation. New immutable `Topic.label_case_sensitive`. `SetTopic` canonicalizes whitelists with the cap and flag. Worker path validates with the topic’s cap/whitelist and param byte limit, and `RegisterEpochLabel` accepts the topic’s case flag. `UpdateTopic` bounds-checks `max_labels_per_submission` and rejects immutable-field changes with `ErrInvalidTopicUpdate`. Protos and `autocli` updated (`create-topic` adds `[label_case_sensitive]`). Materialization and genesis paths thread the same byte cap.

- Migration
  - v15 backfills `Params.MaxCanonicalLabelByteLength`, then canonicalizes stored `label_whitelist` under the ASCII rules using each topic’s `label_case_sensitive` (false for legacy), dropping invalid and post-canonical duplicates; idempotent.

<sup>Written for commit eec29fd866a09779a54f05fe4902d9d29e1fec53. Summary will update on new commits. <a href="https://cubic.dev/pr/allora-network/allora-chain/pull/948?utm_source=github">Review in cubic</a></sup>


x/emissions/proto/emissions/v10/tx.proto
```diff
@@ -341,6 +341,11 @@ message CreateNewTopicRequest {
   uint64 max_labels_per_submission = 26;
   // Per-topic canonical label allowlist. Empty means unrestricted.
   repeated string label_whitelist = 27;
+  // When false (the default), label names are lowercased during canonicalization
+  // so that "Cat", "CAT" and "cat" all collapse to "cat". When true, casing is
+  // preserved and labels that differ only in case are treated as distinct.
+  // The field is immutable after topic creation.
+  bool label_case_sensitive = 28;
```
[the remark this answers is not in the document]
xmariachi: The definition on the ticket has been rewritten, it's aligned now.

x/emissions/types/label_canonicalize_test.go
```diff
@@ -21,35 +29,67 @@ func TestCanonicalLabelName_Accepts(t *testing.T) {
...
+			}
+			if again != got {
+				t.Fatalf("CanonicalLabelName not idempotent: %q -> %q", got, again)
+			}
+		})
+	}
+}
+
+// TestCanonicalLabelName_CaseSensitive verifies that
+// labelCaseSensitive=true preserves uppercase ASCII letters and that the
```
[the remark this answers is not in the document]
xmariachi: The ticket semantics are now clarified/aligned: `label_case_sensitive=false` lowercases labels, while `true` preserves ASCII case and treats case variants as distinct. The latest commit also updates test fixtures that need case preservation to set `LabelCaseSensitive: true`, so the canonicalizer test expectations are consistent with the current requirement.

x/emissions/types/params.go
```diff
@@ -245,6 +246,29 @@ func (p Params) Validate() error {
...
+func validateMaxCanonicalLabelByteLength(i uint64) error {
+	if i < MinMaxCanonicalLabelByteLength {
+		return errorsmod.Wrapf(ErrValidationMustBeGreaterthanZero,
+			"max canonical label byte length must be >= %d, got %d",
+			MinMaxCanonicalLabelByteLength, i)
+	}
+	if i > MaxMaxCanonicalLabelByteLength {
+		return errorsmod.Wrapf(ErrInvalidValue,
+			"max canonical label byte length must be <= %d, got %d",
+			MaxMaxCanonicalLabelByteLength, i)
```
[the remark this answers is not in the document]
xmariachi: incorrect, 64 is a changeable fiducial value. It can be changed later on, between the min and max.

x/emissions/module/autocli.go
```diff
@@ -982,21 +982,14 @@ func (am AppModule) AutoCLIOptions() *autocliv1.ModuleOptions {
...
-						"max_labels_per_submission": {
-							Name:  "max-labels-per-submission",
-							Usage: "Per-topic cap on the number of canonical labels per worker submission. Zero means use params.MaxLabelsPerSubmission.",
-						},
-						"label_whitelist": {
-							Name:  "label-whitelist",
-							Usage: "Per-topic canonical label allowlist. Empty means unrestricted. Stored canonicalized.",
-						},
+						{ProtoField: "max_labels_per_submission"},
+						{ProtoField: "label_whitelist"},
```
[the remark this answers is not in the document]
xmariachi: All params are needed, included the `label_whitelist`. 
Changed in docs.

x/emissions/keeper/topic.go
```diff
@@ -342,20 +346,52 @@ func labelWhitelistChanged(a, b []string) bool {
...
+	// the msgserver does not have to know about this invariant on its own.
+	if topic.LabelCaseSensitive != updatedTopic.LabelCaseSensitive {
+		return types.Topic{}, errorsmod.Wrap(
+			types.ErrInvalidTopicUpdate,
+			"label_case_sensitive is immutable after topic creation",
+		)
+	}
+	if updatedTopic.MaxLabelsPerSubmission < types.MinMaxLabelsPerSubmission {
+		return types.Topic{}, errorsmod.Wrapf(
+			types.ErrValidationMustBeGreaterthanZero,
```
zale144, line 366: should probably use `types.ErrInvalidValue` instead

x/emissions/keeper/topic.go
```diff
@@ -342,20 +346,52 @@ func labelWhitelistChanged(a, b []string) bool {
...
+	}
+	if updatedTopic.MaxLabelsPerSubmission > types.MaxMaxLabelsPerSubmission {
+		return types.Topic{}, errorsmod.Wrapf(
+			types.ErrInvalidValue,
+			"max_labels_per_submission must be <= %d, got %d",
+			types.MaxMaxLabelsPerSubmission,
+			updatedTopic.MaxLabelsPerSubmission,
+		)
+	}
+
```
zale144, line 380: these should probably go to the #947 PR.
also, they should probably be validated in `func (topic Topic) Validate`

x/emissions/proto/emissions/v10/topic.proto
```diff
@@ -73,6 +73,13 @@ message Topic {
   // topic is unrestricted; when non-empty, submitted labels must match (after
   // canonicalization) an entry in this list. Stored in canonical form.
   repeated string label_whitelist = 28;
+  // When false (the default), label names are lowercased during canonicalization
+  // so that "Cat", "CAT" and "cat" all collapse to "cat". When true, casing is
+  // preserved and labels that differ only in case are treated as distinct.
+  // This field is immutable after topic creation because flipping it would
+  // change the canonical form of every already-persisted label (whitelist
+  // entries, active refcounts, epoch registry names).
+  bool label_case_sensitive = 29;
```
zale144: I just don't see a lot of benefit with adding this. can we not just agree that labels should be case-sensitive and call it a day?

x/emissions/proto/emissions/v9/params.proto
```diff
@@ -185,4 +185,11 @@ message Params {
   // topic sets its own max_labels_per_submission, the effective cap is the
   // minimum of the two.
   uint64 max_labels_per_submission = 61;
+  // Max byte length of a canonical label name after NFC normalization and
+  // trimming. Enforced by CanonicalLabelName at both submission time and
+  // topic whitelist canonicalization. Because the canonical form is ASCII
+  // only, this is equivalent to character count. Stored as bytes so every
+  // activeInfererLabelRefCount row key and EpochLabelRegistry label entry
+  // has a deterministic upper bound on serialized size.
+  uint64 max_canonical_label_byte_length = 62;
```
zale144: should this change not belong to #947 PR?

x/emissions/types/label_canonicalize.go
```diff
@@ -2,40 +2,79 @@ package types
...
+// modest. The effective bound used at runtime is the
+// Params.MaxCanonicalLabelByteLength value, which the keeper loads and
+// threads through CanonicalLabelName/CanonicalizeLabelList. Because the
+// canonical form is restricted to ASCII, the byte cap is equivalent to
+// character count. It is still enforced as bytes so every row key in the
+// activeInfererLabelRefCount store and every EpochLabelRegistry entry has
+// a deterministic, modest upper bound on serialized size.
+const (
+	MinMaxCanonicalLabelByteLength = uint64(1)
+	MaxMaxCanonicalLabelByteLength = uint64(1024)
```
zale144: maybe consider moving these where 
`MinMaxLabelsPerSubmission`
`MaxMaxLabelsPerSubmission`
are, in params.go

x/emissions/types/label_canonicalize.go
```diff
@@ -2,40 +2,79 @@ package types
...
-//  2. Trimmed of leading and trailing Unicode whitespace (strings.TrimSpace).
-//  3. Checked to be non-empty after trimming.
-//  4. Checked to contain no control (Cc) or format (Cf) runes — these are
-//     invisible characters (e.g. zero-width space, bidi marks, NULs) that
-//     would otherwise allow two visually-identical labels to differ by a
-//     stealthy invisible rune and thus break the lex-sort used to assign
-//     registry ids deterministically.
-//  5. Bounded at MaxCanonicalLabelByteLength bytes after normalization.
-//  6. Verified to be valid UTF-8.
+//  1. Valid UTF-8 (rejected otherwise).
```
zale144: I wish we had a tl;dr section in these docstrings

zale144 (commented):
left some comments

x/emissions/types/label_canonicalize.go
xmariachi: The idea for these min/max is to set stricter guardrails for `maxCanonicalLabelByteLength`. - if you can modify them in the same way, then they defeat their own purpose. It's an extra protection basically.

x/emissions/proto/emissions/v10/topic.proto
xmariachi: I think there's a case for some being case-insensitive, for better recognition of same label expressed differently. eg. sending `BTC` or `btc` would mean two different labels. It would depend on the topic's context.

x/emissions/types/label_canonicalize.go
zale144: no, I mean, I just thought they could be moved to a different place in the codebase, so they're all in the same place. no biggie though

x/emissions/keeper/topic.go
xmariachi: [done in 947](https://github.com/allora-network/allora-chain/pull/947/commits/dc40c9af63e3bbb4be3c4192cd8e67bcfdbfe245)

x/emissions/keeper/topic.go
xmariachi: This is an error used in other places too. Gives a bit more detail. I think I'll keep current one.

x/emissions/proto/emissions/v9/params.proto
xmariachi: Actually yes, though Being about limits, I'm keeping it here.

xmariachi (commented):
Post-rebase review comments from `REVIEW_AFTER_REBASE_P0P1.md` and `REVIEW_AFTER_REBASE_P2.md`.

x/emissions/migrations/v15/migrate.go
```diff
@@ -44,6 +61,171 @@ func MigrateStore(ctx sdk.Context, emissionsKeeper keeper.Keeper) error {
...
+				droppedForTopic++
+				continue
+			}
+			seen[canonical] = struct{}{}
+			filtered = append(filtered, canonical)
+		}
+
+		labelsDropped += droppedForTopic
+		if !labelWhitelistEqual(topic.LabelWhitelist, filtered) {
+			topic.LabelWhitelist = filtered
```
xmariachi: **P1: v15 whitelist migration can turn a restricted whitelist into an unrestricted topic**

**Validity:** Valid. `MigrateTopicLabelWhitelists` drops labels that fail the new canonicalizer and persists the filtered slice whenever it differs from the stored whitelist. If every original entry is invalid under the new ASCII/byte rules, the topic is stored with `len(LabelWhitelist) == 0`. Submission validation treats an empty whitelist as unrestricted, and the proto docs state "Empty means unrestricted."

**Related to current PR intention? Why:** Yes. This PR explicitly adds stricter label canonicalization and a v15 migration for stored topic whitelists. A migration that widens a non-empty allowlist into no allowlist is directly related to the correctness and safety of that migration.

**Risks fixing:** Failing migration on all-invalid whitelists can block upgrade until operators/governance remediate those topics. Preserving invalid labels avoids widening but leaves non-canonical state that this PR is trying to eliminate. Introducing a "closed whitelist" sentinel would require broader semantics and tests.

**Risks not fixing:** A topic creator/admin that configured a restrictive whitelist can unintentionally end up with an unrestricted topic after upgrade if all entries are filtered out. Workers could submit arbitrary valid canonical labels to that topic, which is the opposite of the PR's limit/allowlist intent.

**Worth doing right now?** Yes. This should be resolved before shipping the migration, because the old-to-new state transition is hard to repair after upgrade.

**Proposed fix:** In `MigrateTopicLabelWhitelists`, detect `len(original) > 0 && len(filtered) == 0`. Prefer returning a migration error with the topic id and drop count so the chain does not silently widen policy. If halting is unacceptable, define and enforce an explicit "closed until reconfigured" behavior instead of reusing empty-as-unrestricted.

x/emissions/types/label_canonicalize.go
```diff
@@ -52,17 +53,28 @@ func CanonicalLabelName(s string) (string, error) {
...
 		return "", errorsmod.Wrap(ErrInvalidLabelName, "label is empty after trimming")
 	}
+	if !labelCaseSensitive {
+		// Using strings.ToLower on ASCII-only input produces ASCII-only
+		// output; non-ASCII runes would be rejected by the charset check
+		// that follows regardless. Lowercasing before the charset check
+		// keeps the rejection message accurate even when the input had
+		// uppercase letters that would have been legal under
+		// label_case_sensitive.
+		trimmed = strings.ToLower(trimmed)
```
xmariachi: **P1: Case-insensitive canonicalization lowercases Unicode before enforcing ASCII-only input**

**Validity:** Valid. `CanonicalLabelName` validates UTF-8, NFC-normalizes, trims, then calls `strings.ToLower` when `labelCaseSensitive == false`; only after that does it enforce the allowed ASCII charset. Go Unicode lowercasing can map some non-ASCII code points to ASCII letters, so an input that is not ASCII can become acceptable after lowercasing. This contradicts the PR's stated ASCII-only input policy and the local comment's assumption that non-ASCII runes will still be rejected by the following charset check.

**Related to current PR intention? Why:** Yes. The central intention of PR 948 is deterministic label canonicalization with an explicit ASCII character set and byte limits. This issue is in the canonicalization function used by whitelists, worker payload validation, and label registration.

**Risks fixing:** Some currently accepted edge-case Unicode inputs will be rejected. That is consistent with the documented ASCII-only behavior but could require updating tests if any intentionally rely on Unicode case folding.

**Risks not fixing:** Non-ASCII labels can be accepted in case-insensitive topics after being transformed to ASCII, which creates surprising whitelist/registry behavior and weakens the "Unicode letters and symbols are intentionally not allowed" guarantee. It can also make reviews and client-side validation disagree with chain behavior.

**Worth doing right now?** Yes. The fix is localized and should be made before clients or operators depend on the final canonicalization contract.

**Proposed fix:** Reject disallowed/non-ASCII runes before any Unicode-wide lowercasing, or replace `strings.ToLower` with an ASCII-only fold that only maps `A` through `Z` to `a` through `z` after validating that every rune is in the allowed ASCII set. Add regression cases for Unicode code points that lower/fold to ASCII.

x/emissions/keeper/label_registry.go
```diff
@@ -312,7 +317,11 @@ func validateEpochLabelRegistry(
 		if lbl.Id != expectedID {
 			return errorsmod.Wrapf(sdkerrors.ErrLogic, "registry label %q has id %d expected %d", lbl.Name, lbl.Id, expectedID)
 		}
-		canonicalLabel, err := types.CanonicalLabelName(lbl.Name)
+		canonicalLabel, err := types.CanonicalLabelName(
+			lbl.Name,
+			types.MaxMaxCanonicalLabelByteLength,
```
xmariachi: **P2: Epoch label registry validation uses the hard ceiling instead of the live label byte param**

**Validity:** Valid. Worker submission and `RegisterEpochLabel` canonicalize labels with `Params.MaxCanonicalLabelByteLength`, but `validateEpochLabelRegistry` uses `types.MaxMaxCanonicalLabelByteLength` instead. That means bulk registry validation, including `SetEpochLabelRegistry`, can accept labels that would not be accepted through the normal runtime registration path under the active module params.

**Related to current PR intention? Why:** Yes. The PR introduces a configurable canonical label byte cap and threads it through whitelist and submission-time canonicalization. Stored registry validation is part of the same canonical label invariant, especially after the rebase restored registry registration on the base branch's model.

**Risks fixing:** The helper currently has a pure signature. Fixing requires passing `maxLabelBytes` or params through `SetEpochLabelRegistry`, `validateTemporaryRegistry`, and callers. Strict validation may reject handcrafted genesis/test state that previously fit under the hard 1024-byte ceiling.

**Risks not fixing:** Genesis/import or internal keeper writes can persist registry labels longer than the active cap. If governance lowers the cap, stored registries can continue to validate under a looser rule than submissions and whitelists, which makes the label-size policy inconsistent.

**Worth doing right now?** Yes, if the PR's byte cap is meant to be a single live policy for canonical labels. If historical registries are intentionally allowed to exceed later param reductions, document that exception explicitly.

**Proposed fix:** Load params in the keeper paths that validate registries and pass `params.MaxCanonicalLabelByteLength` into registry validation instead of `types.MaxMaxCanonicalLabelByteLength`. Keep the hard max only as the governance/genesis guardrail for the param itself.

x/emissions/migrations/v15/migrate.go
```diff
@@ -44,6 +61,171 @@ func MigrateStore(ctx sdk.Context, emissionsKeeper keeper.Keeper) error {
...
+//     every label.
+func MigrateParams(ctx sdk.Context, emissionsKeeper keeper.Keeper) error {
+	params, err := emissionsKeeper.GetParams(ctx)
+	if err != nil {
+		return errorsmod.Wrap(err, "MIGRATION V15: failed to get existing params")
+	}
+
+	changed := false
+	if params.MaxCanonicalLabelByteLength == 0 {
+		params.MaxCanonicalLabelByteLength = 64
```
xmariachi: **P2: v15 migration hardcodes the default byte cap instead of sharing the params default**

**Validity:** Valid. `MigrateParams` backfills `MaxCanonicalLabelByteLength` to literal `64`, and `MigrateTopicLabelWhitelists` has a defensive fallback to the same literal. The module default is also `64`, but the migration is duplicating that value rather than deriving it from the shared params default.

**Related to current PR intention? Why:** Yes. This PR adds `MaxCanonicalLabelByteLength` and a migration that must backfill it before whitelist canonicalization. Keeping the migration's default in sync with `DefaultParams` is part of robust param/migration alignment.

**Risks fixing:** Very low. The main care point is avoiding unintended allocation or validation side effects from calling `DefaultParams()` in migration code.

**Risks not fixing:** If the default changes later, migration behavior can silently drift from module defaults and tests may continue to pass because both literals are not tied together. The fallback path would also encode a stale "module-initial cap" in comments and behavior.

**Worth doing right now?** Yes. It is a small consistency fix in migration code while the param is still being introduced.

**Proposed fix:** Replace both `64` literals with `emissionstypes.DefaultParams().MaxCanonicalLabelByteLength`, or introduce a named default constant used by both `DefaultParams()` and v15 migration backfill/fallback.

x/emissions/migrations/v15/migrate.go
xmariachi: I think this one is N/A for the actual upgrade path.

This migration runs over pre-v15 topic state where `LabelWhitelist` is not populated. When old topic bytes decode into the new `Topic` shape, `LabelWhitelist` is the zero value (`nil` / empty). `MigrateTopics` does not set it either; it only backfills `TopicType`, `OutputArity`, `UnityTolerance`, and `MaxLabelsPerSubmission`.

Then `MigrateTopicLabelWhitelists` skips empty whitelists:

```go
if len(topic.LabelWhitelist) == 0 {
    continue
}

So the problematic path, “non-empty restricted whitelist canonicalizes to empty and becomes unrestricted,” cannot occur for existing chain topics, since they did not have such field prior to this migration.

File: x/emissions/types/label_canonicalize.go (hunk 12 above)
Comment (xmariachi) [reply x/emissions/types/label_canonicalize.go]: seems valid, will tackle.

Comment (xmariachi) [reply x/emissions/types/label_canonicalize.go]: Fixed in [035b2025](https://github.com/allora-network/allora-chain/pull/948/commits/035b2025d7acf24b92874a4c165bbd573dc952bf)

File: x/emissions/migrations/v15/migrate.go (hunk 14 above)
Comment (xmariachi) [reply x/emissions/migrations/v15/migrate.go]: fixed in [this commit](https://github.com/allora-network/allora-chain/pull/948/commits/c49830d03c3b8e64cc12ec7483bf07757872f86e)

File: x/emissions/keeper/label_registry.go (hunk 13 above)
Comment (xmariachi) [reply x/emissions/keeper/label_registry.go]: fixed in [this commit](https://github.com/allora-network/allora-chain/pull/948/commits/c49830d03c3b8e64cc12ec7483bf07757872f86e)

File: x/emissions/keeper/label_registry.go [hunk 15]
```diff
@@ -120,7 +122,15 @@ func (k *TopicKeeper) SetEpochLabelRegistry(
...
-	if err := validateEpochLabelRegistry(registry.TopicId, nonce, registry); err != nil {
+	topic, err := k.GetTopic(ctx, registry.TopicId)
+	if err != nil {
+		return errorsmod.Wrap(err, "failed to get topic for epoch label registry validation")
+	}
+	params, err := k.paramsKeeper.GetParams(ctx)
+	if err != nil {
+		return errorsmod.Wrap(err, "failed to get params for epoch label registry validation")
+	}
+	if err := validateTemporaryRegistry(topic, nonce, registry, params.MaxCanonicalLabelByteLength); err != nil {
```
zale144: no need to pass the whole topic, just `LabelCaseSensitive` would be enough

x/emissions/keeper/topic.go
```diff
@@ -862,7 +901,19 @@ func (k *TopicKeeper) RegisterEpochLabel(
 	nonce types.BlockHeight,
 	labelName string,
 ) (LabelId, error) {
-	canonicalLabel, err := types.CanonicalLabelName(labelName)
+	topic, err := k.GetTopic(ctx, topicId)
```
zale144: the two places where `RegisterEpochLabel` is called already have the topic

x/emissions/keeper/worker.go
```diff
@@ -775,7 +775,16 @@ func (k *WorkerKeeper) GetWorkerLatestInputInferenceByTopicId(
...
-	return MaterializeInputInferenceFromTemporaryRegistry(topic, registry, inference)
+	params, err := k.paramsKeeper.GetParams(ctx)
+	if err != nil {
+		return nil, errorsmod.Wrap(err, "error getting params for input inference materialization")
+	}
+	return MaterializeInputInferenceFromTemporaryRegistry(
+		topic,
+		registry,
+		inference,
+		params.MaxCanonicalLabelByteLength,
```
zale144, line 786: I wish this was hardcoded so we don't have to add so many `GetParams` calls. but the question is at what value should it be

x/emissions/migrations/v15/migrate.go
```diff
@@ -44,6 +61,171 @@ func MigrateStore(ctx sdk.Context, emissionsKeeper keeper.Keeper) error {
...
+//     rather than aborting the migration: on a live chain we prefer to clear
+//     obviously-unusable whitelist entries than to halt the upgrade, because
+//     a whitelisted label that cannot be submitted is already a silent dead
+//     end.
+//   - If the filtered whitelist differs from the stored one, the topic is
+//     persisted.
+//
+// EpochLabelRegistry entries are left untouched: they are nonce-scoped and
+// were already persisted under the previous rules.
+func MigrateTopicLabelWhitelists(
```
zale144: but the topics we will migrate once multi-label is introduced will not have `LabelWhitelist`. or am I missing something?

x/emissions/module/rewards/rewards_multilabel_test.go
```diff
@@ -725,6 +735,7 @@ func (s *RewardsTestSuite) TestMultiLabelMatchesSingleLabelPerLabel() {
 		workerIndexes,
 		reputerIndexes,
 		testutil.WithOutputArity(types.TopicOutputArity_TOPIC_OUTPUT_ARITY_MULTI),
+		testutil.WithLabelCaseSensitive(true),
```
zale144, line 800: no tests here where it's case insensitive?

x/emissions/proto/emissions/v10/topic.proto
```diff
@@ -78,6 +78,13 @@ message Topic {
     (gogoproto.customtype) = "github.com/allora-network/allora-chain/math.Dec",
     (gogoproto.nullable) = false
   ];
+  // When false (the default), label names are lowercased during canonicalization
+  // so that "Cat", "CAT" and "cat" all collapse to "cat". When true, casing is
+  // preserved and labels that differ only in case are treated as distinct.
+  // This field is immutable after topic creation because flipping it would
+  // change the canonical form of every already-persisted label (whitelist
+  // entries and epoch registry names).
+  bool label_case_sensitive = 30;
```
zale144, line 87: again, this addition introduces a lot of new changes to the codebase and more store calls, but I don't quite see the value here. more as a nice to have. for me, I don't think it should be added in phase 1. for the time being, I would make it case sensitive by default.

x/emissions/keeper/worker.go
xmariachi: I think we should favour functionality over performance when penalty is tiny. Making those calls won't hurt too much, but not being able to push that limit if we need to (without an upgrade) could.

x/emissions/proto/emissions/v10/topic.proto
xmariachi: I don't think there's many changes and they are pretty self-contained (thus easy to cover via unit-test) so I think it's not a problem. Am I missing anything special here?

x/emissions/keeper/label_registry.go
xmariachi: Refactored the calls to RegisterEpochLabel [here](https://github.com/allora-network/allora-chain/pull/948/commits/bf8519dcaef44f33540480c4448f8c7e5d2e40bd)

x/emissions/keeper/topic.go
xmariachi: Refactored the calls to RegisterEpochLabel [here](https://github.com/allora-network/allora-chain/pull/948/commits/bf8519dcaef44f33540480c4448f8c7e5d2e40bd)

x/emissions/module/rewards/rewards_multilabel_test.go
xmariachi: good point, added one [here](https://github.com/allora-network/allora-chain/pull/948/commits/bf8519dcaef44f33540480c4448f8c7e5d2e40bd)

x/emissions/migrations/v15/migrate.go
xmariachi: yep - this simplification got lost in the rebase, it was in 947.
Fixed [here](https://github.com/allora-network/allora-chain/pull/948/commits/eec29fd866a09779a54f05fe4902d9d29e1fec53)

PR #942, opened 2026-04-09
classification

zale144:
## Purpose of Changes and their Description

this is the whole feature as merged from classification PRs: #909, #918, #920, #925, #929, #933, and #947 @xmariachi 

note that this version removes the ability to update unityRequired and unityTolerance fields for topics

## Are these changes tested and documented?

- [ ] If tested, please describe how. If not, why tests are not needed.
- [ ] If documented, please describe where. If not, describe why docs are not needed.
- [ ] Added to `Unreleased` section of `CHANGELOG.md`?

## Still Left Todo

*Fill this out if this is a Draft PR so others can help.*

---
## Summary by cubic
Adds multi‑label classification with vectorized inference synthesis and a bounded, canonical label registry via new `x/emissions/api/emissions/v10` APIs. Finalizes the `v0.17.0` upgrade with bundle‑backed network inferences, stricter validation/migration safeguards, and end‑to‑end tests.

- New Features
  - Topic config: `TopicType`, `OutputArity` (SINGLE|MULTI), and label registry settings (`max_labels_per_submission`, `label_whitelist`, `label_default_value`) with limits (`max_canonical_label_byte_length`, `max_topic_label_whitelist_size`, `max_epoch_label_registry_size`).
  - Label registry: canonicalizes labels (UTF‑8, NFC, trim, no control/format, ≤64 bytes), enforces caps/whitelist, compacts and freezes an epoch registry at `CloseWorkerNonce` (emits `EventEpochLabelRegistryFrozen`), guards label updates during the worker window.
  - Inference: MULTI uses labeled `Values` with default padding; SINGLE uses `Value`. Vectorized via `DecMatrix`. Stores `NetworkInferenceBundle`. Only SINGLE updates outlier metrics; outlier‑resistant queries error on MULTI.
  - APIs: new `v10` tx/query/events/gRPC; legacy `v2`–`v9` wired through `v9`. Renamed latest‑inference query to `GetWorkerLatestInputInferenceByTopicId`. AutoCLI `create-topic` arg order updated. Introduced `LossBundles`. Removed unused `emissions.v10.EventNetworkInferences` variants in favor of bundle events.
  - CI/Tests: pin `buf` to `1.67.0`; enable both integration suites with extended timeouts; add multilabel simulator CSV test and `v0.17.0` upgrade rehearsal.

- Bug Fixes
  - Centralized classification validation; enforced canonical single‑label `y`; canonicalized `label_whitelist`.
[...]

PR #950, opened 2026-05-13
DEVOP-586: drop GH_TOKEN --build-arg secret leak

srt0422:
## Summary

Removes `--build-arg "GH_TOKEN=\${{ secrets.GH_READONLY_PAT }}"` from `build_push_docker_hub.yml` and the matching unused `ARG GH_TOKEN` from `Dockerfile`. Closes the Shai-Hulud-amplifier secret leak path documented in DEVOP-586.

## Why removal instead of BuildKit secret mount

The ticket suggested switching to `--mount=type=secret,id=gh_pat` (BuildKit). That's the correct pattern *if* the Dockerfile actually consumes the token. This one doesn't:

- `Dockerfile` declares `ARG GH_TOKEN` but never references it in any `RUN`.
- `make install` resolves only public Go modules (verified by inspecting `go.mod`: no `allora-network/*` or other private deps).

The token has been entirely dead weight — it adds attack surface (leaks into image layer metadata via `docker history --no-trunc`) and provides zero functional value. Removing it is strictly simpler than refactoring to BuildKit and equally secure.

## Verification

- [x] `Dockerfile` has no remaining `GH_TOKEN` / `GH_READONLY_PAT` reference.
- [x] Workflow has no remaining `--build-arg` or env var passing the secret.
- [x] `go.mod` confirmed all-public deps.
- [ ] CI green (reviewer to confirm).
- [ ] After merge: rotate `GH_READONLY_PAT` (previously-built images on Docker Hub may have it baked in via layer metadata). Audit Harbor projects with pull access to `alloranetwork/allora-chain`.

## Related

- Linear: https://linear.app/alloralabs/issue/DEVOP-586
- Companion PR for `explorer` repo: opened in `allora-network/explorer`.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

---
## Summary by cubic
Removed `--build-arg "GH_TOKEN=${{ secrets.GH_READONLY_PAT }}"` from `.github/workflows/build_push_docker_hub.yml` and the unused `ARG GH_TOKEN` from `Dockerfile` to prevent PAT leaks; removed the now-obsolete workflow comment. Added a SECURITY note in `Dockerfile` warning against `--build-arg` secrets and pointing to BuildKit if ever needed; all modules are public, so this completes DEVOP-586.

- **Migration**
  - Rotate `GH_READONLY_PAT` (older images may include it in layer history).
  - Audit Harbor projects with pull access to `alloranetwork/allora-chain` images.

<sup>Written for commit 16f4977a39aea7772c4c019e3f8d34f674228383. Summary will update on new commits. <a href="https://cubic.dev/pr/allora-network/allora-chain/pull/950?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.github/workflows/build_push_docker_hub.yml
+++ b/.github/workflows/build_push_docker_hub.yml
@@ -46,12 +46,6 @@ jobs:
           fi
 
           # Build a docker container and push it to Docker Hub.
-          # GH_TOKEN was previously passed as a --build-arg, which embeds the
-          # value in image layer metadata (visible via `docker history`/`docker
-          # inspect`) and would persist in the OCI manifest forever. The
-          # Dockerfile declared `ARG GH_TOKEN` but never consumed it (the build
-          # uses public Go modules only), so removing both is safe and closes
-          # the leak. See DEVOP-586.
           docker build --pull -t $DOCKERHUB_USERNAME/$DOCKERHUB_REPOSITORY:$IMAGE_TAG .
           docker push $DOCKERHUB_USERNAME/$DOCKERHUB_REPOSITORY:$IMAGE_TAG
 
--- a/.github/workflows/build_push_docker_hub.yml
+++ b/.github/workflows/build_push_docker_hub.yml
@@ -46,7 +46,13 @@ jobs:
           fi
 
           # Build a docker container and push it to Docker Hub.
-          docker build --pull --build-arg "GH_TOKEN=${{ secrets.GH_READONLY_PAT }}" -t $DOCKERHUB_USERNAME/$DOCKERHUB_REPOSITORY:$IMAGE_TAG .
+          # GH_TOKEN was previously passed as a --build-arg, which embeds the
+          # value in image layer metadata (visible via `docker history`/`docker
+          # inspect`) and would persist in the OCI manifest forever. The
+          # Dockerfile declared `ARG GH_TOKEN` but never consumed it (the build
+          # uses public Go modules only), so removing both is safe and closes
+          # the leak. See DEVOP-586.
+          docker build --pull -t $DOCKERHUB_USERNAME/$DOCKERHUB_REPOSITORY:$IMAGE_TAG .
           docker push $DOCKERHUB_USERNAME/$DOCKERHUB_REPOSITORY:$IMAGE_TAG
 
           # Build and PUSH additional tags
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,5 @@
 FROM golang:1.22-bookworm AS builder
 
-ARG GH_TOKEN
-
 ADD . /src
 WORKDIR /src
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,10 @@
 FROM golang:1.22-bookworm AS builder
 
+# SECURITY: Do not pass credentials via `--build-arg` / `ARG` -- they are
+# embedded in image metadata (`docker history`, OCI provenance) and leak to
+# anyone with pull access. If a private Go module is ever needed, use a
+# BuildKit secret (`RUN --mount=type=secret,id=...`). See DEVOP-586.
+
 ADD . /src
 WORKDIR /src
 


.github/workflows/build_push_docker_hub.yml
```diff
@@ -46,7 +46,13 @@ jobs:
           fi
 
           # Build a docker container and push it to Docker Hub.
-          docker build --pull --build-arg "GH_TOKEN=${{ secrets.GH_READONLY_PAT }}" -t $DOCKERHUB_USERNAME/$DOCKERHUB_REPOSITORY:$IMAGE_TAG .
+          # GH_TOKEN was previously passed as a --build-arg, which embeds the
```
xmariachi: This comment is too verbose. The source of issues has been removed, so this comment lies there explaining something is no longer there. I think the comment can be removed

xmariachi (changes requested):
lgtm, but the verbose comment is not needed

.github/workflows/build_push_docker_hub.yml
srt0422: > This comment is too verbose. The source of issues has been removed, so this comment lies there explaining something is no longer there. I think the comment can be removed

Addressed in `7ba09566` — removed the 6-line comment block from `.github/workflows/build_push_docker_hub.yml`. Only the original `# Build a docker container and push it to Docker Hub.` line remains.

srt0422:
@xmariachi:

> lgtm, but the verbose comment is not needed

Addressed in `7ba09566` — removed the verbose `GH_TOKEN`-removal explanation block from `build_push_docker_hub.yml`. The inline review thread has also been resolved.

TaniBuilds (dismissed):

.github/workflows/build_push_docker_hub.yml
srt0422: Re-confirming after force-push (signed commits): the verbose comment block is removed in `e440f7d0` (rebased equivalent of the earlier `7ba09566`). Only the original `# Build a docker container and push it to Docker Hub.` line remains.

srt0422:
## Status update after re-sign push

- ✅ Both commits re-signed (SSH/ED25519, `srt0422`) — `signature.state = VALID` on every commit (GitHub "Verified" badge).
- ✅ Code is unchanged from the previously-approved tree: `Dockerfile` has no `GH_TOKEN` reference, workflow has no `--build-arg` / verbose explanatory comment. Rebased `7ba09566` → `e440f7d0` is content-equivalent (the diff against `dev` is identical).
- ⚠️ **CI / GitHub Actions appears paused org-wide.** The most recent GitHub Actions workflow run on this repo (any branch, any actor) is `2026-05-18T11:57:44Z`. The force-push today did not create any new workflow runs — same is true for PRs `#951`, `#947`, `#942`. Only check-app integrations (GitGuardian) are reaching the PR; GitHub Actions itself is silent. This needs an org-admin to check Actions billing / org-level Actions settings; the previous run on `4bd9b8e1` (the original head) was a clean pass for `golangci-lint` + `format_and_test`.

Re-requesting review from @xmariachi since force-push dismissed the prior approval / changes-requested review pair.

.github/workflows/build_push_docker_hub.yml
```diff
@@ -46,7 +46,7 @@ jobs:
           fi
 
           # Build a docker container and push it to Docker Hub.
-          docker build --pull --build-arg "GH_TOKEN=${{ secrets.GH_READONLY_PAT }}" -t $DOCKERHUB_USERNAME/$DOCKERHUB_REPOSITORY:$IMAGE_TAG .
+          docker build --pull -t $DOCKERHUB_USERNAME/$DOCKERHUB_REPOSITORY:$IMAGE_TAG .
```
TaniBuilds, line 49: 🟠 **HIGH** · 🔒 security · `synth-001`

**Rotate GH_READONLY_PAT — secret may persist in previously published image history/provenance**

<sub>🤝 flagged by **2** agents: deep-analysis, adversarial</sub>

Prior to this PR, the workflow passed `secrets.GH_READONLY_PAT` as a Docker `--build-arg` and the Dockerfile declared `ARG GH_TOKEN` at the top of the builder stage. Even though `make install` does not consume the token, build-args and build environment values are recorded in Docker image build history (`docker history`/`docker inspect`) and may also leak into BuildKit provenance attestations and OCI annotations. Any image pushed to the public `alloranetwork/allora-chain` Docker Hub repo via this workflow before the fix can therefore expose the PAT to anyone with `docker pull` access. Removing the build-arg fixes the leak going forward, but the previously injected token value remains valid until rotated. This PR must be accompanied by (a) rotating/regenerating `GH_READONLY_PAT` in the org settings, and (b) auditing/removing historical Docker Hub image tags that contain the secret in their layer metadata. Without rotation, the security objective of DEVOP-586 is only half-achieved. Adversarial review independently confirmed the severity is correctly HIGH and the fix is sound.

🎭 _adversarial note: Adversarial verified the HIGH severity is justified and the suggested fix (rotate PAT + audit images) does not introduce new bugs._

💡 **Suggestion:** After merging, rotate secrets.GH_READONLY_PAT (and any user/installation token derived from it). Confirm with security/devops that no historical image tags retain the secret in their layer metadata; if any do, delete those tags or replace them. Document the rotation in the DEVOP-586 ticket.

Dockerfile
```diff
@@ -1,7 +1,5 @@
 FROM golang:1.22-bookworm AS builder
```
TaniBuilds, line 1: ⚪ **INFO** · 🔬 deep-analysis · `synth-003`

**Guard future private-dependency introductions with BuildKit secrets**

<sub>🤝 flagged by **2** agents: deep-analysis, adversarial</sub>

Verified the build does not currently need GH_TOKEN: go.mod has no `allora-network/*` dependencies (other than the module itself), the Dockerfile contains no `git config --global url."...".insteadOf` setup, there is no GOPRIVATE/.netrc step, and `make install` performs a vanilla `go build`. Removing the ARG and --build-arg cannot break the present Docker build. The second-order risk is that if private internal dependencies are ever added to go.mod in the future, the docker build will silently fail (or, worse, someone will reintroduce the insecure --build-arg pattern). Consider adding a CONTRIBUTING note or a Dockerfile-level mechanism (e.g. BuildKit `RUN --mount=type=secret,id=gh_token ...`) so that future private modules are wired up securely from day one. Adversarial confirmed sibling workflow `build_push_upgrader_docker_hub.yml` does not have the same anti-pattern, so fix scope is sufficient today.

💡 **Suggestion:** If private allora-network Go modules are added later, use Docker BuildKit secrets (`RUN --mount=type=secret,id=gh_token ...`) rather than --build-arg so the credential is not persisted in image layers or build provenance.

TaniBuilds (commented):
## 🔍 Multi-Agent Inline Review

**2 inline findings** posted as resolvable threads.

1 findings not on changed lines (see PR comment for full report): 🔵 1 low

<sub>Generated by <a href="https://github.com/allora-network/code-review-robonet">code-review-robonet</a></sub>

Dockerfile
srt0422: > Consider adding a CONTRIBUTING note or a Dockerfile-level mechanism (e.g. BuildKit `RUN --mount=type=secret,id=gh_token ...`) so that future private modules are wired up securely from day one.

Addressed in `16f4977a` — added a `SECURITY` comment to the top of the `builder` stage in `Dockerfile` warning future contributors not to pass credentials via `--build-arg`/`ARG` (they are embedded in image metadata and OCI provenance and leak to anyone with pull access), and pointing at BuildKit secrets (`RUN --mount=type=secret,id=...`) as the correct mechanism if a private Go module is ever needed. Keeps the guard right next to where someone would otherwise reintroduce the pattern.

.github/workflows/build_push_docker_hub.yml
srt0422: > This PR must be accompanied by (a) rotating/regenerating `GH_READONLY_PAT` in the org settings, and (b) auditing/removing historical Docker Hub image tags that contain the secret in their layer metadata. Without rotation, the security objective of DEVOP-586 is only half-achieved.

Needs human input — the in-repo fix (dropping `--build-arg "GH_TOKEN=…"` from the workflow and removing `ARG GH_TOKEN` from the `Dockerfile`) has shipped on this branch and closes the leak going forward. The remaining recommended actions are operational and live outside this repository:

1. **Rotate `secrets.GH_READONLY_PAT`** in `allora-network` org settings (and any user/installation token derived from it).
2. **Audit/remove historical Docker Hub image tags** on `alloranetwork/allora-chain` that may have the previously-injected PAT recorded in their layer metadata / OCI provenance (`docker history --no-trunc`, `docker inspect`); delete or replace any tags that retain it.
3. **(Low, from synth-002)** Once rotation is complete, remove `GH_READONLY_PAT` from the GitHub repo/org secrets if no other workflow consumes it, or document the remaining consumer.
4. **Document** the rotation + image audit on the DEVOP-586 ticket.

Flagging here so the operational owner can complete these steps; leaving this thread open until rotation/audit is confirmed.

xmariachi: @gh-allora  here we are removing this no-longer-in-use parameter. All good from my end. But make sure the steps are followed (ie. cleaning steps) and that there is indeed no impact in the deployment.

gh-allora: All good on this one already from cleanup 3 days ago.

Image

xmariachi (approved):

PR #951, opened 2026-05-14
feat(ci): adopt go-install-hardened workflow (Shai-Hulud defense)

srt0422:
Adopts the `go-install-hardened.yml` reusable workflow + pins Go env vars in `go.env` for module-integrity safety. Extends the Shai-Hulud supply-chain defense (currently rolled out for npm/PyPI) to Go.\n\n- Reusable workflow: allora-network/ci-workflows-private#17 (SHA-pinned to `fa8a0e1361216ec0c75e4d822f648248083e229b`; will re-pin to `@v1` after merge)\n- Skill: allora-network/skills#69\n- Linear umbrella: "Adopt go-install-hardened.yml across all 11 active Go repos"\n- `cgo_enabled: '1'`\n\n🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Adds `.github/workflows/go-hardened.yml` to run the reusable `go-install-hardened.yml` with `go_version: '1.23.5'` and `cgo_enabled: '1'`, plus `go.env` pinning `GOPROXY`, `GOSUMDB`, and `GOFLAGS=-mod=readonly`. The workflow runs on all PRs and pushes (main/dev/release-*), skips fork PRs, and cancels stale PR runs to make it a safe required check, extending the Shai‑Hulud defense and aligning with the Linear rollout.

- **Bug Fixes**
  - Workflow: re-pinned to the correct reusable workflow; removed `paths` filter; added `concurrency` for PRs; added fork-PR `if` guard; aligned push branches.
  - Policy/docs: expanded `go.env` guidance (local opt-in and `GOFLAGS` caveat), documented `,direct` fallback and `GOPRIVATE` approach.

<sup>Written for commit 615698986e003e7421c57190c5dab3745d117f0e. Summary will update on new commits. <a href="https://cubic.dev/pr/allora-network/allora-chain/pull/951?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.github/workflows/go-hardened.yml
+++ b/.github/workflows/go-hardened.yml
@@ -4,6 +4,7 @@ on:
     paths:
       - 'go.mod'
       - 'go.sum'
+      - 'go.env'
       - '**/*.go'
       - '.github/workflows/go-hardened.yml'
   push:
--- a/.github/workflows/go-hardened.yml
+++ b/.github/workflows/go-hardened.yml
@@ -1,0 +1,18 @@
+name: Go hardened install
+on:
+  pull_request:
+    paths:
+      - 'go.mod'
+      - 'go.sum'
+      - '**/*.go'
+      - '.github/workflows/go-hardened.yml'
+  push:
+    branches: [main, master, dev, staging]
+permissions:
+  contents: read
+jobs:
+  hardened-install:
+    uses: allora-network/ci-workflows-private/.github/workflows/go-install-hardened.yml@fa8a0e1361216ec0c75e4d822f648248083e229b
+    with:
+      go_version: stable
+      cgo_enabled: '1'
--- a/.github/workflows/go-hardened.yml
+++ b/.github/workflows/go-hardened.yml
@@ -13,7 +13,7 @@ permissions:
   contents: read
 jobs:
   hardened-install:
-    uses: allora-network/ci-workflows-private/.github/workflows/go-install-hardened.yml@fa8a0e1361216ec0c75e4d822f648248083e229b
+    uses: allora-network/ci-workflows-private/.github/workflows/go-install-hardened.yml@0d80a856ef798b01a39f65cb88a5cf7e74e4ebf9
     with:
       go_version: stable
       cgo_enabled: '1'
--- a/.github/workflows/go-hardened.yml
+++ b/.github/workflows/go-hardened.yml
@@ -15,5 +15,8 @@ jobs:
   hardened-install:
     uses: allora-network/ci-workflows-private/.github/workflows/go-install-hardened.yml@0d80a856ef798b01a39f65cb88a5cf7e74e4ebf9
     with:
-      go_version: stable
+      # Pinned to match the `toolchain` directive in go.mod (and the version
+      # used by golangci-lint.yml). Keeps hardened install checks deterministic
+      # across runs instead of tracking whatever "stable" resolves to.
+      go_version: '1.23.5'
       cgo_enabled: '1'
--- a/.github/workflows/go-hardened.yml
+++ b/.github/workflows/go-hardened.yml
@@ -1,22 +1,37 @@
 name: Go hardened install
 on:
+  # No `paths:` filter on purpose. A path-filtered workflow that is
+  # promoted to a required status check would block PRs that touch
+  # only docs/YAML, because GitHub does not auto-pass path-filtered
+  # required checks. The hardened install completes quickly when the
+  # module cache is warm, so the cost of running on every PR is
+  # acceptable in exchange for safe required-check eligibility.
   pull_request:
-    paths:
-      - 'go.mod'
-      - 'go.sum'
-      - 'go.env'
-      - '**/*.go'
-      - '.github/workflows/go-hardened.yml'
   push:
-    branches: [main, master, dev, staging]
+    branches: [main, dev, release-*]
 permissions:
   contents: read
+# Cancel in-flight hardened-install runs for the same PR when new
+# commits arrive; keep push-event runs (main/dev/release-*) intact.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 jobs:
   hardened-install:
+    # Skip on fork PRs: the reusable workflow lives in a private repo
+    # and a fork PR's GITHUB_TOKEN cannot resolve it, which would
+    # surface as a confusing not-found failure for external
+    # contributors. Internal PRs and push events always run.
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository
     uses: allora-network/ci-workflows-private/.github/workflows/go-install-hardened.yml@0d80a856ef798b01a39f65cb88a5cf7e74e4ebf9
     with:
-      # Pinned to match the `toolchain` directive in go.mod (and the version
-      # used by golangci-lint.yml). Keeps hardened install checks deterministic
-      # across runs instead of tracking whatever "stable" resolves to.
+      # Pinned to match the `toolchain` directive in go.mod. Keeps
+      # the hardened install check deterministic across runs instead
+      # of tracking whatever "stable" resolves to. Sibling workflows
+      # (golangci-lint.yml @ 1.23.6, format_and_test.yml @ 1.22.x)
+      # diverge today; centralizing on go.mod's toolchain is tracked
+      # as a follow-up.
       go_version: '1.23.5'
+      # cosmwasm/wasmvm (a transitive Cosmos SDK dep) requires CGO to
+      # link against libwasmvm; matches the goreleaser build path.
       cgo_enabled: '1'
--- a/go.env
+++ b/go.env
@@ -1,0 +1,5 @@
+# Shai-Hulud defense - Go module integrity pinning.
+# See allora-network/skills (skills/shai-hulud-defense) for rationale.
+GOPROXY=https://proxy.golang.org,direct
+GOSUMDB=sum.golang.org
+GOFLAGS=-mod=readonly
--- a/go.env
+++ b/go.env
@@ -1,5 +1,39 @@
 # Shai-Hulud defense - Go module integrity pinning.
-# See allora-network/skills (skills/shai-hulud-defense) for rationale.
+#
+# This file is documentation/reference only. Go does NOT auto-load a
+# repo-root go.env (only $GOROOT/go.env and $GOENV / ~/.config/go/env
+# are consulted by the `go` toolchain). The values below are enforced
+# in CI by the go-install-hardened reusable workflow (see
+# .github/workflows/go-hardened.yml), which exports them into
+# $GITHUB_ENV before any `go` command runs. Other Go workflows
+# (golangci-lint.yml, format_and_test.yml, goreleaser.yml,
+# build_push_docker_hub.yml) and local developer builds do NOT
+# currently consume this file; wiring them up is tracked separately.
+#
+# Local opt-in: `set -a; . ./go.env; set +a`. WARNING: sourcing
+# locally inherits GOFLAGS=-mod=readonly, which causes `go mod tidy`
+# and `go get` to fail with "updates to go.mod needed". Unset
+# GOFLAGS (or run those commands in a sub-shell) before doing module
+# maintenance.
+#
+# If/when this repo (or a transitive dep) takes on a private module,
+# add `GOPRIVATE=github.com/allora-network/<private-org-or-repo>`
+# rather than disabling GOSUMDB. Do not weaken GOSUMDB wholesale.
+#
+# Rationale skill:
+# https://github.com/allora-network/skills/tree/main/skills/shai-hulud-defense
+
+# Use Google's module proxy (cached, integrity-checked) with a direct
+# VCS fallback for modules the proxy hasn't seen yet (routine for
+# newly tagged or yanked modules). The ",direct" fallback is a known
+# trade-off: in a maintainer-compromise scenario the proxy may serve
+# a cached-clean version while direct would fetch the malicious tag.
+# Drop ",direct" if you want misses to fail loudly instead.
 GOPROXY=https://proxy.golang.org,direct
+
+# Verify every module against the Go checksum database.
 GOSUMDB=sum.golang.org
+
+# Prevent `go build` / `go test` from silently mutating go.mod.
+# See the WARNING above about sourcing this file locally.
 GOFLAGS=-mod=readonly


.github/workflows/go-hardened.yml
```diff
@@ -0,0 +1,18 @@
...
+      - '.github/workflows/go-hardened.yml'
+  push:
+    branches: [main, master, dev, staging]
+permissions:
+  contents: read
+jobs:
+  hardened-install:
+    uses: allora-network/ci-workflows-private/.github/workflows/go-install-hardened.yml@fa8a0e1361216ec0c75e4d822f648248083e229b
+    with:
+      go_version: stable
```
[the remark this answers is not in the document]
xmariachi: agree, otherwise lgtm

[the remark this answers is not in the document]
srt0422: Addressed in c233dd6a — pinned `go_version` from `stable` to `1.23.5` (matches the `toolchain` directive in `go.mod` and the version used by `golangci-lint.yml`). Hardened install checks are now deterministic across runs.

srt0422:
## Status update after re-sign + fix push

- ✅ All 4 commits re-signed (SSH/ED25519, `srt0422`) — `signature.state = VALID` on every commit (GitHub "Verified" badge).
- ✅ cubic-dev-ai (P2) + xmariachi review feedback addressed in `c233dd6a`: `go_version` pinned from `stable` → `1.23.5` (matches the `toolchain` directive in `go.mod` and the version used by `golangci-lint.yml`).
- ⚠️ **CI / GitHub Actions appears paused org-wide.** The most recent GitHub Actions workflow run on this repo (any branch, any actor) is `2026-05-18T11:57:44Z`. The `pull_request` / `push` events triggered by today's force-push did not create any new workflow runs — same is true for PRs `#947` and `#942`. Only check-app integrations (GitGuardian, cubic-dev-ai, Greptile Apps) are reaching the PR; GitHub Actions itself is silent. This needs an org-admin to check Actions billing / org-level Actions settings; it is not something the PR author can resolve from inside the PR.

Once Actions is re-enabled, the existing workflows (`format_and_test`, `golangci-lint`, `go-hardened`) should pick the new HEAD up automatically — the hardened-install workflow itself triggered + passed against `0d80a856ef798b01a39f65cb88a5cf7e74e4ebf9` on the previous push, and the only change since then is pinning `go_version` to a real release.

Re-requesting review from @xmariachi.

go.env
```diff
@@ -0,0 +1,5 @@
+# Shai-Hulud defense - Go module integrity pinning.
```
TaniBuilds, line 1: 🟡 **MEDIUM** · 🏗️ architecture · `synth-001`

**Repo-root `go.env` is not auto-loaded by Go tooling — Shai-Hulud hardening only applies to one CI job**

<sub>🤝 flagged by **4** agents: correctness, security, architecture, deep-analysis</sub>

Go only reads `$GOROOT/go.env` (system defaults) and `$GOENV`/`~/.config/go/env` (user). A `go.env` checked in at the repo root is NOT consulted by `go build`, `go mod download`, `go test`, etc. As a result, the hardening (`GOPROXY=https://proxy.golang.org,direct`, `GOSUMDB=sum.golang.org`, `GOFLAGS=-mod=readonly`) only takes effect inside the new reusable `go-install-hardened.yml` workflow IF that workflow explicitly exports them. The other module-fetching workflows in this repo — `format_and_test.yml`, `golangci-lint.yml`, `goreleaser.yml`, `build_push_docker_hub.yml` — all use `actions/setup-go` directly and do not source `go.env`. Local developer builds and `make install/build` also ignore the file. The Shai-Hulud defense advertised by this PR therefore covers a single new CI job; release artifacts (goreleaser) and the dev path remain unhardened, which is materially narrower than the file's existence implies. Four independent agents (correctness, security, architecture, deep-analysis) flagged this.

💡 **Suggestion:** Either (1) wire these values into the workflows that actually fetch modules (e.g., `cat go.env >> $GITHUB_ENV` in a setup step, or move the values into each workflow's `env:` block), or (2) rename/relocate the file to make its scope explicit (e.g., `.ci/go.env` or `docs/go.env.example`) and add a comment + CONTRIBUTING note explaining how to opt in locally (`set -a; source go.env; set +a`, or `go env -w ...`). At minimum, document explicitly which workflow consumes this file.

.github/workflows/go-hardened.yml
```diff
@@ -0,0 +1,22 @@
...
+      - '.github/workflows/go-hardened.yml'
+  push:
+    branches: [main, master, dev, staging]
+permissions:
+  contents: read
+jobs:
+  hardened-install:
+    uses: allora-network/ci-workflows-private/.github/workflows/go-install-hardened.yml@0d80a856ef798b01a39f65cb88a5cf7e74e4ebf9
+    with:
+      # Pinned to match the `toolchain` directive in go.mod (and the version
```
TaniBuilds: 🟡 **MEDIUM** · 🏗️ architecture · `synth-002`

**Go toolchain version is pinned in four diverging locations**

<sub>🤝 flagged by **4** agents: correctness, architecture, deep-analysis, style</sub>

The PR pins `go_version: '1.23.5'` in `.github/workflows/go-hardened.yml` and the comment claims this matches both `go.mod`'s `toolchain` directive AND `golangci-lint.yml`. In fact the repo now has four diverging sources of truth: `go.mod` (`go 1.22.11`, `toolchain go1.23.5`), `golangci-lint.yml` (`go-version: 1.23.6`), `format_and_test.yml` (`go-version: 1.22.x`), and the new `go-hardened.yml` (`1.23.5`). The comment is therefore inaccurate (1.23.5 vs 1.23.6) and will rot further on the next bump. More importantly, the hardened-install check certifies an install at a version that the lint and test jobs do not actually use, weakening the supply-chain reasoning the PR is trying to establish. Flagged by four independent agents.

💡 **Suggestion:** Centralize the Go version. Easiest: have the reusable workflow accept `go-version-file: go.mod` (or read a `.go-version` file) and update sibling workflows (`golangci-lint.yml`, `format_and_test.yml`) to source the same file. As a minimum scoped to this PR, either drop the misleading comment about matching golangci-lint.yml, or open a follow-up to bump golangci-lint.yml to 1.23.5 so the claim is true.

.github/workflows/go-hardened.yml
```diff
@@ -0,0 +1,22 @@
+name: Go hardened install
+on:
+  pull_request:
```
TaniBuilds, line 9: 🟡 **MEDIUM** · 🎭 adversarial · `synth-003`

**Path-filtered workflow used as a required status check will block PRs that don't touch Go files**

The hardened-install workflow only runs when one of `go.mod | go.sum | go.env | **/*.go | the workflow itself` changes. If this job is (or becomes) a required status check on `dev`/`main` branch protection, PRs that touch only docs, YAML, or non-Go assets will be stuck at 'Expected — Waiting for status to be reported' because GitHub does not auto-pass path-filtered required checks. This is a well-known GitHub Actions footgun and the most likely operational regression this PR can introduce. Phase A reviewers missed this branch-protection interaction; surfaced by adversarial.

🎭 _adversarial note: Phase B 'missed' finding — not flagged by any Phase A reviewer._

💡 **Suggestion:** Decide explicitly whether `hardened-install` should be a required check. If yes, either drop the `paths:` filter (so the job always runs and short-circuits when nothing relevant changed), or add a sentinel job with the same final check-name that always succeeds when paths are unaffected, or use a `workflow_call` aggregator. Document the decision in CONTRIBUTING.md.

.github/workflows/go-hardened.yml
```diff
@@ -0,0 +1,22 @@
...
+      - 'go.sum'
+      - 'go.env'
+      - '**/*.go'
+      - '.github/workflows/go-hardened.yml'
+  push:
+    branches: [main, master, dev, staging]
+permissions:
+  contents: read
+jobs:
+  hardened-install:
```
TaniBuilds, line 20: 🔵 **LOW** · 🏗️ architecture · `synth-004`

**Reusable workflow lives in a private repo — fork PRs cannot run the hardened check**

<sub>🤝 flagged by **3** agents: security, architecture, deep-analysis</sub>

`uses: allora-network/ci-workflows-private/.github/workflows/go-install-hardened.yml@<sha>` references a private repo. GitHub Actions cannot resolve a private reusable workflow when invoked from a fork PR (the fork's `GITHUB_TOKEN` lacks read access). External contributors' PRs will therefore see this job fail or skip with a not-found error, so the Shai-Hulud check the PR title advertises silently does not apply to community contributions, and red CI confuses drive-by contributors. Additionally, the workflow's internals cannot be audited from this PR; SHA pinning mitigates but does not eliminate trust placed in that private repo. Flagged by three agents (security, architecture, deep-analysis).

💡 **Suggestion:** Either (a) mirror `go-install-hardened.yml` into a public `ci-workflows` repo (still SHA-pinned), (b) add a job-level `if: github.event.pull_request.head.repo.full_name == github.repository || github.event_name == 'push'` guard so fork PRs cleanly skip, or (c) document in CONTRIBUTING.md that this check only gates internal PRs. Independently, audit the called workflow to confirm it pins all third-party actions to SHAs, uses minimum permissions, and sets `persist-credentials: false` on any checkout.

🔗 Cross-repo: affects **allora-network/ci-workflows-private**

go.env
```diff
@@ -0,0 +1,5 @@
...
+# See allora-network/skills (skills/shai-hulud-defense) for rationale.
+GOPROXY=https://proxy.golang.org,direct
+GOSUMDB=sum.golang.org
+GOFLAGS=-mod=readonly
```
TaniBuilds, line 39: 🔵 **LOW** · 🔬 deep-analysis · `synth-005`

**GOFLAGS=-mod=readonly in go.env breaks `go mod tidy` / `go get` if developers source it locally**

Placing `GOFLAGS=-mod=readonly` in a repo-root `go.env` invites developers to source it (`set -a; . ./go.env; set +a`). Once sourced, every `go` command — including `go mod tidy`, `go get`, `go mod download -x` — inherits `-mod=readonly` and fails with `updates to go.mod needed` instead of doing the work. The repo's `Makefile` already passes `-mod=readonly` explicitly to `go build`/`go install` where it is appropriate, so a global env-level pin is redundant for builds but harmful for module maintenance. Note: this is partly by-design (the constraint is the desired posture for Shai-Hulud), but the placement at repo root invites surprise. Adversarial challenged severity as info/by-design; keeping at low.

🎭 _adversarial note: Adversarial (adv-challenge-goflags-readonly) argues this is by-design: `-mod=readonly` is exactly the posture wanted in CI, and the file is not auto-sourced. The risk only materializes if a developer naively sources the file without understanding it. Severity may be info rather than low._

💡 **Suggestion:** Drop `GOFLAGS=-mod=readonly` from `go.env` and move it into the workflow step that benefits, OR add a clear comment in `go.env` warning that sourcing it disables `go mod tidy`/`go get`. Keep `GOPROXY`/`GOSUMDB` in the shared file.

.github/workflows/go-hardened.yml
```diff
@@ -0,0 +1,22 @@
+name: Go hardened install
```
TaniBuilds, line 1: 🔵 **LOW** · 🎭 adversarial · `synth-006`

**No `concurrency:` group — rapid pushes to a PR queue redundant hardened-install runs**

The workflow has no `concurrency:` block. Successive pushes to the same PR branch spawn parallel hardened-install jobs that all consume GitHub Actions minutes and pull modules from proxy.golang.org repeatedly. Best practice for PR-only verification workflows is to define a concurrency group keyed on workflow + ref and cancel in-progress runs for pull_request events. Phase A reviewers did not flag this.

🎭 _adversarial note: Phase B 'missed' finding._

💡 **Suggestion:** Add a `concurrency:` block: `concurrency: { group: ${{ github.workflow }}-${{ github.ref }}, cancel-in-progress: ${{ github.event_name == 'pull_request' }} }`.

go.env
```diff
@@ -0,0 +1,5 @@
+# Shai-Hulud defense - Go module integrity pinning.
+# See allora-network/skills (skills/shai-hulud-defense) for rationale.
+GOPROXY=https://proxy.golang.org,direct
```
TaniBuilds, line 32: 🔵 **LOW** · 🔬 deep-analysis · `synth-007`

**`GOPROXY=...,direct` fallback partially undercuts Shai-Hulud defense; no GOPRIVATE guidance**

<sub>🤝 flagged by **2** agents: deep-analysis, adversarial</sub>

`go.env` sets `GOPROXY=https://proxy.golang.org,direct` and `GOSUMDB=sum.golang.org`. Two related concerns: (1) the `direct` fallback means when proxy.golang.org returns 410/404 (routine for newly tagged/yanked modules), `go` fetches directly from the VCS — in a Shai-Hulud scenario where a maintainer is compromised and publishes a malicious tag, the proxy may have cached the clean version while `direct` would fetch the malicious one, so `direct` is a downgrade vector. (2) If this repo (or any transitive dep) ever needs a private/internal module, `direct` will query `sum.golang.org`, which cannot verify private modules and will fail the build, tempting a future developer to weaken hardening rather than add `GOPRIVATE`.

💡 **Suggestion:** Either drop the `,direct` fallback so misses fail loudly, or scope it via `GOPRIVATE=github.com/allora-network/*-private` for known private deps. Add an inline comment in `go.env` describing how to extend with `GOPRIVATE`/`GONOSUMCHECK` when private modules are introduced, so the next maintainer does not disable `GOSUMDB` wholesale.

TaniBuilds (commented):
## 🔍 Multi-Agent Inline Review

**10 inline findings** posted as resolvable threads.

<sub>Generated by <a href="https://github.com/allora-network/code-review-robonet">code-review-robonet</a></sub>

.github/workflows/go-hardened.yml
```diff
@@ -0,0 +1,22 @@
...
+on:
+  pull_request:
+    paths:
+      - 'go.mod'
+      - 'go.sum'
+      - 'go.env'
+      - '**/*.go'
+      - '.github/workflows/go-hardened.yml'
+  push:
```
TaniBuilds, line 10: ⚪ **INFO** · 🎨 style · `synth-008`

**Push trigger branch list (`main, master, dev, staging`) inconsistent with other workflows**

<sub>🤝 flagged by **3** agents: correctness, deep-analysis, style</sub>

`go-hardened.yml` triggers on push to `[main, master, dev, staging]`, while `golangci-lint.yml` uses `[main, dev, release-*]`. The repo does not appear to use `master` or `staging` branches. Harmless but adds noise/inconsistency across the CI suite and could drift further over time. Flagged by three agents.

💡 **Suggestion:** Align with the convention used by `golangci-lint.yml` — likely `branches: [main, dev, release-*]` — unless `master`/`staging` are intentionally reserved.

.github/workflows/go-hardened.yml
```diff
@@ -0,0 +1,22 @@
...
+  contents: read
+jobs:
+  hardened-install:
+    uses: allora-network/ci-workflows-private/.github/workflows/go-install-hardened.yml@0d80a856ef798b01a39f65cb88a5cf7e74e4ebf9
+    with:
+      # Pinned to match the `toolchain` directive in go.mod (and the version
+      # used by golangci-lint.yml). Keeps hardened install checks deterministic
+      # across runs instead of tracking whatever "stable" resolves to.
+      go_version: '1.23.5'
+      cgo_enabled: '1'
```
TaniBuilds, line 37: ⚪ **INFO** · 🎭 adversarial · `synth-009`

**`cgo_enabled: '1'` in hardened workflow is undocumented and diverges from sibling workflows**

The reusable workflow is called with `cgo_enabled: '1'`, but no other workflow in `.github/workflows/` sets `CGO_ENABLED` explicitly. The value is not stable across runner images. If the hardened install path uses CGO to compile dependencies (e.g., for `replace`d Cosmos modules) but goreleaser/format_and_test do not, the hardened check can certify an install that the release pipeline does not actually perform — reproducibility skew. Phase A did not flag this.

🎭 _adversarial note: Phase B 'missed' finding._

💡 **Suggestion:** Add a code comment justifying `cgo_enabled: '1'` (e.g., 'matches goreleaser build because cosmwasm requires CGO') and consider centralizing the value alongside `go_version`.

go.env
TaniBuilds, line 1: ⚪ **INFO** · 🎨 style · `synth-010`

**`go.env` references an unqualified `allora-network/skills` path**

The comment in `go.env` says `See allora-network/skills (skills/shai-hulud-defense) for rationale.` This is ambiguous — it doesn't indicate whether `allora-network/skills` is a public GitHub repo, an internal doc, or something else. External contributors may not be able to resolve the reference. Combined with the discovery that the file is not auto-loaded, a short inline rationale per setting would help.

💡 **Suggestion:** Replace with a full URL (e.g. https://github.com/allora-network/skills/tree/main/skills/shai-hulud-defense) or add a one-line description per variable (proxy/sumdb pinning, readonly to prevent silent module mutation).

srt0422: > Repo-root go.env is not auto-loaded by Go tooling — Shai-Hulud hardening only applies to one CI job

Addressed in 61569898: expanded the `go.env` header to make explicit that the file is documentation/reference only and is NOT consulted by Go from the repo root. Listed which workflows currently consume it (hardened install, via the reusable workflow exporting into `$GITHUB_ENV`) vs. which do not yet (`format_and_test.yml`, `golangci-lint.yml`, `goreleaser.yml`, `build_push_docker_hub.yml`, local `make install/build`). Wiring the other Go workflows to source these values is intentionally out of scope for this PR and tracked as the follow-up.

srt0422: > Go only reads `$GOROOT/go.env` (system defaults) and `$GOENV`/`~/.config/go/env` (user). A `go.env` checked in at the repo root is NOT consulted by `go build`, `go mod download`, `go test`, etc.

Addressed: Expanded the `go.env` header to make explicit that the file is documentation/reference only and NOT auto-loaded by Go; listed which workflows consume it vs. which ignore it; documented the local opt-in command alongside the side-effects warning. See commit 61569898.

.github/workflows/go-hardened.yml
srt0422: > The PR pins `go_version: '1.23.5'` … and the comment claims this matches both `go.mod`'s `toolchain` directive AND `golangci-lint.yml`. In fact the repo now has four diverging sources of truth

Addressed: Dropped the inaccurate "(and the version used by golangci-lint.yml)" parenthetical from the `go_version` comment; documented the current divergence (1.23.6 / 1.22.x / 1.23.5) inline and noted centralization on go.mod's `toolchain` as a follow-up. See commit 61569898.

.github/workflows/go-hardened.yml
srt0422: > If this job is (or becomes) a required status check on `dev`/`main` branch protection, PRs that touch only docs, YAML, or non-Go assets will be stuck

Addressed: Dropped the `pull_request` `paths:` filter so the job can safely be promoted to a required status check; added a header comment explaining the trade-off. See commit 61569898.

.github/workflows/go-hardened.yml
srt0422: > GitHub Actions cannot resolve a private reusable workflow when invoked from a fork PR (the fork's `GITHUB_TOKEN` lacks read access).

Addressed: Added an `if:` guard so fork PRs skip cleanly — `if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository`. See commit 61569898.

go.env
srt0422: > Once sourced, every `go` command — including `go mod tidy`, `go get`, `go mod download -x` — inherits `-mod=readonly` and fails with `updates to go.mod needed`

Addressed: Added an explicit WARNING in `go.env` that sourcing locally inherits `-mod=readonly` and breaks `go mod tidy` / `go get`, with instructions to `unset GOFLAGS` before module-maintenance commands. See commit 61569898.

.github/workflows/go-hardened.yml
srt0422: > The workflow has no `concurrency:` block. Successive pushes to the same PR branch spawn parallel hardened-install jobs

Addressed: Added a `concurrency:` block keyed on workflow+ref, with `cancel-in-progress` scoped to `pull_request` events only so push runs on `main`/`dev`/`release-*` are not cancelled. See commit 61569898.

go.env
srt0422: > the `direct` fallback means when proxy.golang.org returns 410/404 … in a Shai-Hulud scenario where a maintainer is compromised and publishes a malicious tag, the proxy may have cached the clean version while `direct` would fetch the malicious one

Addressed: Documented the `,direct` trade-off inline in `go.env` and added `GOPRIVATE` guidance for the day a private module dep appears so the next maintainer doesn't disable `GOSUMDB` wholesale. Kept `,direct` (with documented rationale) rather than dropping it, to avoid breaking newly-tagged module fetches. See commit 61569898.

.github/workflows/go-hardened.yml
srt0422: > `go-hardened.yml` triggers on push to `[main, master, dev, staging]`, while `golangci-lint.yml` uses `[main, dev, release-*]`

Addressed: Aligned push branches with the `golangci-lint.yml` convention — `[main, dev, release-*]`. See commit 61569898.

.github/workflows/go-hardened.yml
srt0422: > The reusable workflow is called with `cgo_enabled: '1'`, but no other workflow in `.github/workflows/` sets `CGO_ENABLED` explicitly … reproducibility skew

Addressed: Added an inline comment explaining `cgo_enabled='1'` — `cosmwasm/wasmvm` (a transitive Cosmos SDK dep) requires CGO; matches the goreleaser build path. See commit 61569898.

go.env
srt0422: > The comment in `go.env` says `See allora-network/skills (skills/shai-hulud-defense) for rationale.` This is ambiguous

Addressed: Replaced the ambiguous reference with the full GitHub URL: `https://github.com/allora-network/skills/tree/main/skills/shai-hulud-defense`. See commit 61569898.

xmariachi (approved):

PR #957, opened 2026-06-15
Diego/942 p0 p1 fixes (all types) ENGN-8600

xmariachi:

Fixes from P0-P1 fixes on #942 from [this list](https://github.com/allora-network/allora-chain/pull/942#pullrequestreview-4467631766)

- [x] If tested, please describe how. If not, why tests are not needed. -- unit tests
- [x] If documented, please describe where. If not, describe why docs are not needed. -- where appropriate, on CHANGELOG mostly.

---
## Summary by cubic
Hardens worker emissions validation and event handling, fixes label padding for MULTI topics to use `LabelDefaultValue`, fixes one‑out forecast recompute to exclude the withheld inferer, makes `max_canonical_label_byte_length` governable, and makes the v15 inference migration idempotent. Also aligns `create-topic` AutoCLI args, gates the latest outlier‑resistant query on topic arity, and keeps the reputer bundle API compatible (ENGN-8600).

- **Bug Fixes**
  - Enforce registry caps in `InsertWorkerPayload` without partial writes (`ErrEpochLabelRegistrySaturated`); reject non‑whitelisted labels (`ErrLabelNotInWhitelist`).
  - Validate `CompactRegistryAndRemapInferences` inputs, including empty inferer and topic/nonce mismatches.
  - Emit forecaster events only when forecast elements are stored.
  - Pad missing MULTI‑label values with `topic.LabelDefaultValue` in `ConvertInferenceValuesFromProto`; plumb through synthesis (forecast‑implied, one‑out) and add unit/e2e tests.
  - Exclude the withheld inferer in one‑out forecast‑implied recomputes; simplify the single‑inferer path.
  - Guard v15 inference migrations against re‑runs to avoid zeroing values; add idempotence tests.
  - Gate `GetLatestNetworkInferencesOutlierResistant`: return `InvalidArgument` for MULTI topics; `NotFound` for unknown topics.

- **Refactors**
  - Support `MaxCanonicalLabelByteLength` via `UpdateParams` with unit/integration tests.
  - Align AutoCLI `create-topic` positional args: `topic_type`, `output_arity`, `require_unity`, `unity_tolerance`.
  - Use `types.NewReputerValueBundleFromInput`; keep deprecated `types.NewInputReputerValueBundleFromInput` for compatibility.
  - Update CHANGELOG to note removal of `types.NewInferenceFromInput`, `types.NewInferenceForecastBundleFromInput`, and `types.NewWorkerDataBundleFromInput`; add the multi‑label error note for outlier‑resistant queries.
  - Contributor docs: add self‑contained code comment guidance in `AGENTS.md` and `CLAUDE.md`.

<sup>Written for commit 394c3ed7d938186e73a8e3c8425f3450a0ee7af5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/allora-network/allora-chain/pull/957?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


x/emissions/keeper/msgserver/msg_server_reputer_payload.go
```diff
@@ -27,7 +27,7 @@ func (ms msgServer) InsertReputerPayload(ctx context.Context, msg *types.InsertR
 	if err != nil {
 		return nil, errorsmod.Wrapf(err, "Error getting params for reputer: %v", &msg.ReputerValueBundle.ValueBundle.Reputer)
 	}
-	rvb, err := types.NewLossBundleFromInput(msg.ReputerValueBundle)
+	rvb, err := types.NewReputerValueBundleFromInput(msg.ReputerValueBundle)
```
zale144, line 30: why rename this back though?

x/emissions/proto/emissions/v10/tx.proto
```diff
@@ -266,6 +266,9 @@ message OptionalParams {
     (gogoproto.customtype) = "github.com/allora-network/allora-chain/math.Dec",
     (gogoproto.nullable) = false
   ];
+  // Max byte length of a canonical label name after NFC normalization and
+  // trimming.
+  repeated uint64 max_canonical_label_byte_length = 61;
```
zale144, line 271: why do we need this again?

zale144 (commented):
just two comments, otherwise looks good

x/emissions/keeper/msgserver/msg_server_reputer_payload.go
xmariachi: I think introducing "loss bundle" concept when we use ReputerValueBundle field may be confusing. 
I think ReputerValueBundle (or ReputerBundle) is more general for a reputer submission, eg if we want to add more stuff in the future (not only losses) it would fit well.

x/emissions/proto/emissions/v10/tx.proto
xmariachi: To prevent people from defining interminable labels, and it's a param so we can change it via module param upgrade.

x/emissions/keeper/inference_synthesis/network_inference_builder.go
```diff
@@ -420,6 +424,28 @@ func calcOneOutInfererInference(args CalcOneOutInfererInferenceArgs) (
...
+		// mirroring GetOneOutInfererForecastImpliedInferences.
+		remainingForecasterToForecast := make(map[Forecaster]*emissions.Forecast, len(args.ForecasterToForecast))
+		for forecaster, forecast := range args.ForecasterToForecast {
+			filteredForecastElements := make([]*emissions.ForecastElement, 0, len(forecast.ForecastElements))
+			for _, element := range forecast.ForecastElements {
+				if element.Inferer != args.WithheldInferer {
+					filteredForecastElements = append(filteredForecastElements, element)
+				}
+			}
+			if len(filteredForecastElements) == 0 {
```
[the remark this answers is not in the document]
xmariachi: Looked into this — it doesn't skew the one-out inference. Tracing a dropped forecaster `F` (all of its forecast elements pointed at the withheld inferer, so it's removed by the `continue` at line 437):

1. **Final weighted average (`calcWeightedInference`).** The forecaster loop looks up `forecasterToForecastImpliedInference[F]`, finds nothing (F produced no implied inference), and `continue`s *before* `accumulateWeights`. So F lands in neither the numerator nor the denominator (`sumWeights`) of `Σ wₗ·Iₗ / Σ wₗ`. It's correctly absent from the actual weight normalization.

2. **`maxRegret` (`CalcWeightsGivenWorkers`).** F's regret *does* enter here — this is its only residual influence. But the weight kernel is `wₗ = Exp1DivExp1(a, bₗ) = (e^a + 1) / (e^{bₗ} + 1)`, where `a = -pNorm·(maxRegret − cNorm)` is **identical for every worker**. That common `(e^a + 1)` factor cancels in the ratio:

   ```
   Σ wₗ·Iₗ / Σ wₗ  =  Σ Iₗ/(e^{bₗ}+1)  /  Σ 1/(e^{bₗ}+1)
   ```

   `maxRegret` drops out entirely, so the result is invariant to F being included in it. (The weights are local to `calcOneOutInfererInference` and only ever consumed by this ratio.)

3. **MAD regret-scale.** `RegretScalePlusEpsilon` is precomputed at the topic level and threaded straight through, so the `> 0` guard skips the MAD path — the one other place a stray regret could matter non-uniformly. F never touches it.

The only theoretical exception is the `sumWeights < epsilonSafeDiv` clamp, where the denominator becomes a fixed constant instead of `(e^a+1)·Σh`. But every `(e^{bₗ}+1) > 1` and the argmax worker has weight 1, so `sumWeights` is normally ≥ 1 ≫ ε; reaching the clamp needs astronomically large regret gaps and yields a degenerate near-zero inference regardless. Not reachable in practice.

Net: no skew to the one-out network inference. What remains is cosmetic — a weight is computed for F and then discarded. Closing as not-a-bug. If we want the two sets to stay consistent we could pass the filtered forecaster keyset to the weight call, but it doesn't change the output.

PR #958, opened 2026-06-16
fix: SMELL-03 label default value usage and tests

xmariachi:
###  Summary
Addresses review finding SMELL-03 (P1) on [#942](https://github.com/allora-network/allora-chain/pull/942) ([comment](https://github.com/allora-network/allora-chain/pull/942#discussion_r3388159608)).

ConvertInferenceValuesFromProto right-padded short MULTI-arity inference vectors with alloraMath.ZeroDec(), while the rest of the chain (NormalizeInputInference, close-time CompactRegistryAndRemapInferences) treats an unset label slot as topic.LabelDefaultValue — which may be non-zero when require_unity == false. This PR makes the conversion pad with LabelDefaultValue so the proto→internal representation matches keeper semantics.

###  Severity note (latent, not live)
This is a latent mismatch, not active data corruption. The only callers of ConvertInferenceValuesFromProto run during topic close on compacted inferences — CompactRegistryAndRemapInferences already rebuilds every vector to full registry length, padded with LabelDefaultValue. So at the call site len(values) == len(labels), copy() overwrites the whole slice and the zero-fill is never observed today; the query path returns a stored bundle and never re-synthesizes. It's still worth fixing: the guard silently padded-short / rejected-long, and a unit test asserted the wrong 0 semantic — a correctness landmine for any future on-demand synthesis path.

###  Changes
- x/emissions/types/conversions.go — ConvertInferenceValuesFromProto takes a labelDefaultValue alloraMath.Dec and pads unset trailing slots with it instead of ZeroDec().
- x/emissions/keeper/inference_synthesis/ — threaded topic.LabelDefaultValue through the synthesis args, mirroring the existing NumLabels plumbing, only on the structs that actually reach CalcForecastImpliedInferences (CalcNetworkInferencesArgs, CalcForecastImpliedInferencesArgs, GetOneOutInfererForecastImpliedInferencesArgs, CalcOneOutInfererInferenceArgs / GetOneOutInfererInferencesArgs), sourced from topic.LabelDefaultValue in GetCalcNetworkInferenceArgs.
- x/emissions/testutil/test_utils.go — added a WithLabelDefaultValue topic-setup option for integration tests.
Graceful by design: if a short vector ever reaches the conversion, it degrades to the correct value rather than hard-erroring in EndBlock.

###  Tests
- TestConvertInferenceValuesFromProto rebuilt as a 20-case matrix: SINGLE (default ignored; NaN/Inf rejected), MULTI guards, unspecified-arity (default switch branch), exact-length (no default leak), padding with zero/negative/fractional/large defaults + extreme-short + pad-by-one, and NaN/Inf-default rejected only when it actually lands.
- TestCalcForecastImpliedInferencesMultiArityPadsWithLabelDefaultValue and …WeightedPadsWithLabelDefaultValue — cover both synthesis call sites (MEDIAN and WEIGHTED branches).
- TestMultiLabelNonZeroDefaultValuePadsMissingLabels — end-to-end submit→close→synthesis with a non-zero default, locking the semantic through the full pipeline.
Both synthesis guards are verified to fail on the reverted one-line change (want -1, got 0).

---
## Summary by cubic
Use topic `LabelDefaultValue` to pad missing label slots in MULTI inferences and wire it through synthesis and network inference paths. This fixes zero-padding divergence and keeps stored inferences, forecast-implied synthesis, and rewards consistent (addresses SMELL-03).

- **Bug Fixes**
  - `types`: `ConvertInferenceValuesFromProto` pads trailing slots with `LabelDefaultValue` (not zero) and validates padded values; SINGLE arity unchanged.
  - `inference_synthesis`: Plumb `LabelDefaultValue` into `CalcForecastImpliedInferences` and one-out helpers; applies to both MEDIAN and WEIGHTED branches.
  - `network_inferences`: Pass `topic.LabelDefaultValue` into calc args for all synthesis paths.
  - `testutil`: Add `WithLabelDefaultValue` to set non-zero defaults in topic creation.
  - Tests: Add MULTI padding tests for both branches, expand conversion tests to cover various defaults, add end-to-end rewards test for non-zero defaults, and update existing tests to pass `LabelDefaultValue` (zero by default).

<sup>Written for commit 47262208ab21d048cc5a6b42785d6234dc3fcedc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/allora-network/allora-chain/pull/958?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/x/emissions/keeper/inference_synthesis/forecast_implied.go
+++ b/x/emissions/keeper/inference_synthesis/forecast_implied.go
@@ -28,6 +28,9 @@ type CalcForecastImpliedInferencesArgs struct {
 	RegretScalePlusEpsilon alloraMath.Dec
 	LabelRegistry          *emissionstypes.EpochLabelRegistry
 	NumLabels              int
+	// LabelDefaultValue is topic.LabelDefaultValue, used to pad unset trailing
+	// slots when converting stored inferences (mirrors keeper semantics).
+	LabelDefaultValue alloraMath.Dec
 }
 
 // Calculate the forecast-implied inferences I_ik given inferences, forecasts and network losses.
@@ -85,7 +88,7 @@ func CalcForecastImpliedInferences(args CalcForecastImpliedInferencesArgs) (map[
 					blockHeight = inf.BlockHeight
 				}
 
-				iv, err := emissionstypes.ConvertInferenceValuesFromProto(args.TopicArity, labels, inf)
+				iv, err := emissionstypes.ConvertInferenceValuesFromProto(args.TopicArity, labels, args.LabelDefaultValue, inf)
 				if err != nil {
 					return nil, err
 				}
@@ -196,7 +199,7 @@ func CalcForecastImpliedInferences(args CalcForecastImpliedInferencesArgs) (map[
 				blockHeight = inf.BlockHeight
 			}
 
-			iv, err := emissionstypes.ConvertInferenceValuesFromProto(args.TopicArity, labels, inf)
+			iv, err := emissionstypes.ConvertInferenceValuesFromProto(args.TopicArity, labels, args.LabelDefaultValue, inf)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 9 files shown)


zale144 (approved):
lgtm

PR #964, opened 2026-07-03
clamp emitted values on events

xmariachi:
## Summary

Very small or very large `Dec` values produced by the network can serialize into
unnecessarily long strings when emitted in events. This clamps their magnitude at
event-construction time so event payloads stay compact. Event output only —
consensus state and hashing are unaffected.

## Change

- Add `alloraMath.ClampMagnitude(d, min, max Dec) Dec`: clamps `|d|` into a given
  window, preserving sign; zero/NaN/non-finite pass through. It never errors, so
  it is safe on the block-emission path.
- Apply it in `x/emissions/types/events_utils.go` to the `Dec` value fields of
  emissions events, using an event-owned window constant (`[1e-40, 1e40]`). The
  clamp is folded into the existing `convertArray` extraction (single pass).

## Scope

- Events only. State is marshaled separately and unchanged; events are not part of
  the app hash or state-sync snapshots, so there is no migration or
  consensus-version bump. Values already bounded at ingress are left as-is.

## Testing

- Table tests for `ClampMagnitude`: boundaries, sign, zero/NaN/infinity, degenerate
  window, input-not-mutated.
- Table tests over the clamping event constructors and both bundle paths (including
  NaN pass-through and empty→nil).
- `go build ./...`, tests, and `golangci-lint` clean.

### DIFF
--- a/math/bounded_dec.go
+++ b/math/bounded_dec.go
@@ -171,6 +171,38 @@ func MustNewCappedBoundedExp40DecFromString(s string) BoundedExp40Dec {
 	return dec
 }
 
+// ClampMagnitude clamps |d| into [minMagnitude, maxMagnitude], keeping the sign.
+// Zero, NaN and non-finite pass through. Total: never errors, so it is safe on
+// the event-emission path. Bounds must be positive with min <= max.
+// Purpose: cap the serialized textual size of a Dec (a tiny value like 1e-38000
+// otherwise expands to a ~38 KB fixed-notation string). Does not touch state.
+func ClampMagnitude(d, minMagnitude, maxMagnitude Dec) Dec {
+	if !d.IsFinite() || d.IsZero() {
+		return d
+	}
+	abs, err := d.Abs()
+	if err != nil {
+		return d
+	}
+	var bound Dec
+	switch {
+	case abs.Lt(minMagnitude):
+		bound = minMagnitude
+	case abs.Gt(maxMagnitude):
+		bound = maxMagnitude
+	default:
+		return d // in range
+	}
+	if !d.IsNegative() {
+		return bound
+	}
+	neg, err := bound.Neg() // restore sign
+	if err != nil {
+		return d
+	}
+	return neg
+}
+
 // Returns the maximum value that can be represented by a BoundedExp40Dec
 func GetMaxPositiveBoundaryExp40Dec() BoundedExp40Dec {
 	return BoundedExp40Dec{dec: maxBoundValue}
--- a/x/emissions/README.md
+++ b/x/emissions/README.md
@@ -43,7 +43,7 @@ See `x/emissions/metrics/` for details.
 
 ## Event value representation
 
-`Dec` values placed in emissions events are magnitude-clamped to `[1e-40, 1e40]` so extremely small or large values stay compact when serialized. This shapes event and query output only; consensus state is stored separately and is not clamped, so an event value may differ from the stored value at the extremes. See `ClampMagnitude` and the `clamp*` helpers in `x/emissions/types/events_utils.go`.
+`Dec` values placed in emissions events are magnitude-clamped to `[1e-40, 1e40]` so extremely small or large values stay compact when serialized. This shapes event output only; consensus state is stored separately and is not clamped, so an event value may differ from the stored value at the extremes. See `ClampMagnitude` and the `clamp*` helpers in `x/emissions/types/events_utils.go`.
 
 ## Topic configuration updates (UpdateTopic)
 
--- a/x/emissions/README.md
+++ b/x/emissions/README.md
@@ -41,6 +41,10 @@ query/tx: `allora_emissions_request_total` for occurrences, `allora_emissions_re
 Different labels are applied where appropriate (eg "topic_id", "address", "nonce", etc.)
 See `x/emissions/metrics/` for details.
 
+## Event value representation
+
+`Dec` values placed in emissions events are magnitude-clamped to `[1e-40, 1e40]` so extremely small or large values stay compact when serialized. This shapes event and query output only; consensus state is stored separately and is not clamped, so an event value may differ from the stored value at the extremes. See `ClampMagnitude` and the `clamp*` helpers in `x/emissions/types/events_utils.go`.
+
 ## Topic configuration updates (UpdateTopic)
 
 The `UpdateTopic` tx is intentionally limited to a small set of mutable fields. Today, topic creators can update:
--- a/x/emissions/types/events_utils.go
+++ b/x/emissions/types/events_utils.go
@@ -1,6 +1,8 @@
 package types
 
 import (
+	"fmt"
+
 	"cosmossdk.io/math"
 	"github.com/cosmos/gogoproto/proto"
 
@@ -24,7 +26,7 @@ func NewScoresSetEventBase(actorType ActorType, scores []Score) proto.Message {
 		TopicId:     topicId,
 		BlockHeight: blockHeight,
 		Addresses:   addresses,
-		Scores:      scoreValues,
+		Scores:      clampDecs(scoreValues),
 	}
 }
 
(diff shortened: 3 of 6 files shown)


fstecker (dismissed):

xmariachi (commented):
## Review summary

Reviewed for correctness, consensus safety, and scope. **No blocking issues (P0–P2) found** — the change is well-scoped and well-tested. Two central safety claims both hold up:

- **Consensus safety.** The clamped `Dec` values flow only to the ABCI event manager (`EmitTypedEvent`); they are never written to state, hashed into the app hash, included in state-sync snapshots, or read back by a query. No migration or `ConsensusVersion` bump is required.
- **"Values already bounded at ingress are left as-is."** The only Dec-carrying event constructors left unclamped source their values from `BoundedExp40Dec` proto fields, which already enforce the same `[1e-40, 1e40]` window at decode time.

`ClampMagnitude` is also correct at the extremes it targets: `Abs()` is a pure sign flip (errors only on NaN, which is already excluded), and `Neg()` is applied to the small-coefficient bound constant rather than to the large input, so there is no error path that could leak an unclamped value back out. NaN / ±Inf / zero pass through unchanged (all serialize to short strings). Test coverage over the boundaries, sign handling, and both bundle paths is thorough.

## One design decision worth an explicit ack: the lower clamp is a floor, not a truncation-to-zero

The clamp bounds *magnitude* into `[1e-40, 1e40]` while preserving sign. On the **low** side this means any nonzero finite value whose magnitude is below `1e-40` has its magnitude replaced by exactly `1e-40` — it is raised **up** to the floor, not collapsed to `0`:

- `1e-200`   → `1e-40`
- `-1e-200`  → `-1e-40`
- `1e-38000` → `1e-40`
- `0` (literal zero) → stays `0`

So a value like `1e-200`, which is extremely close to zero, is reported in the event as `1e-40` — roughly 10^160× larger than the true value.

This is a defensible choice:

- it preserves the fact that the value was **nonzero and had a sign**, which a clamp-to-zero would discard;
- events are display/indexing output only, so this never affects consensus.

The tradeoff is on the **consumer** side, and is what's worth acknowledging explicitly:

- a consumer that **sums** event values gets a small upward bias, since many genuinely-near-zero entries each become `1e-40`;
- a consumer testing exact `== 0` is unaffected (zero stays zero), but one testing "effectively zero" via a magnitude threshold will now see `1e-40` where the true value was far smaller;
- the event value therefore diverges from the stored state value at the low extreme (as the README notes).

If reporting sub-`1e-40` magnitudes as `0` (losing sign, but not inflating near-zero values) were preferred instead, that's a small change on the low branch of the clamp. Either behavior is reasonable — flagging it purely so downstream indexer/explorer owners can confirm the floor semantics match their expectations before this ships.

Otherwise LGTM.

x/emissions/README.md
```diff
@@ -41,6 +41,10 @@ query/tx: `allora_emissions_request_total` for occurrences, `allora_emissions_re
 Different labels are applied where appropriate (eg "topic_id", "address", "nonce", etc.)
 See `x/emissions/metrics/` for details.
 
+## Event value representation
+
+`Dec` values placed in emissions events are magnitude-clamped to `[1e-40, 1e40]` so extremely small or large values stay compact when serialized. This shapes event and query output only; consensus state is stored separately and is not clamped, so an event value may differ from the stored value at the extremes. See `ClampMagnitude` and the `clamp*` helpers in `x/emissions/types/events_utils.go`.
```
zale144: This says clamping shapes "event **and query** output". I don't think the query part is accurate: the query server endpoints (e.g. `GetNetworkInferencesAtBlock`, `GetLatestNetworkInferences`) read the `NetworkInferenceBundle` directly from state and return it without going through these `clamp*` helpers. State is stored unclamped (proto serializes `Dec` in scientific notation, so it stays compact), so query responses will still contain the full-precision, unclamped values.

Either drop "and query" here, or, if clamped query output is actually desired, the query path needs the same treatment. As written the doc over-promises.

x/emissions/types/events_utils.go
```diff
@@ -285,6 +289,33 @@ func convertArray[I, O any](in []I, get func(I) O) (out []O) {
 	return
 }
 
+// Magnitude window for Dec values emitted in events, so tiny/huge computed
+// values don't serialize into oversized strings. Independent of the ingress
+// BoundedExp40Dec guard; shapes event output only, never consensus state.
+const eventDecClampExponent = 40
```
zale144, line 295: Two design questions on the `1e±40` window:

1. **Lossy vs. lossless.** Clamping discards information (`1e-37962` → `1e-40`). The original problem is purely a *serialization* issue — a tiny `Dec` only bloats because `String()`/`MarshalJSON` render fixed-point. Serializing event `Dec`s in scientific notation would be both compact *and* lossless for every magnitude. Was that considered and rejected? If so, a one-line rationale here (e.g. downstream parsers expect fixed-point, or a global marshaler change was too risky) would save the next reader the investigation.

2. **Upper bound.** `1e40` as a paranoia cap is reasonable for tiny-value bloat, but for computed fields that aren't ingress-bounded (regrets, scores, weights, topic weights) it could silently truncate a legitimately large magnitude and misrepresent it. Are any emitted fields expected to legitimately exceed `1e40`? A brief note on that would help.

Minor maintainability note (non-blocking): this relies on every current and future `Dec`-bearing event field remembering to call `clampDec`/`clampDecs`. A new field or new event can silently bypass it. The enumeration test helps today, but a single choke point (e.g. a marshaler used only on the event path) would be more regression-proof.

x/emissions/types/events_utils.go
```diff
@@ -265,7 +269,7 @@ func ValueBundleToEventValueBundleBase(bundle *ValueBundle) *EventValueBundle {
 		}
 		for _, cell := range row.OneOutInfererValues {
 			if j, ok := infererIndex[cell.Worker]; ok {
-				oo[j] = cell.Value
+				oo[j] = clampDec(cell.Value)
```
zale144, line 272: Good — this is the field from the original bug report (`oneOutInfererForecasterValues`). Confirms the nested matrix path is covered, not just the top-level arrays.

zale144 (commented):
Solid, well-scoped mitigation for the event-payload bloat, and the tests are thorough (edge cases, non-mutation, degenerate windows, output-size assertion). It correctly shapes only the event path and leaves consensus state untouched.

A few things worth addressing before merge — one is a correctness issue in the docs (the "query output" claim), and the rest are design/robustness notes I've left inline. TL;DR:

1. **README claim about query output looks wrong** — query endpoints read the bundle straight from state and don't go through these `clamp*` helpers, so they still return unclamped values. See inline comment.
2. **Clamping is lossy; scientific notation would be lossless + compact.** Worth a sentence on why clamping was chosen over just serializing `Dec` in scientific notation for events.
3. **Lower clamp turns a legit tiny value into a fake `1e-40`** (not zero) — is that the intended semantics vs. clamping to `0`?
4. **This is a presentation-layer fix, not an input-validation fix.** The root cause is that `ValidateDec` allows arbitrary-magnitude finite `Dec` inputs, which is what lets a forecaster produce `1e-37962` in the first place. Worth tracking a follow-up to bound ingress magnitude so tiny values never get computed/persisted, since state is still stored unclamped.

Nice work overall — none of these block the immediate goal of shrinking event strings.

math/bounded_dec.go
```diff
@@ -171,6 +171,38 @@ func MustNewCappedBoundedExp40DecFromString(s string) BoundedExp40Dec {
 	return dec
 }
 
+// ClampMagnitude clamps |d| into [minMagnitude, maxMagnitude], keeping the sign.
+// Zero, NaN and non-finite pass through. Total: never errors, so it is safe on
+// the event-emission path. Bounds must be positive with min <= max.
+// Purpose: cap the serialized textual size of a Dec (a tiny value like 1e-38000
+// otherwise expands to a ~38 KB fixed-notation string). Does not touch state.
+func ClampMagnitude(d, minMagnitude, maxMagnitude Dec) Dec {
```
zale144, line 179: Function is correct and handles the edge cases well (NaN/inf/zero pass-through, sign preservation). Two small notes:

1. **Sub-`min` values become a fake non-zero.** A value smaller than `minMagnitude` is raised to `minMagnitude` rather than collapsed to `0`. So `1e-37962` is reported as `1e-40` — semantically "tiny but present" rather than "effectively zero". If a consumer treats `1e-40` as a meaningful weight/regret, that's a (small) distortion. Worth confirming that's the intended semantics vs. clamping tiny magnitudes to zero.

2. **Contract isn't enforced.** The doc says "bounds must be positive with `min <= max`" but nothing checks it. It's fine for the internal fixed-constant caller, but this is exported from `math/`; a future caller passing `min > max` or a negative bound gets silently wrong results. Consider a cheap guard/panic on invalid bounds, or note that validation is the caller's responsibility.

x/emissions/types/events_utils.go
xmariachi: known

math/bounded_dec.go
xmariachi: not a problem / nit

x/emissions/README.md
xmariachi: corrected

zale144 (approved):

PR #968, opened 2026-07-21
Max num of inferers at topic level (with global ranges)

xmariachi:

Moves the cap on how many inferers a topic admits from a single network-wide number to a per-topic setting, bounded by a global range. Classification topics need a shorter inferer set than scalar ones, and until now every topic was stuck with the same global value.

What this delivers, at a glance:

- Topic creators can now size their own active inferer set, instead of every topic inheriting a single network-wide number.
- The network keeps the last word: the global params define a range, enforced both when a topic is configured and again on every admission, so a topic can never operate outside what governance allows.
- A floor was added so a topic cannot be configured to run with an unreasonably small set. That closes a Sybil-adjacent hole where a single inferer could collect a topic's entire inference reward with a zero score and no competitive pressure.
- The floor is active from day one (default 5), so the protection does not wait for a governance vote. The trade-off is that the global ceiling can no longer be lowered below 5 without lowering the floor first.
- Turnout is not policed: an epoch in which fewer inferers than the floor take part is still valid and still pays out. The floor constrains configuration, not participation.
- Moving a global never breaks existing topics. Stored values are kept as they are even if a later change puts them outside the range, and only the effective value applies.
- The upgrade is behaviour-neutral: existing topics are backfilled with the value they were already effectively using, so nothing changes for participants at the upgrade height.
- Mid-epoch safety: the cap cannot be changed while a worker submission window is open, so admission for an in-flight epoch is never re-based under the participants.

Proto changes are **purely additive** (v10 kept, no renumbering), and this ships as chain upgrade **`v0.18.0`** (emissions `ConsensusVersion` 15 → 16).

## Design and decisions

**Per-topic cap.** New additive field `Topic.max_top_inferers_to_reward` (31), also on `CreateNewTopicRequest` (30) and `UpdateTopicRequest` (12). Settable at create and via `UpdateTopic`, which is a full replacement: `0` means "use the global maximum", not "leave unchanged", consistent with every other field on that message. Documented on the proto field and in the `update-topic` help text.

**Global floor.** New additive global `min_top_inferers_to_reward` (64 on `emissions.v9.Params` and `emissions.v10.OptionalParams`), so it reaches `UpdateParams`, genesis, the `GetParams` query and the `ParamsSet` event.

**One resolution rule, in one place.** `types.EffectiveMaxTopInferersToReward(topicCap, globalMin, globalMax)` is the single definition, used by admission and by create/update:

```go
EffectiveMaxTopInferersToReward(3,  5, 32) // => 5   floor raises it
EffectiveMaxTopInferersToReward(6,  5, 32) // => 6   within range, kept
EffectiveMaxTopInferersToReward(33, 5, 32) // => 32  ceiling clamps it
EffectiveMaxTopInferersToReward(0,  5, 32) // => 32  0 means "use the global maximum"
```

The **ceiling is applied last**, so the result can never exceed the global maximum even if the range is somehow inverted. That is what keeps the active set inside the global-sized score-history retention window in `scores.go`, which still uses the global value.

**Globals bound admission, not storage.** `Topic.Validate` intentionally does not check the cap against either global. Stored values are persisted verbatim — including a cap above the ceiling or below the floor, and including through genesis import — and the stored cap is re-resolved on every payload rather than trusted. This was the most consequential issue found in self-review: an earlier version enforced the ceiling at rest, and because `SetTopic` runs `Topic.Validate` on the per-epoch write paths, lowering the global silently froze epoch progression for every topic. Request-time validation still rejects a cap outside `[min, max]` (`ErrTopicMaxTopInferersToRewardTooBig`, and the new `ErrTopicMaxTopInferersToRewardTooSmall`, code 102).

**Well-formed global range.** `Params.Validate` rejects a floor above the ceiling, so governance cannot invert the range from either side. This is the module's first cross-param validation; it is safe here because params are written rarely, unlike the per-epoch `SetTopic` path.

**Guarded field.** Changing the cap is rejected while a worker submission window is open for any unfulfilled nonce, alongside `merit_sortition_alpha`, `max_labels_per_submission`, the label whitelist and the label default value.

**Hardened global ceiling validator.** The pre-existing `max_top_inferers_to_reward` validator now rejects `0`; the floor's validator deliberately accepts `0`, which simply means "no floor".

## Migration and upgrade

The emissions v16 migration backfills both new values, and two ordering decisions matter:

- **Topics before params.** `MigrateTopics` repairs a degenerate zero ceiling first, because persisting the floor requires a sane ceiling now that params validation rejects `floor > ceiling`. Wrong order would halt the upgrade, so it is pinned by a test.
- **No full topic re-validation.** The backfill sets only the new field and does not run `Topic.Validate`, so a dormant topic that violates some unrelated, param-dependent rule cannot halt the chain upgrade for a reason unrelated to this change.

The floor backfill clamps to the ceiling as defence in depth, so a chain whose ceiling sits below the default floor cannot fail validation mid-upgrade. No live chain is in that state (the ceiling is 32). Because existing topics are backfilled to the ceiling — the value they were already effectively using — installing the floor changes no admission outcome at the upgrade height.

Also wired up: the `v0.18.0` upgrade handler, a matching integration upgrade scenario (emissions 15 → 16, made the default upgrade target), and `Dockerfile.upgrade` bumped to `v0.17.0 → v0.18.0`.

- [x] **Tested.** Unit coverage for the resolution rule (including `0`, both bounds, an inverted range, and the ceiling-last ordering); request-time rejection above and below the range; `Params.Validate` on the range; storage kept verbatim at both ends through genesis; the window guard and its change detection; the v16 topic and param backfills, idempotency, the zero-ceiling repair, the ordering hazard, and that a topic invalid for an unrelated reason is still backfilled rather than halting; and a v15 test proving the cap survives that migration's rewrite path. Two end-to-end tests submit real worker payloads to show admission honouring the ceiling and the floor, each asserting the opposite outcome from the same fixture.
- [x] **Documented.** New `x/emissions/docs/global-parameters.md` and `x/emissions/docs/topic-parameters.md` (with mutability groups), README updates, the proto field comments, the `update-topic` help text, and `CHANGELOG.md`.
- [x] Added to `Unreleased` section of `CHANGELOG.md`.

## Still Left Todo

Nothing outstanding — all self-review comments are addressed and resolved.


x/emissions/types/validations.go
```diff
@@ -1073,6 +1073,22 @@ func (topic Topic) Validate(params Params) error {
...
+	//
+	// A zero value is intentionally NOT rejected here. Topics persisted before
+	// this field existed decode as zero, and the already-shipped v15 topic
+	// migration calls this method on the topics it backfills while they still
+	// have cap 0 (v16 has not run yet); rejecting zero here would make that
+	// shipped migration fail. The ">= 1 for live topics" invariant is instead
+	// enforced at every write path: the msgserver resolves a zero request to the
+	// global default, genesis import resolves zero to the global default, and the
+	// v16 migration backfills every existing topic to a positive value.
+	if topic.MaxTopInferersToReward > params.MaxTopInferersToReward {
```
xmariachi: ### [1] Per-topic ceiling in `Topic.Validate` silently freezes existing topics when the global is lowered

**Priority:** P1
**Severity:** High
**Summary:** This `topic.MaxTopInferersToReward > params.MaxTopInferersToReward` check runs on *every* `SetTopic`. A governance decrease of the global param then makes every topic whose stored cap now exceeds it fail validation on the per-epoch write paths, silently halting those topics' epoch progression and inference synthesis.
**PR alignment:** Introduced by this PR (the field and this check are new).

**Risks:** Network-wide loss of liveness for affected topics after a routine governance action; the failure is silent (only a debug/warn log), has a broad blast radius, and directly contradicts the stated design that the global "stays a live ceiling … without re-validating existing topics."

**Details:**
`SetTopic` calls `topic.Validate(params)` on every persist, and is reached from the per-epoch paths `UpdateTopicEpochLastEnded` and `UpdateTopicInitialRegret`. Because a create/update with `max_top_inferers_to_reward = 0` resolves to and stores the *concrete* global value (e.g. 32), and the v16 migration backfills existing topics to the global, essentially every topic ends up storing `cap == global`. Then:

1. Global = 32; topics stored with cap 32.
2. Governance lowers the global to 5 via `UpdateParams` (which merges + persists params and does **not** reconcile existing topics).
3. Every such topic now has stored cap 32 > global 5.
4. End-block churn: `UpdateTopicEpochLastEnded → SetTopic → Topic.Validate` fails; the reward loop logs a warning and `continue`s, skipping the worker-nonce add, so the topic never opens another epoch.
5. Inference synthesis: `UpdateTopicInitialRegret → SetTopic → Validate` returns the error, aborting the regret step for the topic.

The admission-time clamp `min(topic.cap, global)` already keeps the global a live ceiling for the only thing it needs to bound — how many inferers are admitted — and is explicitly designed to tolerate a lowered global. This `Validate` check is therefore redundant with the real enforcement and is the sole reason a lowered global bricks topics.

**Recommendation:** Remove the `> params.MaxTopInferersToReward` check from `Topic.Validate` and rely solely on the admission clamp for the ceiling; accept any stored value (including one above the current global) as a valid persisted state, since `min(cap, global)` makes it behave correctly. Add a regression test that lowers the global below a stored cap and asserts the topic still churns and still admits at most `global`.

If instead you want `cap ≤ global` to remain a hard at-rest invariant, `UpdateParams` must iterate and clamp every topic's stored cap when the global is lowered — heavier, and it turns a param change into an O(topics) state migration, so the admission-clamp approach is preferred. This decision is coupled with the genesis-normalization and centralization comments; please pick one answer and make validate/genesis/resolve all reflect it.

x/emissions/migrations/v16/migrate.go
```diff
@@ -0,0 +1,116 @@
...
+		// untouched so re-runs and already-migrated topics are no-ops.
+		if topic.MaxTopInferersToReward != 0 {
+			continue
+		}
+		topic.MaxTopInferersToReward = backfillValue
+
+		// Abort (do not skip) if the topic is invalid after backfill: a
+		// half-migrated topic would leave state inconsistent for post-v16 code.
+		// Halting is deterministic across validators.
+		if err := topic.Validate(params); err != nil {
```
xmariachi: ### [2] Migration aborts the entire upgrade if any pre-existing topic fails *full* `Topic.Validate`

**Priority:** P1
**Severity:** Medium
**Summary:** After backfilling the cap, the migration runs the whole `topic.Validate(params)` and returns on the first failure, halting the upgrade. The backfill itself can never fail this — but an unrelated, pre-existing field that is now out of range against current params can, turning a dormant topic into a network-wide upgrade halt.
**PR alignment:** Introduced by this PR (the halt-on-invalid choice; the iterate/validate pattern mirrors prior migrations).

**Risks:** All validators halt deterministically at the upgrade height (no fork, but a stuck chain requiring an emergency patched binary) for a reason unrelated to this feature.

**Details:**
`backfillValue == global`, so the new-field bound (`global > global`) can never trip. The only way `topic.Validate(params)` fails here is a cross-field/param-dependent rule on an *existing* field — e.g. `GroundTruthLag > MaxUnfulfilledReputerRequests * EpochLength` after governance lowered `MaxUnfulfilledReputerRequests`, or metadata/whitelist longer than a since-lowered limit. Active topics are re-validated every epoch (via `SetTopic`), so they currently pass; but **dormant/inactive topics are never re-validated in normal operation**, so a grandfathered-invalid one sits harmlessly until this migration promotes it to an upgrade-blocking error.

**Recommendation:** Scope the migration's enforcement to the invariant it actually establishes. Preferred — after backfilling, check/clamp only the new field and drop the full `Validate` call:
```go
if topic.MaxTopInferersToReward > params.MaxTopInferersToReward {
    topic.MaxTopInferersToReward = params.MaxTopInferersToReward
}
```
Alternatively, downgrade the failure to log-and-skip for non-cap validation errors while still writing the backfilled cap. Either removes the upgrade-halt vector while still guaranteeing every topic ends up with a valid cap. If you want to keep full validation as a canary, at minimum make the pre-upgrade rehearsal explicitly re-validate all existing (including dormant) topics against the target params, so a halt is caught before the upgrade rather than during it.

Question: is halt-on-any-invalid intentional here, or is clamp-only (safe, silent) acceptable? I'd recommend clamp-only.

app/upgrades/v0_18_0/upgrades.go
```diff
@@ -0,0 +1,41 @@
...
+	upgradetypes "cosmossdk.io/x/upgrade/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/module"
+
+	"github.com/allora-network/allora-chain/app/keepers"
+	"github.com/allora-network/allora-chain/app/upgrades"
+)
+
+const (
+	UpgradeName = "v0.18.0"
```
xmariachi, line 16: ### [3] This upgrade has no companion integration-test scenario, and the upgrade image still targets the previous release

**Priority:** P1
**Severity:** Medium
**Summary:** The end-to-end migration/upgrade path added here isn't exercised. The integration upgrade test is driven by a scenario registry keyed on upgrade name that has no entry for this upgrade (and its default target still points at the previous one), and the upgrade Docker image's baseline/target version env vars still reference the previous release.
**PR alignment:** Introduced by this PR (companion release wiring for this new upgrade).

**Risks:** The `15 → 16` emissions migration and this handler get no automated end-to-end coverage; a regression in the migration (e.g. the halt-on-invalid-topic risk, or any resolution/clamp bug) would not be caught, and the local migration-upgrade image would rehearse the wrong version pair.

**Details:**
- `test/integration/upgrade_scenarios_test.go` defines `UpgradeScenarios` (a map keyed by upgrade name) and `DefaultUpgradeTarget`; both currently stop at the previous release. There is no scenario asserting that the `emissions` module consensus version increases across this upgrade (it should reach 16), nor entries for the unchanged modules. `upgrade_test.go`'s `UpgradeChecks` reads these, so without a new entry this upgrade is never proposed/applied in the test.
- `Dockerfile.upgrade`'s `BASELINE_VERSION_TAG` / `UPGRADE_VERSION_TAG` still name the previous baseline→target pair, so the migration-upgrade image builds/tests the prior transition, not this one.

**Recommendation:** Add a scenario for this upgrade to the registry — `emissions` with an "expect increase" (expected version 16) and the other modules "expect equal" — and repoint `DefaultUpgradeTarget` at this upgrade's name so CI exercises it. Bump `Dockerfile.upgrade` so the baseline is the previous release and the target is this one, then run the local migration-upgrade rehearsal against realistic pre-upgrade state (the most reliable way to surface finding [2]). If these are deliberately handled at release-tagging time rather than in the feature PR, a one-line note on the PR would make that explicit.

x/emissions/keeper/msgserver/msg_server_topics.go
```diff
@@ -12,6 +12,26 @@ import (
 	"github.com/allora-network/allora-chain/x/emissions/types"
 )
 
+// resolveMaxTopInferersToReward applies the per-topic inferer-cap rules shared
+// by CreateNewTopic and UpdateTopic: a requested value of 0 means "use the
+// global default", and any concrete value must not exceed the global ceiling.
+// The returned value is always in [1, globalMax] (globalMax >= 1 is guaranteed
+// by params validation). Topic.Validate re-checks these bounds before
+// persistence; this early check exists only to return a friendly typed error.
+func resolveMaxTopInferersToReward(requested, globalMax uint64) (uint64, error) {
```
xmariachi: ### [4] Centralize the "0 → global, clamp to ceiling" rule — it exists in four places and admission diverges on `0`

**Priority:** P2
**Severity:** Medium
**Summary:** The same domain rule is hand-written four times: `resolveMaxTopInferersToReward` (reject `> global`), genesis (clamp), the v16 backfill (`0 → global`), and admission (`min`). The admission copy maps a stored `0 → 0` instead of `0 → global` — a silent divergence held together only by an invariant.
**PR alignment:** Introduced by this PR.

**Risks:** Four copies will drift (the zero-handling already has). If any write path ever leaves a `0` on a live topic, admission (`min(0, global) = 0`) admits nobody while every other consumer treats `0` as the global default — a subtle, hard-to-trace liveness bug.

**Details:**
The four expressions:
- resolve (here): `0 → global`, `> global → error`.
- genesis (`topic_genesis.go`): `if cap == 0 || cap > global { cap = global }` (silent).
- v16 backfill (`migrate.go`): `0 → global`; non-zero untouched.
- admission (`msg_server_worker_payload.go:137`): `min(topic.cap, global)` → maps `0 → 0`.

Everything except admission treats stored `0` as "use the global default"; admission alone would admit zero inferers on a `0` cap. That is safe *only* because three other paths guarantee live topics never store `0` — exactly the coupling that centralization removes. Separately, `resolveMaxTopInferersToReward` is unexported in `msgserver`, which is why genesis and the migration re-implemented the rule instead of reusing it; and its `> global → error` check duplicates the bound in `Topic.Validate` (reasonably kept for the friendlier typed error).

**Recommendation:** Add one exported helper in `types`, next to `Topic.Validate`:
```go
// EffectiveMaxTopInferersToReward resolves a stored/requested per-topic cap
// against the global: 0 means "use the global default", and the result is
// clamped to the global ceiling.
func EffectiveMaxTopInferersToReward(topicCap, global uint64) uint64 {
    if topicCap == 0 {
        topicCap = global
    }
    return min(topicCap, global)
}
```
Have genesis normalization, the v16 backfill, and the admission clamp all call it (uniform `0 → global` plus a single clamp), and layer `resolveMaxTopInferersToReward`'s stricter "reject if requested `> global`" typed error on top for the message path. This collapses four clamps to one source of truth and eliminates the admission `0 → 0` divergence. If the `Topic.Validate` ceiling check is removed per finding [1], this helper is what keeps the ceiling live everywhere.

xmariachi (commented):
## Review: per-topic `max_top_inferers_to_reward`

Thorough review of making `max_top_inferers_to_reward` a per-topic value (additive proto field + request resolution + admission clamp + v16 migration + genesis normalization). This is a careful, well-tested change and the state-format mechanics are fundamentally sound.

**Verified correct — no change needed:** additive proto field numbering (`Topic` #31, `CreateNewTopicRequest` #30, `UpdateTopicRequest` #12; no reuse/rename, `reserved` ranges intact); migration/upgrade wiring (`ConsensusVersion` 15→16, `RegisterMigration(15→16)`, new upgrade handler registered last, empty `StoreUpgrades` — correct as no new store key is introduced); genesis export/import parity for in-range topics; the score-retention invariant (`active set ≤ min(cap, global) ≤ global`-sized retention window in `scores.go`); determinism (no map-order/float/`time`/`rand` on consensus paths); creator-only authorization on `UpdateTopic`; migration idempotency and the deferred-write iteration pattern (matches the prior shipped migration); and error-code uniqueness (101).

**Main theme.** There is one substantive correctness issue plus several consistency/robustness items, most sharing a single root cause: **the "resolve `0` → global, clamp to the global ceiling" rule is expressed in four places with subtly different semantics, and the ceiling is additionally enforced inside `Topic.Validate`, which runs on every persist.** Fixing the factoring resolves several items at once. Nothing here is triggerable by an ordinary participant — the headline issue requires a governance parameter decrease — but it should be resolved before that parameter can ever be lowered on a live network.

Two operational gaps worth closing before release (companion upgrade-test scenario and upgrade-image version tags) are also called out inline.

### Issues (sorted by Priority desc, then Severity desc)

| # | Priority | Severity | Issue | PR alignment | Location |
|---|----------|----------|-------|--------------|----------|
| 1 | P1 | High | `Topic.Validate` per-topic ceiling silently freezes existing topics when the global is lowered | This PR | `x/emissions/types/validations.go:1089` |
| 2 | P1 | Medium | v16 migration aborts the whole upgrade if any pre-existing topic fails *full* validation | This PR (pattern mirrors prior migrations) | `x/emissions/migrations/v16/migrate.go:96` |
| 3 | P1 | Medium | No companion integration upgrade scenario for this upgrade; upgrade-image version tags stale | This PR | `test/integration/upgrade_scenarios_test.go`, `Dockerfile.upgrade` |
| 4 | P2 | Medium | "0 → global / clamp" rule duplicated across 4 paths; admission maps `0 → 0` (silent divergence) | This PR | `x/emissions/keeper/msgserver/msg_server_topics.go:21` (+ genesis, migrate, worker_payload) |
| 5 | P2 | Medium | Genesis import clamp is a lossy round-trip; adjacent zero-global fallback is dead & inconsistent | This PR | `x/emissions/keeper/topic_genesis.go:51` |
| 6 | P2 | Low | `UpdateTopic` treats request `0` as "reset to global default", not "leave unchanged" | This PR | `x/emissions/keeper/msgserver/msg_server_topics.go:180` |
| 7 | P2 | Low | New required CLI positional arg is backward-incompatible — make it `Optional` | This PR | `x/emissions/module/autocli.go:993` |
| 8 | P3 | Low | Near-duplicate `UpdateTopicRequest` test builders will drift | This PR | `x/emissions/keeper/msgserver/msg_server_topic_max_inferers_test.go:82` |
| 9 | P3 | Info | `validateMaxTopInferersToReward` now rejects `0` (behavior change; validator asymmetry) | This PR (tightens pre-existing param) | `x/emissions/types/params.go:611` |
| 10 | P3 | Info | Comment maintainability: trim rationale; reword version cross-reference | Mixed (new comment + pre-existing line) | `x/emissions/types/validations.go:1081` |

Detailed findings and fix proposals are inline. Findings 1, 4 and 5 are best addressed together via one shared resolution helper plus a single decision about whether `cap > global` is a supported at-rest state.

x/emissions/keeper/topic_genesis.go
```diff
@@ -22,12 +22,36 @@ func (k *TopicKeeper) InitGenesis(ctx context.Context, data *types.GenesisState)
...
+		globalMaxTopInferersToReward = types.DefaultParams().MaxTopInferersToReward
+	}
 	for _, topic := range data.Topics {
 		if topic != nil {
 			if topic.Topic == nil {
 				return errors.Wrap(types.ErrInvalidValue, "topic cannot be nil")
 			}
-			if err := k.SetTopic(ctx, topic.TopicId, *topic.Topic); err != nil {
+			resolvedTopic := *topic.Topic
+			if resolvedTopic.MaxTopInferersToReward == 0 || resolvedTopic.MaxTopInferersToReward > globalMaxTopInferersToReward {
```
xmariachi: ### [5] Genesis import silently clamps caps above the global (lossy round-trip), and the zero-global fallback just above is dead & self-inconsistent

**Priority:** P2
**Severity:** Medium
**Summary:** `InitGenesis` rewrites any topic with `cap > global` down to the global, so `export → import` is not identity for that (legitimately reachable) state; and the `globalMax == 0` fallback a few lines up can never fire and would be internally inconsistent if it did.
**PR alignment:** Introduced by this PR.

**Risks:** Silent data loss on genesis round-trips (state export/import, forks, snapshots) for topics whose frozen cap exceeds a since-lowered global; and a dead branch that, if ever reached, would set a value the immediately-following `SetTopic` validation would reject.

**Details:**
1. **Lossy clamp (this line):** `if resolvedTopic.MaxTopInferersToReward == 0 || … > globalMaxTopInferersToReward { = globalMaxTopInferersToReward }`. The `== 0` half (resolve unset → global) is correct and matches create/migration. The `> global` half silently rewrites a stored cap that a running chain can legitimately hold (after the global is lowered, topics keep their higher frozen cap — see finding [1]). Re-importing an exported genesis then changes those caps (e.g. 32 → 5), so the round-trip is not faithful. It is masked today only because admission clamps behavior anyway and finding [1] makes such topics non-functional — but the two are entangled: fix [1] to allow `cap > global` at rest and this clamp becomes an active data-loss bug. It is also a *third* policy for the same over-ceiling condition (resolve rejects, admission uses `min`, genesis clamps silently).
2. **Dead fallback (line ~43):** `if globalMaxTopInferersToReward == 0 { globalMaxTopInferersToReward = types.DefaultParams()… }`. Params are imported before topics and `SetParams`/`validateMaxTopInferersToReward` now reject a zero global, so this cannot trigger. Worse, if it did, it sets the topic to `DefaultParams()` (>0) while the imported `params` stays 0, so the following `SetTopic → Topic.Validate` ceiling check (`>0 > 0`) would fail.

**Recommendation:** Reduce this to the resolve-unset semantics only — normalize `cap == 0 → global` and drop the `> global` clamp (import the stored value as-is), consistent with removing the `Topic.Validate` ceiling in finding [1] — and route it through the shared helper from finding [4]. Delete the dead zero-global branch, or, if kept purely defensively, also repair the in-memory `params` so the subsequent `SetTopic` stays consistent.

Decision to make once, applied to genesis + validate + resolve: is `cap > global` a *supported* at-rest state (preferred, with `min` at admission) or a *forbidden* one? All three paths should reflect the same answer.

x/emissions/keeper/msgserver/msg_server_topics.go
```diff
@@ -135,13 +162,22 @@ func (ms msgServer) UpdateTopic(ctx context.Context, msg *types.UpdateTopicReque
...
 	updatedTopic.Metadata = msg.Metadata
 	updatedTopic.LossMethod = msg.LossMethod
 	updatedTopic.AlphaRegret = msg.AlphaRegret
 	updatedTopic.MeritSortitionAlpha = msg.MeritSortitionAlpha
 	updatedTopic.PNorm = msg.PNorm
 	updatedTopic.CNorm = msg.CNorm
+	// Per-topic inferer cap: full-replacement semantics with 0 -> global
+	// default. The keeper's UpdateTopic rejects this change while a worker
+	// submission window is open (see the guarded-fields set there).
+	updatedTopic.MaxTopInferersToReward = maxTopInferersToReward
```
xmariachi, line 183: ### [6] `UpdateTopic` treats a request cap of `0` as "reset to global default", not "leave unchanged"

**Priority:** P2
**Severity:** Low
**Summary:** `UpdateTopic` is full-replacement, and `0` is both the "use global default" sentinel and the protobuf wire-default for an omitted field. A client updating only (say) metadata without echoing the current cap will silently re-pin the cap to the current global — or have the whole update rejected mid-window.
**PR alignment:** Introduced by this PR.

**Risks:** Operator surprise — an explicitly-lowered per-topic cap silently jumps to the global on an unrelated edit; or, during an open worker submission window, the update is rejected with `ErrWorkerNonceWindowNotAvailable` because change-detection sees `storedCap != resolve(0)`.

**Details:**
`resolveMaxTopInferersToReward(0, global)` returns the global, and `updatedTopic.MaxTopInferersToReward` is then assigned as a full replacement. Concretely: topic created with cap 10 (global 32); an operator sends `UpdateTopicRequest{MaxTopInferersToReward: 0, …}` to change metadata → the cap becomes 32 (or the update is window-rejected because `10 != 32`). This is consistent with the module's existing full-replacement convention for `UpdateTopic` (metadata, loss method, whitelist, etc. all behave this way), so it is a footgun rather than a bug.

**Recommendation:** Keep the semantics (they match the rest of `UpdateTopic`) but make them unmissable. The autocli `Long` text already says "0 resets the cap to the global default" — mirror that in the `UpdateTopicRequest.max_top_inferers_to_reward` proto field doc so SDK/codegen consumers see it, and note that clients must echo the current value to preserve a sub-global cap. If preserving an explicit sub-global cap across unrelated edits is a real requirement, the alternative is a dedicated "unchanged" sentinel (optional/wrapper), but that's a larger change and probably not worth it given the convention. Is the full-replacement behavior intended here? If so, documenting is sufficient.

x/emissions/module/autocli.go
```diff
@@ -990,11 +990,12 @@ func (am AppModule) AutoCLIOptions() *autocliv1.ModuleOptions {
 						{ProtoField: "label_whitelist"},
 						{ProtoField: "label_default_value"},
 						{ProtoField: "label_case_sensitive"},
+						{ProtoField: "max_top_inferers_to_reward"},
```
xmariachi, line 1000: ### [7] New required CLI positional arg is backward-incompatible — mark it `Optional`

**Priority:** P2
**Severity:** Low
**Summary:** Appending `max_top_inferers_to_reward` as a positional makes `create-topic` (and `update-topic`, below) require one more positional than before; autocli positionals are required by default, so existing invocations/scripts break. Since the field is optional (`0` = default) and last, it can be `Optional`.
**PR alignment:** Introduced by this PR.

**Risks:** Every existing `allorad tx emissions create-topic …` / `update-topic …` call and script that passed the previous positional count now errors out.

**Details:**
`create-topic` goes from 25 to 26 required positionals, and `update-topic` from 11 to 12 (same change around line 1023). This matches how the earlier label fields were added, so it is consistent with repo style and the "prefer positional over flags" guidance — but those additions were also technically breaking. Because `max_top_inferers_to_reward` is semantically optional (`0` → global default) and is the last positional in both commands, marking it optional preserves backward compatibility without changing the positional style.

**Recommendation:** Mark the arg optional in both commands:
```go
{ProtoField: "max_top_inferers_to_reward", Optional: true},
```
Existing 25-arg / 11-arg invocations then keep working, with the field defaulting to `0` (= global default). If forcing callers to supply it was deliberate, that is worth a note; otherwise `Optional: true` is the low-risk choice.

x/emissions/keeper/msgserver/msg_server_topic_max_inferers_test.go
```diff
@@ -0,0 +1,292 @@
...
+		LabelWhitelist:         nil,
+		LabelDefaultValue:      alloraMath.ZeroDec(),
+	}
+}
+
+// baseWSWUpdateMsg builds an UpdateTopicRequest whose fields match the topic
+// created by createTopicForWSWTests, so only the overridden field changes.
+//
+//nolint:exhaustruct
+func (s *MsgServerTestSuite) baseWSWUpdateMsg(sender string, topicId uint64) *types.UpdateTopicRequest {
```
xmariachi, line 82: ### [8] Near-duplicate `UpdateTopicRequest` test builders will drift

**Priority:** P3
**Severity:** Low
**Summary:** `baseWSWUpdateMsg` and `baseCapUpdateMsg` (just above) are identical except for `Metadata`, `MaxLabelsPerSubmission`, and `LabelWhitelist`; the copy-paste will silently drift if `UpdateTopicRequest` gains a field.
**PR alignment:** Introduced by this PR.

**Risks:** Test-only — future maintenance cost, and a risk the two builders diverge in ways that weaken the "only the overridden field changes" guarantee each is meant to provide.

**Details:**
Each builder mirrors a different creation helper so an update changes only the field under test — a reasonable goal. But holding two full copies means a new required field must be added in both, and it is easy to update one and not the other.

**Recommendation:** Factor the shared fields into one builder parameterized by the base topic (or by just the differing label fields), so the "matches the created topic" invariant lives in one place. Minor; safe to defer.

x/emissions/types/params.go
```diff
@@ -604,9 +604,14 @@ func validateMaxSamplesToScaleScores(i uint64) error {
...
 }
 
-// max this many top workers by score are rewarded for a topic
-// Should be zero or positive. Enforced by uint type
-func validateMaxTopInferersToReward(_ uint64) error {
+// max this many top inferers by score are admitted and rewarded for a topic.
+// Must be greater than zero: this global value is also the ceiling and default
+// for the per-topic Topic.MaxTopInferersToReward, so a zero global would make
+// the per-topic ceiling zero and reject every topic.
+func validateMaxTopInferersToReward(i uint64) error {
```
xmariachi, line 621: ### [9] `validateMaxTopInferersToReward` now rejects `0` — a governance behavior change, and the sibling "top-N" validators remain unchecked

**Priority:** P3
**Severity:** Info
**Summary:** This validator previously accepted any value and now rejects `0` (correct — the global is now the per-topic default/ceiling), so a governance `UpdateParams` setting it to `0`, previously allowed, is now rejected. The analogous forecaster/reputer cap validators remain no-ops, making the three asymmetric.
**PR alignment:** Introduced by this PR (tightens a pre-existing param's validation).

**Risks:** Minimal and intended; noted for release awareness (a value that used to be accepted is now rejected) and for consistency.

**Details:**
The tightening is correct: a zero global would make the per-topic ceiling zero and reject every topic, and the v16 migration's zero-global self-repair plus a `params_test.go` case cover the upgrade path. Two awareness notes: (a) `validateMaxTopForecastersToReward` and `validateMaxTopReputersToReward` are still no-op validators, so the three "top-N" caps now enforce different things; (b) external tooling or a gov proposal that relied on setting this to `0` will now fail validation.

**Recommendation:** No change required to the zero-rejection. Optionally add a one-line comment that the asymmetry with the forecaster/reputer validators is intentional (the inferer cap alone is the per-topic ceiling/default), so a future reader doesn't "harmonize" it by loosening this one.

x/emissions/types/validations.go
```diff
@@ -1073,6 +1073,22 @@ func (topic Topic) Validate(params Params) error {
 	if topic.CNorm.Lt(validationCNormMinDec) || topic.CNorm.Gt(validationCNormMaxDec) {
 		return errors.Wrap(sdkerrors.ErrInvalidType, "topic c_norm must be between -100 and 100")
 	}
+	// The per-topic inferer cap may not exceed the global param, which is both
+	// the default (applied at creation when a topic leaves it unset) and the
+	// ceiling: a larger value would let the active set exceed the global-sized
+	// score-retention window in scores.go.
+	//
+	// A zero value is intentionally NOT rejected here. Topics persisted before
```
xmariachi: ### [10] Comment maintainability: trim the zero-accept rationale and reword the version cross-reference

**Priority:** P3
**Severity:** Info
**Summary:** The zero-accept rationale here is self-contained but long (9 lines) and duplicated in two other places; and an adjacent `UpdateTopic` doc block this PR edits (in `keeper/topic.go`) contrasts behavior against "the v14 behavior," a version cross-reference that will keep aging.
**PR alignment:** Mixed — the long comment is new in this PR; the "v14 behavior" line is pre-existing but sits in a block this PR modifies.

**Risks:** None functional; maintainability only. Per the repo's comment guidance, comments should describe behavior self-containedly rather than lean on version/history references that go stale.

**Details:**
- Here (`validations.go`): the rationale for accepting `0` is correct and self-contained, but the same reasoning also appears in `params.go` and in the tests, and the 9-line form is heavier than needed.
- In `keeper/topic.go` (the `UpdateTopic` doc block this PR edits to add the `max_top_inferers_to_reward` bullet), a neighboring sentence contrasts with "the v14 behavior." Since the block is already being touched, it is a cheap moment to reword.

**Recommendation:** Trim the zero-accept comment to the essential invariant — e.g. *"zero is accepted for migration safety; the ≥1 guarantee for live topics is enforced at every write path (message resolution, genesis import, and the v16 backfill)."* And reword the "v14 behavior" reference to describe the behavior directly — e.g. *"checks every unfulfilled nonce within its submission window, not just the newest."* Both are optional polish, grouped here so they can be handled in one pass.

x/emissions/migrations/v16/migrate.go
xmariachi: Addressed in 9225895a. The migration no longer runs full `Topic.Validate(params)` after backfill — the backfilled cap equals the global and is valid by construction (`>= 1`, `<= global`), so unrelated pre-existing field invalidity can no longer halt the upgrade. The former halt test was reworked into `TestMigrateTopicsBackfillsDespiteUnrelatedInvalidTopic`, which seeds a topic invalid for a non-cap reason (ground-truth-lag `<` epoch-length) with cap `0` and asserts the migration succeeds and backfills the cap to the global.

app/upgrades/v0_18_0/upgrades.go
xmariachi: Addressed in 9225895a. Added a `v0.18.0` integration upgrade scenario (emissions `ExpectIncrease` → 16, scheduler/mint `ExpectEqual`) and repointed `DefaultUpgradeTarget` to it, and bumped `Dockerfile.upgrade` to baseline `v0.17.0` → target `v0.18.0`.

x/emissions/types/validations.go
xmariachi: Addressed in `9e584d2f`. `Topic.Validate` no longer bounds the per-topic cap against the global: the global is an admission-only ceiling, applied via `EffectiveMaxTopInferersToReward` (`0` → global, otherwise `min(cap, global)`). Lowering the global therefore no longer fails the per-epoch `SetTopic` path (`UpdateTopicEpochLastEnded` / initial regret), so topics are not silently frozen. Regression test added: `TestSetTopicSucceedsWhenStoredCapExceedsLoweredGlobal`.

x/emissions/keeper/msgserver/msg_server_topics.go
xmariachi: Addressed in `9e584d2f`. The resolution rule now lives in one place, `types.EffectiveMaxTopInferersToReward` (`0` → global, clamped to the global), and is used by both the admission path and `resolveMaxTopInferersToReward`. This removes the divergence where admission mapped a stored `0` to `0` (admitting nobody) while every other path treated `0` as "use the global". Unit test added for the helper.

x/emissions/keeper/topic_genesis.go
xmariachi: Addressed in `9e584d2f`. `InitGenesis` now stores the imported topic verbatim: no clamp of above-global caps and no zero-global fallback, so `export → import` round-trips faithfully. Both `0` and above-global stored values are resolved at admission instead. Tests updated to assert preservation (`TestInitGenesisPreservesZeroMaxTopInferersToReward`, `TestInitGenesisPreservesMaxTopInferersAboveGlobal`).

x/emissions/types/params.go
xmariachi: Addressed in `55c52c7c`, though not quite as suggested. The zero-rejection is unchanged (intended). Rather than adding the sibling-asymmetry note, this comment needed a correctness fix: it claimed the global is the ceiling "so a zero global would make the per-topic ceiling zero and reject every topic", which stopped being true once `Topic.Validate` dropped the at-rest ceiling. It now reads:

```go
// Must be greater than zero: it is the admission ceiling and the default for an
// unset per-topic Topic.MaxTopInferersToReward, so zero would admit nobody.
```

The note about `validateMaxTopForecastersToReward`/`validateMaxTopReputersToReward` staying unbounded was deliberately left out — three extra lines for a marginal detail. Happy to add a short version if you would rather have it explicit.

x/emissions/types/validations.go
xmariachi: Addressed across `9e584d2f` (the `validations.go` half) and `55c52c7c` (the version references).

- `validations.go`: the zero-accept rationale is now three lines stating the invariant (the global bounds admission, not storage; zero and above-global stored values are valid at rest).
- `topic.go`: both version references are gone — the one on `UpdateTopic` and a second one on `isAnyUnfulfilledWorkerNonceWithinWindow` that was outside the original scope. Rather than rewording them in place (which duplicated the same sentence in two comments), the behaviour is now described once, on the function that implements it: `isAnyUnfulfilledWorkerNonceWithinWindow` owns "every unfulfilled nonce is checked, not just the newest, because workers can submit against any currently-open nonce", and `UpdateTopic` keeps only the list of guarded fields, next to the code that checks them. Its trailing version-contrast paragraph was dropped entirely.
- Also removed the stale field enumeration on the helper: it listed the guarded fields but had not been updated with `max_top_inferers_to_reward`, and it duplicated the authoritative list on `UpdateTopic`.

A sweep confirms no version-referencing comments remain under `x/` outside `migrations/` and tests, where they describe specific migrations and are meaningful. This keeps the comments stable across module-version bumps.

x/emissions/module/autocli.go
xmariachi: Decision: the positional stays **required** — no code change.

Confirming the mechanics first, since they matter for the trade-off: autocli positionals are required by default (`cobra.ExactArgs(N)`), and `Optional` widens that to `cobra.RangeArgs(N-1, N)` but is only accepted on the **last** positional. So `create-topic` goes from 25 to 26 required args and `update-topic` from 11 to 12, and marking this field optional would indeed have restored backward compatibility.

We are keeping it required anyway, for two reasons:

1. **Consistency.** None of the other positionals on these commands is `Optional`, including ones that have a sentinel or a natural default (`unity_tolerance`, `label_whitelist`, `label_default_value`). Making this single field special would be an asymmetry in the command's UX for no functional gain — the field is a normal topic parameter and reads better as one.
2. **It would pin the field in place.** `Optional` is only valid on the last positional, so adopting it here permanently constrains where future fields can be added — the next parameter would have to go before it, or the optionality would have to be dropped again.

This is a breaking CLI change and we are taking it deliberately. It matches how the label parameters (`max_labels_per_submission`, `label_whitelist`, `label_default_value`, `label_case_sensitive`) were added, so it is the established convention for this command rather than a new regression. Worth calling out in the release notes for anyone with scripted invocations.

x/emissions/keeper/msgserver/msg_server_topic_max_inferers_test.go
xmariachi: Decision: **not doing this** — the two builders stay as they are.

The refactor was written and then dropped, because it came out net negative:

- It turns 2 functions into 3 and saves roughly 6 lines, at the cost of a permanent extra layer of indirection.
- The shared helper needs 5 positional parameters (`sender, topicId, metadata, maxLabelsPerSubmission, labelWhitelist`), so call sites read `baseUpdateMsg(sender, id, "wsw test", 4, []string{"a","b","c"})` — the `4` and the `nil` carry no meaning at the call site. The current flat builders show the whole `UpdateTopicRequest` at a glance.
- **The drift risk it was meant to prevent does not exist**: both builders carry `//nolint:exhaustruct`, so a new field on `UpdateTopicRequest` would not be flagged in either shape — it would simply default to its zero value. Deduplicating moves where that zero lives; it adds no guarantee.

The duplication here is literal, visible and benign, which is the case where flat beats factored. If these builders later grow more differences, or a third one appears, the right move would be an options struct rather than more positional parameters — happy to revisit then.

x/emissions/keeper/msgserver/msg_server_topics.go
xmariachi: Follow-up merged in #969, which reworked this area: the cap is now resolved into a validated global range `[min_top_inferers_to_reward, max_top_inferers_to_reward]`. The floor is a new global param, active by default at `5`; requests outside the range are rejected, and the effective cap is clamped at admission with the ceiling applied last. Stored caps stay verbatim at both ends.

The `0` semantics this thread is about are deliberately unchanged: it still means "use the global default", i.e. full replacement, consistent with every other field on `UpdateTopic`. It is documented on the proto field and in the `update-topic` help text, so clients must echo the current value to preserve a cap below the default. Details in #969.

x/emissions/proto/emissions/v9/params.proto
```diff
@@ -189,4 +189,8 @@ message Params {
   // Max absolute number of labels allowed in an epoch label registry
   // for a single (topic, nonce).
   uint64 max_epoch_label_registry_size = 63;
+  // Floor for the per-topic inferer admission cap. Together with
+  // max_top_inferers_to_reward it is the range a topic may request, and the
+  // range the effective cap is clamped into. Zero means no floor.
+  uint64 min_top_inferers_to_reward = 64;
```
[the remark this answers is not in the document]
xmariachi: Not reproducing — the generated bindings are checked in, and have been since the #969 merge commit `b04ce647`.

At the current branch head, `MinTopInferersToReward` appears in `x/emissions/types/params.pb.go` and in `OptionalParams` in `x/emissions/types/tx.pb.go`, and `min_top_inferers_to_reward` in `x/emissions/api/emissions/v9/params.pulsar.go`. Field 64 round-trips; nothing is discarded as unknown. `TestUpdateAllParams` and the v16 migration backfill tests both exercise the field end-to-end and would fail if it were being dropped.

This most likely ran against a tree before `b04ce647` landed. For the record the suggested remedy is the right one — `make -C x/emissions proto-gen` (Docker, via `scripts/protocgen.sh`) is the supported path; a bare `buf generate` over `x/emissions/proto` is not, since all versions share one `go_package` and buf silently keeps the oldest colliding file.

Resolving as stale.

x/emissions/keeper/msgserver/msg_server_topics.go
xmariachi: Closing the loop on the recommendation in this thread: the semantics are now on the proto fields, so codegen/SDK consumers see them without reading the CLI help. Done in `8fb23e40`.

- `CreateNewTopicRequest.max_top_inferers_to_reward` and `UpdateTopicRequest.max_top_inferers_to_reward` both state that `0` means "use the global `max_top_inferers_to_reward`", that the value is resolved and stored (so the topic keeps it if the global later moves), and that any other value must sit within `[min_top_inferers_to_reward, max_top_inferers_to_reward]`. The `UpdateTopic` one also spells out full replacement and the open-window rejection.
- `Topic.max_top_inferers_to_reward` no longer claims the stored value is always non-zero — that was untrue for a genesis-supplied `0` — and now distinguishes the msg path from genesis.
- `update-topic` autocli help says clients must echo the current value to keep a cap below the default; `create-topic` gained a `Long` explaining the sentinel, having had none.
- Added `x/emissions/docs/topic-parameters.md` and `x/emissions/docs/global-parameters.md`, linked from both READMEs, with the full-replacement caveat and the consequences of a floor raise.

Semantics are deliberately unchanged: `0` still means "use the global default", consistent with every other field on `UpdateTopic`. Resolving as documented-by-design.

x/emissions/module/autocli.go
```diff
@@ -962,8 +962,15 @@ func (am AppModule) AutoCLIOptions() *autocliv1.ModuleOptions {
...
+					Use:       "create-topic [creator] [metadata] [loss_method] [epoch_length] [ground_truth_lag] [worker_submission_window] [p_norm] [alpha_regret] [allow_negative] [epsilon] [merit_sortition_alpha] [active_inferer_quantile] [active_forecaster_quantile] [active_reputer_quantile] [enable_worker_whitelist] [enable_reputer_whitelist] [c_norm] [topic_type] [output_arity] [require_unity] [unity_tolerance] [max_labels_per_submission] [label_whitelist] [label_default_value] [label_case_sensitive] [max_top_inferers_to_reward]",
 					Short:     "Add a new topic to the network",
+					Long: `Add a new topic to the network.
+
+max_top_inferers_to_reward caps how many inferers are admitted to the topic's
+active set. Pass 0 to use the global max_top_inferers_to_reward: the current
+global value is resolved and stored. The topic keeps that number even if the
+global changes later - but is effectively clamped to the global range.
```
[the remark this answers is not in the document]
xmariachi: Valid, fixed in `a3d0d90a` — trailing space stripped from the `CreateNewTopic` `Long` raw string in `x/emissions/module/autocli.go`. No trailing whitespace remains anywhere in the file.

x/emissions/migrations/v15/migrate_test.go
```diff
@@ -126,6 +126,7 @@ func (s *EmissionsV15MigrationTestSuite) TestMigrateTopicsPreservesExistingClass
 	topicStore := prefix.NewStore(store, emissionstypes.TopicsKey)
 
 	classifTopic := emissionstypes.Topic{
+		MaxTopInferersToReward:   8,
```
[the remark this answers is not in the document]
xmariachi: Valid, fixed in `fce9cd8a`.

The one-line assertion this comment implies would not have caught anything, which is worth recording. `MigrateTopics` has `if !changed { continue }`, and the classification topic here is fully populated, so v15 skips it outright — asserting its cap only proves the migration left alone a topic it never opened. Every topic v15 *does* rewrite in this file is a `v14oldtypes.Topic`, which has no cap field at all, so nothing here could show a cap surviving a rewrite.

So the fix seeds a second topic (id 43) with an unspecified type and arity, which puts it on the changed-and-revalidated path, plus a coherent schedule so revalidation passes, and a cap of `7`. The new assertions check that the defaults were applied — proving it went through the rewrite — and that the cap came out intact. The original topic keeps its assertion too, covering the skip path.

Mutation-checked: dropping the cap inside `MigrateTopics` fails with `expected: 0x7, actual: 0x0`, isolated to the new assertion.

Worth being explicit that v15 already shipped in `v0.17.0`, so this is not about the past migration. It guards the present coupling: `v15/migrate.go` decodes with the current `emissionstypes.Topic`, so every field added to `Topic` later — like this one in #968 — starts flowing through that round-trip, and a regression there would only surface in chained-migration tests and snapshot upgrade rehearsals.

xmariachi:
All comments addressed up to here.

zale144 (approved):
lgtm

PR #980, opened 2026-08-12
fix(deps): patch iavl fast-node cache/commit race causing AppHash divergence

helder-moreira:

Patches a fast-node cache/commit race in `iavl` that caused **`AppHash` divergence** — nodes forking with `wrong Block.Header.AppHash` and only recoverable via rollback/resync.

Root cause is in `iavl`, not our code:
- The ecosystem treasury account is emptied to **zero every block**, and x/bank **deletes** a key at zero balance.
- In `SaveVersion`, `DeleteFastNode` dropped the key from the fast-node cache **immediately** but only *buffered* the on-disk delete until `Commit()`.
- A concurrent point read (an LCD balance query → `GetFastNode` sharing the same nodeDB) in that window missed the cache, reloaded the **stale pre-delete row** from disk, and re-cached it.
- The next block read the stale balance instead of `0`, minted the wrong amount, and forked.

Reproduced live on a testnet node: read balance **40 instead of 0**, minted 40 uallo short. Same signature seen at 50 and 290 uallo on other nodes.

**The change is `go.mod`/`go.sum` only** — no allora-chain code. It points `iavl` at [`allora-network/iavl`](https://github.com/allora-network/iavl) (`v1.2.6` + backport of upstream [cosmos/iavl#1142](https://github.com/cosmos/iavl/pull/1142), "fix race between updating fast node cache and db commit"). #1142 is released only in iavl `v1.2.7+`, which is API-incompatible with our `cosmossdk.io/store v1.1.1`, so it was backported onto `v1.2.6` — the newest tag that still builds with our stack.

**Rollout / safety:** not consensus-breaking. A node that never hit the race computes the identical app hash with or without this change (the fast cache is a read-accelerator over the authoritative tree); only buggy nodes change — they stop forking. No migration, no upgrade handler, no height coordination — dependency bump + rebuild + rolling restart; mixed patched/unpatched fleet is safe during rollout. Already-forked nodes still need a rollback/resync to recover.

**Immediate mitigation, no deploy:** the trigger is a service polling the ecosystem account balance over LCD. Caching or blocking `GET /cosmos/bank/v1beta1/balances/<ecosystem addr>/by_denom` at the gateway removes the trigger while this rolls out.

_None (incident-driven). Upstream: cosmos/iavl#1142._

- [x] If tested, please describe how. A deterministic repro test on the fork (`fastnode_poison_test.go`) reproduces the stale read: **fails on unpatched v1.2.6, passes with the fix**, plus two controls (fastnode disabled; no concurrent read). iavl full suite has no regressions vs pristine v1.2.6 (the only local failures are a macOS-only `legacydump` helper-exec issue, identical with/without the patch — pending a Linux/CI run). `allora-chain` builds (`allorad` binary) and `x/mint` tests pass against the fork. Also confirmed live by reproducing the fork on a real testnet node and then verifying the mechanism.
- [x] If documented, please describe where. `CHANGELOG.md` `Unreleased` → `Security`.
- [x] Added to `Unreleased` section of `CHANGELOG.md`? Yes, under `Security` (chain-continuity bug).

## Still Left Todo

- Linux/CI run of the fork's full iavl suite to confirm green (blocked locally by the macOS `legacydump` helper).
- Independent review of the backport is in progress.
- After merge to `dev`: cut the `v0.17` patch release via the `candidate-*` flow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security
 
+* [#980](https://github.com/allora-network/allora-chain/pull/980) Patch a fast-node cache/commit race in `iavl` that caused `AppHash` divergence: the ecosystem treasury account is deleted every block at zero balance, and a concurrent balance query during commit could read its value one block stale, minting the wrong amount and forking the node. Points `iavl` at `github.com/allora-network/iavl` (`v1.2.6` + backport of [cosmos/iavl#1142](https://github.com/cosmos/iavl/pull/1142)). Not consensus-breaking; no migration required.
+
 ### API Breaking Changes
 
 #### Removed
--- a/go.mod
+++ b/go.mod
@@ -96,7 +96,7 @@ require (
 	github.com/cosmos/btcutil v1.0.5 // indirect
 	github.com/cosmos/go-bip39 v1.0.0 // indirect
 	github.com/cosmos/gogogateway v1.2.0 // indirect
-	github.com/cosmos/iavl v1.2.2 // indirect
+	github.com/cosmos/iavl v1.2.6 // indirect
 	github.com/cosmos/ics23/go v0.11.0 // indirect
 	github.com/cosmos/ledger-cosmos-go v0.14.0 // indirect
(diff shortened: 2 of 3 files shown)


xmariachi (dismissed):
lgtm

xmariachi (approved):

PR #985, opened 2026-08-21
Register historical nested pulsar types for tx queries

zale144:
## Summary
- After the v10 proto bump, `/cosmos/tx/v1beta1/txs` failed to decode pre-upgrade payloads: `unknownproto` looks up nested (non-Any) fields in the gogo registry, but `x/emissions/types` now registers those names as `emissions.v10.*`.
- Register the existing pulsar nested types under their historical names (emissions v2–v9 worker/reputer/params graphs, plus `mint.v1beta1.Params`). These are distinct Go types from the live gogo structs, so new txs still encode as v10 / mint v5.
- Based on `v0.17.1` (`release-v0.17.1`).

## Test plan
- [x] `go test ./x/emissions/api/emissions/v2/ ./x/emissions/api/emissions/v3/ ./x/emissions/api/emissions/v9/ ./x/mint/api/mint/v1beta1/`
- [x] Local in-place-testnet from a mainnet snapshot: `GET /cosmos/tx/v1beta1/txs?query=tx.height=...` returns worker txs instead of the parse error

### DIFF
--- a/x/emissions/api/emissions/v2/codec.go
+++ b/x/emissions/api/emissions/v2/codec.go
@@ -4,7 +4,34 @@ import (
 	"github.com/cosmos/cosmos-sdk/codec"
 	"github.com/cosmos/cosmos-sdk/codec/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	gogoproto "github.com/cosmos/gogoproto/proto"
 )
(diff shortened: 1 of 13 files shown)


zale144:
closing in favor of #986

PR #986, opened 2026-08-24
ENGN-9226 Diego/query tx separation decoding path

xmariachi:
## Problem

After the v10 proto bump, transaction query endpoints (e.g. `GET /cosmos/tx/v1beta1/txs`) can no longer decode pre-upgrade payloads. The strict unknown-field check resolves nested (non-`Any`) message types by name through the global gogo proto registry, and several historical names are now registered as their v10 equivalents. As a result the whole transaction fails to decode and the query returns a parse error instead of the stored tx.

## Constraints (must not affect consensus)

The transaction decoder is shared between the query path and the consensus path (`CheckTx`, `ProcessProposal`, `FinalizeBlock`). Any change that makes a node accept a transaction it previously rejected is therefore consensus-relevant: it changes the set of transactions a node will admit and partially process, which can diverge application state between nodes running different binaries during a rolling upgrade. 

Hard requirements for the fix:
- Do **not** modify the consensus decoder or its behavior.
- Do **not** mutate the app's interface registry or the process-global proto registry.
- The set of transactions accepted by consensus must be byte-for-byte identical before and after this change.

## Solution

Introduce a read-only decoding path used **only** by the historical-read methods of the tx service (`GetTx`, `GetTxsEvent`, `GetBlockWithTxs`):

- A **separate interface registry**, mirroring the app's registry plus the historical `mint.v2` types, built without touching the app's own registry.
- A **tolerant decoder** that tries the standard strict decoder first (so current transactions keep the identical code path and type) and only falls back to a plain unmarshal — which skips the strict unknown-field walk — for payloads the strict decoder rejects.
- A **composite tx service** that routes only the three read methods to the tolerant backend; every other method (including `TxDecode`) keeps the app's standard strict behavior.

The consensus decoder and the app's registry are left completely unchanged.

## Details

- **`app/querytx.go`** (new): the separate registry builder, the tolerant decoder plus a minimal `sdk.Tx` shim for decoded historical txs, the composite tx service, and an override of `RegisterTxService` that wires it in.
- **`app/app.go`**: small, behavior-preserving wiring — hoist the custom module basics into a reusable helper, and build/store the read-only registry and tx config at startup.
- **Tests**:
  - Historical emissions worker payloads (v2–v9) decode on the query path, both standalone and when nested inside another message.
  - Guard tests asserting the consensus decoder still rejects the same payloads, and that the added `mint.v2` types resolve only on the query registry — never on the consensus registry or the global proto registry.
  - An end-to-end test that delivers such a payload through a real block and asserts consensus state (balances, account sequence, gas) is unchanged.
- **Known limitation (separate follow-up)**: query-only historical types (`mint.v2` and the v7/v8 whitelist txs) decode over gRPC but don't render over the REST gateway, which resolves messages via the app registry; a page containing one fails to render. All other historical versions are unaffected.

---
## Summary by cubic
Restores decoding of pre‑v10 transactions on tx query endpoints without changing what consensus accepts. Previously these queries failed to parse stored pre‑v10 payloads; now `GetTx`, `GetTxsEvent`, and `GetBlockWithTxs` decode via a read‑only path while consensus decoding stays strict.

- Builds a read‑only `InterfaceRegistry` for queries that mirrors the app registry and adds historical `mint.v2` plus missing v7/v8 whitelist messages.
- Uses a tolerant query‑only decoder that tries strict first, then falls back; rejects message `Any` with empty or unresolvable type URLs.
- Routes only the three gRPC read methods to the tolerant backend; all other tx service methods (including `TxDecode`) remain strict.
- Keeps consensus behavior unchanged and does not mutate the app registry or the global registry; no migrations or rollout steps.
- Adds tests for v2–v9 decoding (bare and `authz`‑wrapped), consensus rejection, registry superset checks, REST gateway behavior (only consensus‑registered types render), and a FinalizeBlock harness that proves no state changes for wrapped v9.
- Known limitation: REST JSON cannot render `mint.v2`; gRPC queries are unaffected.

**Dependencies**
- Pins `golangci-lint` workflows to Go 1.23.5 and updates the `go-install-hardened` workflow reference.

<sup>Written for commit c0a6015f85caaa33cbe3be79fa3e7d34c80063be. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/allora-network/allora-chain/pull/986?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/querytx.go
+++ b/app/querytx.go
@@ -1,0 +1,205 @@
+package app
+
+import (
+	"context"
+	"fmt"
+
+	txsigning "cosmossdk.io/x/tx/signing"
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/codec"
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
+	"github.com/cosmos/cosmos-sdk/std"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/cosmos/cosmos-sdk/types/module"
+	txtypes "github.com/cosmos/cosmos-sdk/types/tx"
+	authtx "github.com/cosmos/cosmos-sdk/x/auth/tx"
+	gogoproto "github.com/cosmos/gogoproto/proto"
+	protov2 "google.golang.org/protobuf/proto"
+
+	mintv2 "github.com/allora-network/allora-chain/x/mint/api/mint/v2"
+)
+
+// Compile-time guards for the two exported contracts these types must satisfy.
+// The read path also asserts intoAny (AsAny) and protoTxProvider (GetProtoTx) on
+// historicalTx structurally; those are covered by TestHistoricalTxImplementsReadInterfaces.
+var (
+	_ sdk.Tx                = historicalTx{}      //nolint:exhaustruct // interface assertion only
+	_ txtypes.ServiceServer = compositeTxServer{} //nolint:exhaustruct // interface assertion only
+)
+
+// The consensus tx decoder (app.txConfig) rejects pre-upgrade payloads: after the
+// v10 proto bump, unknownproto can no longer resolve historical nested message
+// names in the process-global gogo registry. Widening the consensus decoder to
+// accept them would be consensus-breaking (a rejected tx and an accepted-then-failed
+// tx write different state), so instead the historical-read RPCs get their own
+// tolerant decoder over a separate registry. Nothing here touches consensus state.
+
+// historicalTx is the minimal sdk.Tx the tolerant decoder returns for a payload
+// the strict decoder rejects. It exposes exactly the three shapes the read-only
+// tx service asserts: sdk.Tx, intoAny (AsAny), and protoTxProvider (GetProtoTx).
+type historicalTx struct {
+	tx *txtypes.Tx
+}
+
+// GetMsgs returns the already-unpacked messages; the tolerant decoder unpacks the
+// Anys during Unmarshal, so the cached values are present.
+func (h historicalTx) GetMsgs() []sdk.Msg { return h.tx.GetMsgs() }
+
+// GetMsgsV2 is never called on the read path; fail loudly rather than return a
+// half-built value if that ever changes.
+func (h historicalTx) GetMsgsV2() ([]protov2.Message, error) {
+	return nil, fmt.Errorf("historicalTx: GetMsgsV2 is not supported on the read-only query path")
+}
+
+// GetProtoTx satisfies protoTxProvider, used by GetBlockWithTxs.
+func (h historicalTx) GetProtoTx() *txtypes.Tx { return h.tx }
+
+// AsAny satisfies intoAny, used by GetTx/GetTxsEvent. UnsafePackAny (not
+// NewAnyWithValue) is required so the cached value is the concrete *txtypes.Tx
+// that mkTxResult then type-asserts.
+func (h historicalTx) AsAny() *codectypes.Any { return codectypes.UnsafePackAny(h.tx) }
+
+// newHistoricalTxDecoder decodes committed txs for the query path. It tries the
+// strict SDK decoder first, so every tx that decodes today keeps the identical
+// code path and concrete type; only when that fails (a historical payload) does
+// it fall back to a plain unmarshal that skips unknownproto's nested-field walk.
+func newHistoricalTxDecoder(cdc codec.Codec) sdk.TxDecoder {
+	strict := authtx.DefaultTxDecoder(cdc)
+	return func(txBytes []byte) (sdk.Tx, error) {
+		// Current-format txs: unchanged behavior and unchanged return type.
+		if tx, err := strict(txBytes); err == nil {
+			return tx, nil
+		}
+
+		// Historical payloads: unmarshal the envelope directly. Any resolution
+		// still goes through cdc's registry, so an unknown type URL still errors
+		// here; only the strict unknown-field rejection is skipped.
+		var raw txtypes.TxRaw
+		if err := cdc.Unmarshal(txBytes, &raw); err != nil {
+			return nil, sdkerrors.ErrTxDecode.Wrap(err.Error())
+		}
+		var body txtypes.TxBody
+		if err := cdc.Unmarshal(raw.BodyBytes, &body); err != nil {
+			return nil, sdkerrors.ErrTxDecode.Wrap(err.Error())
+		}
+		var authInfo txtypes.AuthInfo
+		if err := cdc.Unmarshal(raw.AuthInfoBytes, &authInfo); err != nil {
+			return nil, sdkerrors.ErrTxDecode.Wrap(err.Error())
+		}
+		return historicalTx{tx: &txtypes.Tx{
+			Body:       &body,
+			AuthInfo:   &authInfo,
+			Signatures: raw.Signatures,
+		}}, nil
+	}
+}
+
+// buildQueryDecodingPath constructs the read-only registry and tx config and
+// stores them on the app. Called once at construction, after every module is on
+// ModuleManager. It never mutates app.interfaceRegistry or the gogo registry.
+func (app *AlloraApp) buildQueryDecodingPath() error {
+	reg, err := app.buildQueryInterfaceRegistry()
+	if err != nil {
+		return err
+	}
+	app.queryInterfaceRegistry = reg
+
+	// The tolerant decoder and the tx config share one codec so Any resolution is
+	// consistent across decode and (gRPC) response marshaling.
+	queryCodec := codec.NewProtoCodec(reg)
+	//nolint:exhaustruct // only the decoder and sign modes are overridden; the rest default
+	txConfig, err := authtx.NewTxConfigWithOptions(queryCodec, authtx.ConfigOptions{
+		EnabledSignModes: authtx.DefaultSignModes,
+		ProtoDecoder:     newHistoricalTxDecoder(queryCodec),
+	})
+	if err != nil {
+		return err
+	}
+	app.queryTxConfig = txConfig
+	return nil
+}
+
+// buildQueryInterfaceRegistry mirrors the app's registry (same address codecs,
+// same modules) into a fresh registry, then adds the historical tx types that are
+// not registered for consensus. Kept separate so these additions can never widen
+// what the consensus decoder accepts.
+func (app *AlloraApp) buildQueryInterfaceRegistry() (codectypes.InterfaceRegistry, error) {
+	// Reuse the app's exact address codecs so signer resolution matches.
+	sc := app.interfaceRegistry.SigningContext()
+	reg, err := codectypes.NewInterfaceRegistryWithOptions(codectypes.InterfaceRegistryOptions{
+		ProtoFiles: gogoproto.HybridResolver,
+		//nolint:exhaustruct // the remaining signing options default exactly as the app's registry
+		SigningOptions: txsigning.Options{
+			AddressCodec:          sc.AddressCodec(),
+			ValidatorAddressCodec: sc.ValidatorAddressCodec(),
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// std registers sdk.Msg/sdk.Tx/cryptotypes.PubKey as interfaces; without it
+	// RegisterImplementations below has nothing to attach to.
+	std.RegisterInterfaces(reg)
+
+	// Re-run every module's RegisterInterfaces against the new registry. This
+	// covers auth/bank/gov/emissions v2-v9/mint v1beta1 and the IBC/feemarket
+	// modules added imperatively, exactly as SetupAppBuilder does for consensus.
+	module.NewBasicManagerFromManager(app.ModuleManager, CustomModuleBasics()).RegisterInterfaces(reg)
+
+	// The historical extras. mint.v2 is the only tx-bearing version not registered
+	// anywhere for consensus (the mint tx lineage is v1beta1 -> v2 -> v5; v3 is
+	// events-only and v4 is query/genesis-only, so neither declares an sdk.Msg).
+	// This is a query-registry-only addition and does not touch consensus; verified
+	// by TestMintV2QueryRegistryDoesNotAffectConsensus.
+	mintv2.RegisterInterfaces(reg)
+
+	if err := reg.SigningContext().Validate(); err != nil {
+		return nil, err
+	}
+	return reg, nil
+}
+
+// compositeTxServer serves the tx service with two backends: the strict, embedded
+// server handles every method by default (and any method a future SDK adds), while
+// only the historical-read methods are delegated to the tolerant server.
+type compositeTxServer struct {
+	txtypes.ServiceServer                       // strict; default for all methods
+	tolerant              txtypes.ServiceServer // tolerant; historical reads only
+}
+
+// GetTx reads a committed tx by hash; route to the tolerant server so pre-upgrade
+// payloads decode.
+func (s compositeTxServer) GetTx(ctx context.Context, req *txtypes.GetTxRequest) (*txtypes.GetTxResponse, error) {
+	return s.tolerant.GetTx(ctx, req)
+}
+
+// GetTxsEvent searches committed txs; route to the tolerant server.
+func (s compositeTxServer) GetTxsEvent(ctx context.Context, req *txtypes.GetTxsEventRequest) (*txtypes.GetTxsEventResponse, error) {
+	return s.tolerant.GetTxsEvent(ctx, req)
+}
+
+// GetBlockWithTxs decodes every tx in a committed block; route to the tolerant server.
+func (s compositeTxServer) GetBlockWithTxs(ctx context.Context, req *txtypes.GetBlockWithTxsRequest) (*txtypes.GetBlockWithTxsResponse, error) {
+	return s.tolerant.GetBlockWithTxs(ctx, req)
+}
+
+// RegisterTxService overrides the promoted runtime.App method. It registers a
+// composite tx service: strict for user-supplied and non-decoding methods
+// (including TxDecode, which asserts the SDK's own concrete type), tolerant only
+// for historical reads. It must NOT also call the embedded method, or the gRPC
+// router panics on duplicate service registration.
+func (app *AlloraApp) RegisterTxService(clientCtx client.Context) {
+	// Strict server: the app's own client context, unchanged consensus behavior.
+	strict := authtx.NewTxServer(clientCtx, app.Simulate, app.interfaceRegistry)
+
+	// Tolerant server: same context but with the read-only tx config swapped in.
+	tolerantCtx := clientCtx.WithTxConfig(app.queryTxConfig)
+	tolerant := authtx.NewTxServer(tolerantCtx, app.Simulate, app.queryInterfaceRegistry)
+
+	txtypes.RegisterServiceServer(
+		app.GRPCQueryRouter(),
+		compositeTxServer{ServiceServer: strict, tolerant: tolerant},
+	)
+}
--- a/app/querytx.go
+++ b/app/querytx.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	txsigning "cosmossdk.io/x/tx/signing"
@@ -55,9 +56,9 @@ func (h historicalTx) GetMsgsV2() ([]protov2.Message, error) {
 // GetProtoTx satisfies protoTxProvider, used by GetBlockWithTxs.
 func (h historicalTx) GetProtoTx() *txtypes.Tx { return h.tx }
 
-// AsAny satisfies intoAny, used by GetTx/GetTxsEvent. UnsafePackAny (not
-// NewAnyWithValue) is required so the cached value is the concrete *txtypes.Tx
-// that mkTxResult then type-asserts.
+// AsAny satisfies intoAny, used by GetTx/GetTxsEvent. UnsafePackAny caches the
+// concrete *txtypes.Tx that mkTxResult type-asserts, without an error return
+// AsAny cannot surface.
 func (h historicalTx) AsAny() *codectypes.Any { return codectypes.UnsafePackAny(h.tx) }
 
 // newHistoricalTxDecoder decodes committed txs for the query path. It tries the
@@ -68,8 +69,15 @@ func newHistoricalTxDecoder(cdc codec.Codec) sdk.TxDecoder {
 	strict := authtx.DefaultTxDecoder(cdc)
 	return func(txBytes []byte) (sdk.Tx, error) {
 		// Current-format txs: unchanged behavior and unchanged return type.
-		if tx, err := strict(txBytes); err == nil {
+		tx, strictErr := strict(txBytes)
+		if strictErr == nil {
 			return tx, nil
+		}
+
+		// Both decoders failing means the payload is malformed rather than
+		// historical, and the strict error is the one that carries the reason.
+		fail := func(err error) (sdk.Tx, error) {
+			return nil, sdkerrors.ErrTxDecode.Wrap(errors.Join(err, strictErr).Error())
 		}
 
 		// Historical payloads: unmarshal the envelope directly. Any resolution
@@ -77,15 +85,22 @@ func newHistoricalTxDecoder(cdc codec.Codec) sdk.TxDecoder {
 		// here; only the strict unknown-field rejection is skipped.
 		var raw txtypes.TxRaw
 		if err := cdc.Unmarshal(txBytes, &raw); err != nil {
-			return nil, sdkerrors.ErrTxDecode.Wrap(err.Error())
+			return fail(err)
 		}
 		var body txtypes.TxBody
 		if err := cdc.Unmarshal(raw.BodyBytes, &body); err != nil {
-			return nil, sdkerrors.ErrTxDecode.Wrap(err.Error())
+			return fail(err)
+		}
+		// An Any with an empty type URL unmarshals without error but caches no
+		// value, which makes GetMsgs panic; unknown non-empty URLs already fail above.
+		for _, msg := range body.Messages {
+			if msg == nil || msg.GetCachedValue() == nil {
+				return fail(errors.New("message Any has no resolvable type URL"))
+			}
 		}
 		var authInfo txtypes.AuthInfo
 		if err := cdc.Unmarshal(raw.AuthInfoBytes, &authInfo); err != nil {
-			return nil, sdkerrors.ErrTxDecode.Wrap(err.Error())
+			return fail(err)
 		}
 		return historicalTx{tx: &txtypes.Tx{
 			Body:       &body,
--- a/app/querytx.go
+++ b/app/querytx.go
@@ -18,6 +18,8 @@ import (
 	gogoproto "github.com/cosmos/gogoproto/proto"
 	protov2 "google.golang.org/protobuf/proto"
 
+	emissionsv7 "github.com/allora-network/allora-chain/x/emissions/api/emissions/v7"
+	emissionsv8 "github.com/allora-network/allora-chain/x/emissions/api/emissions/v8"
 	mintv2 "github.com/allora-network/allora-chain/x/mint/api/mint/v2"
 )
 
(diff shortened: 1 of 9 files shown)


app/querytx.go
```diff
@@ -0,0 +1,205 @@
...
+// historicalTx is the minimal sdk.Tx the tolerant decoder returns for a payload
+// the strict decoder rejects. It exposes exactly the three shapes the read-only
+// tx service asserts: sdk.Tx, intoAny (AsAny), and protoTxProvider (GetProtoTx).
+type historicalTx struct {
+	tx *txtypes.Tx
+}
+
+// GetMsgs returns the already-unpacked messages; the tolerant decoder unpacks the
+// Anys during Unmarshal, so the cached values are present.
+func (h historicalTx) GetMsgs() []sdk.Msg { return h.tx.GetMsgs() }
```
helder-moreira, line 50: An Any with an empty TypeUrl slips through `UnpackAny` uncached, so the fallback returns a tx whose `GetMsgs()` panics. Strict rejects the same bytes. Latent today since the routed RPCs use `GetCachedValue`/`GetProtoTx`, but this is a public decoder handing back an `sdk.Tx` that panics on a core method.

app/querytx.go
```diff
@@ -0,0 +1,205 @@
...
+// GetMsgsV2 is never called on the read path; fail loudly rather than return a
+// half-built value if that ever changes.
+func (h historicalTx) GetMsgsV2() ([]protov2.Message, error) {
+	return nil, fmt.Errorf("historicalTx: GetMsgsV2 is not supported on the read-only query path")
+}
+
+// GetProtoTx satisfies protoTxProvider, used by GetBlockWithTxs.
+func (h historicalTx) GetProtoTx() *txtypes.Tx { return h.tx }
+
+// AsAny satisfies intoAny, used by GetTx/GetTxsEvent. UnsafePackAny (not
```
helder-moreira: `NewAnyWithValue` caches the same way — the two differ only in error handling. The choice is fine (matches the SDK's `wrapper.AsAny`), the stated reason isn't.

app/querytx.go
```diff
@@ -0,0 +1,205 @@
...
+
+// newHistoricalTxDecoder decodes committed txs for the query path. It tries the
+// strict SDK decoder first, so every tx that decodes today keeps the identical
+// code path and concrete type; only when that fails (a historical payload) does
+// it fall back to a plain unmarshal that skips unknownproto's nested-field walk.
+func newHistoricalTxDecoder(cdc codec.Codec) sdk.TxDecoder {
+	strict := authtx.DefaultTxDecoder(cdc)
+	return func(txBytes []byte) (sdk.Tx, error) {
+		// Current-format txs: unchanged behavior and unchanged return type.
+		if tx, err := strict(txBytes); err == nil {
```
helder-moreira: When both decoders fail you surface the fallback's error, but the strict one is what says *why*. Keep it for that case?

app/querytx.go
```diff
@@ -0,0 +1,205 @@
...
+			return tx, nil
+		}
+
+		// Historical payloads: unmarshal the envelope directly. Any resolution
+		// still goes through cdc's registry, so an unknown type URL still errors
+		// here; only the strict unknown-field rejection is skipped.
+		var raw txtypes.TxRaw
```
helder-moreira, line 88: Unknown critical fields get silently dropped here too — re-marshaling gives different bytes than were committed, so `GetBlockWithTxs` renders a mutated tx where it used to skip it. Minor on its own, but with the empty-TypeUrl case above I'd rather reject bodies with unresolvable or empty message Anys than return a half-built tx.

app/querytx.go
```diff
@@ -0,0 +1,205 @@
...
+// same modules) into a fresh registry, then adds the historical tx types that are
+// not registered for consensus. Kept separate so these additions can never widen
+// what the consensus decoder accepts.
+func (app *AlloraApp) buildQueryInterfaceRegistry() (codectypes.InterfaceRegistry, error) {
+	// Reuse the app's exact address codecs so signer resolution matches.
+	sc := app.interfaceRegistry.SigningContext()
+	reg, err := codectypes.NewInterfaceRegistryWithOptions(codectypes.InterfaceRegistryOptions{
+		ProtoFiles: gogoproto.HybridResolver,
+		//nolint:exhaustruct // the remaining signing options default exactly as the app's registry
+		SigningOptions: txsigning.Options{
```
helder-moreira, line 153: `ProvideInterfaceRegistry` also injects any `CustomGetSigner`s; this takes only the address codecs. None exist today, but if one ever appears this `Validate()` fails and nodes won't boot. Any way to reuse the app's signing context instead of rebuilding from two fields?

app/querytx.go
```diff
@@ -0,0 +1,205 @@
...
+	// covers auth/bank/gov/emissions v2-v9/mint v1beta1 and the IBC/feemarket
+	// modules added imperatively, exactly as SetupAppBuilder does for consensus.
+	module.NewBasicManagerFromManager(app.ModuleManager, CustomModuleBasics()).RegisterInterfaces(reg)
+
+	// The historical extras. mint.v2 is the only tx-bearing version not registered
+	// anywhere for consensus (the mint tx lineage is v1beta1 -> v2 -> v5; v3 is
+	// events-only and v4 is query/genesis-only, so neither declares an sdk.Msg).
+	// This is a query-registry-only addition and does not touch consensus; verified
+	// by TestMintV2QueryRegistryDoesNotAffectConsensus.
+	mintv2.RegisterInterfaces(reg)
```
helder-moreira, line 176: Worth rewording in the description: the gateway marshals a page as a unit, so one `mint.v2` tx fails the whole query rather than just rendering badly. Looks like zero impact though — gov history has no `mint.v2` message at all, only `mint.v5` in #12 — so wording rather than code.

test/api/historical_tx_decode_test.go
```diff
@@ -0,0 +1,71 @@
...
+					TopicId: 1, BlockHeight: 1, Inferer: sender, Value: "1.0",
+				},
+			},
+			InferencesForecastsBundleSignature: []byte{0x01},
+			Pubkey:                             "pubkey",
+		},
+	}
+}
+
+// TestHistoricalTxDoesNotMoveConsensus is the end-to-end regression PR #985 lacked.
```
helder-moreira: No PR references in comments per `CLAUDE.md` — describe the behaviour instead. Same on line 65.

test/api/historical_tx_decode_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	ctx := chain.GetContext()
+	balBefore := alloraApp.BankKeeper.GetBalance(ctx, sender, sdk.DefaultBondDenom)
+	seqBefore := alloraApp.AccountKeeper.GetAccount(ctx, sender).GetSequence()
+
+	resp := SignAndDeliver(t, chain, wrapped)
+	require.Len(t, resp.TxResults, 1)
+
+	// Consensus decoder rejected it: non-zero code and, crucially, nothing executed.
+	require.NotZero(t, resp.TxResults[0].Code, "wrapped historical tx must be rejected")
```
helder-moreira: Pin the exact code (2)? `NotZero` still passes if the rejection moves from the decoder to the ante handler, which is the regression this is guarding against.

helder-moreira (commented):
Checked the consensus claim rather than taking it on reading, since #985 looked fine on inspection too. Diffed both registries at runtime (query = app + exactly the two `mint.v2` msgs, nothing missing), and ran a `mint.v2` wrapped in `MsgExec` through `FinalizeBlock` on dev vs this branch with the Any hand-encoded so dev never links `mintv2` — identical code, gas and state, so the new import doesn't move consensus. Also confirmed a v9 tx renders through the real gateway marshaler, which nothing here covers (the tests all call the decoder directly; the bug report is a REST endpoint).

Consensus safety holds. Comments are mostly about the tolerant decoder accepting more than historical payloads.

On rollout: agree on snapshot testing and the single-RPC canary. Worth hitting `GetTx` and `GetBlockWithTxs` too, not just `txs?query=` — different paths, and the shim's interfaces are satisfied structurally, so a mismatch only shows at runtime.

test/api/harness_test.go
```diff
@@ -0,0 +1,93 @@
...
+func AppInitializer() (ibctesting.TestingApp, map[string]json.RawMessage) {
+	testApp, err := app.NewAlloraApp(
+		log.NewNopLogger(),
+		dbm.NewMemDB(),
+		nil,
+		true,
+		simtestutil.EmptyAppOptions{},
+	)
+	if err != nil {
+		return nil, nil
```
helder-moreira: `return nil, nil` hides construction failures. Now that `buildQueryDecodingPath` can fail, that's the error you'd most want to see.

app/querytx_test.go
```diff
@@ -0,0 +1,369 @@
...
+var (
+	sharedAppOnce sync.Once
+	sharedAppInst *AlloraApp
+)
+
+func sharedApp(t *testing.T) *AlloraApp {
+	t.Helper()
+	sharedAppOnce.Do(func() {
+		a, err := NewAlloraApp(log.NewNopLogger(), dbm.NewMemDB(), nil, true, simtestutil.EmptyAppOptions{})
+		require.NoError(t, err)
```
helder-moreira: `FailNow` inside the `Once` means later tests nil-deref instead of showing the real error. Store the err and re-assert per call.

app/querytx_test.go
```diff
@@ -0,0 +1,369 @@
...
+			}
+		})
+		t.Run(tc.name+"_wrapped", func(t *testing.T) {
+			inner, err := codectypes.NewAnyWithValue(tc.msg)
+			require.NoError(t, err)
+			//nolint:exhaustruct // only Grantee and Msgs matter
+			exec := &authz.MsgExec{Grantee: "allo1grantee", Msgs: []*codectypes.Any{inner}}
+			tx, err := dec(buildTxBytes(t, exec))
+			require.NoError(t, err)
+			require.Equal(t, "/cosmos.authz.v1beta1.MsgExec", sdk.MsgTypeURL(tx.GetMsgs()[0]))
```
helder-moreira, line 179: Only asserts the outer `MsgExec` URL — an undecoded inner Any would pass this.

app/querytx_test.go
```diff
@@ -0,0 +1,369 @@
...
+// Scenario 2: the SAME bare v9 bytes are rejected by the consensus decoder. This
+// is the guard that the fix does not widen what consensus accepts.
+func TestConsensusDecoderRejectsBareHistoricalV9(t *testing.T) {
+	a := sharedApp(t)
+	_, err := a.txConfig.TxDecoder()(buildTxBytes(t, historicalV9Worker()))
+	require.Error(t, err, "consensus decoder must still reject historical payloads")
+}
+
+// Scenario 4: a v9 payload wrapped in a routable authz.MsgExec is rejected by the
+// consensus decoder. This
```
helder-moreira: Nit: line break mid-sentence, and the scenario numbering skips 3.

app/querytx_test.go
```diff
@@ -0,0 +1,369 @@
...
+	require.NoError(t, err)
+	_, isFallback := tx.(historicalTx)
+	require.False(t, isFallback, "current txs must take the strict path")
+	require.Equal(t, v10WorkerURL, sdk.MsgTypeURL(tx.GetMsgs()[0]))
+}
+
+// Scenario 6: the query registry resolves mint.v2 while the consensus registry and
+// the process-global gogo registry do not. This proves the mint.v2 addition is
+// isolated to the read path and cannot affect consensus.
+func TestMintV2QueryRegistryDoesNotAffectConsensus(t *testing.T) {
```
helder-moreira, line 229: Two worth adding: something that marshals through the gateway (REST resolves via the *app* registry, and nothing guards that), and a `consensus ⊆ query` registry assertion to freeze the mirror. Also `RecalculateTargetEmissionRequest` is registered but never asserted.

app/querytx.go
xmariachi: Confirmed and fixed in f31f367.

Reproduced it before changing anything: a `TxBody` with a single `Any{TypeUrl: ""}` is rejected by the strict decoder but accepted by the fallback, and `GetMsgs()` then panics with `any cached value is nil, transaction messages must be correctly packed any values`. The cause is the early return in `statefulUnpacker.UnpackAny` — empty `TypeUrl` returns `nil` with no error and no `cachedValue`, so `cdc.Unmarshal` succeeds.

`newHistoricalTxDecoder` now rejects a body carrying such a message:

```go
for _, msg := range body.Messages {
    if msg == nil || msg.GetCachedValue() == nil {
        return fail(errors.New("message Any has no resolvable type URL"))
    }
}
```

Worth noting why this cannot narrow what the query path accepts: a non-empty but unregistered type URL already errors inside `cdc.Unmarshal` (`no concrete type registered for type URL %s`), so `TypeUrl == ""` is the only way a nil cached value survives a successful unmarshal. Every v2-v9 and `mint.v2` payload still decodes. `TestQueryDecoderRejectsUnresolvableMessageAny` covers it, and I verified it fails (returns nil error) with the guard reverted.

app/querytx.go
xmariachi: You are right — fixed in f31f367. `UnsafePackAny` calls `NewAnyWithValue` and returns its result for any valid `proto.Message`, so the caching is identical; the only difference is the error return. Comment now states the actual reason:

```go
// AsAny satisfies intoAny, used by GetTx/GetTxsEvent. UnsafePackAny caches the
// concrete *txtypes.Tx that mkTxResult type-asserts, without an error return
// AsAny cannot surface.
```

app/querytx.go
xmariachi: Good catch — kept in f31f367. The strict error is now retained and joined into every fallback failure:

```go
fail := func(err error) (sdk.Tx, error) {
    return nil, sdkerrors.ErrTxDecode.Wrap(errors.Join(err, strictErr).Error())
}
```

Fallback error stays primary so historical rejections read naturally, with the strict reason appended — which is what explains a malformed current-format tx.

app/querytx.go
xmariachi: Took the option you preferred: f31f367 rejects bodies with unresolvable or empty message `Any`s rather than returning a half-built tx, which closes this together with the panic above.

Confirmed the mechanism while checking: `txtypes.TxBody` has no `XXX_unrecognized` field, so unknown fields genuinely are dropped and a re-marshal is not byte-identical. Did not pursue full byte fidelity — the envelope fields are stable across the versions in play and the inner `Any.Value` bytes are preserved verbatim, so the remaining exposure needed the half-built tx to be reachable at all. It no longer is.

test/api/historical_tx_decode_test.go
xmariachi: Fixed in f31f367 — both comments reworded to describe the behaviour, and I grepped the diff for any other ticket/PR references. Now:

```go
// TestHistoricalTxDoesNotMoveConsensus delivers a historical v9 payload wrapped in
// a routable authz.MsgExec through a real AlloraApp FinalizeBlock and asserts
// consensus state is untouched: rejected at decode, with no gas, no fee deduction
// and no sequence increment. A decoder that accepted the nested payload would run
// the ante handler and move all four.
```

and line 65 is now `// State a widened decoder would have moved must be unchanged.`

test/api/historical_tx_decode_test.go
xmariachi: Agreed and pinned in f31f367:

```go
require.Equal(t, sdkerrors.ErrTxDecode.ABCICode(), resp.TxResults[0].Code,
    "wrapped historical tx must be rejected at decode")
require.Equal(t, sdkerrors.ErrTxDecode.Codespace(), resp.TxResults[0].Codespace)
```

Measured the actual values rather than assuming: current branch gives `code=2 codespace="sdk" gasUsed=0 log="tx parse error"`. For contrast, an executing tx through the same harness gives `code=0 gasUsed=142303` — so your scenario (rejection migrating to the ante handler) would now fail on the code, and separately on `GasUsed`.

test/api/harness_test.go
xmariachi: Fixed in f31f367, with the error wrapped for context:

```go
if err != nil {
    // The hook signature cannot return an error, and a nil app surfaces as an
    // unrelated nil dereference inside the coordinator.
    panic(fmt.Errorf("initialize AlloraApp test fixture: %w", err))
}
```

app/querytx_test.go
xmariachi: Fixed in f31f367 — error stored and re-asserted per call, so the `Goexit` path cannot leave a nil app behind:

```go
var (
    sharedAppOnce sync.Once
    sharedAppInst *AlloraApp
    errSharedApp  error
)

func sharedApp(t *testing.T) *AlloraApp {
    t.Helper()
    sharedAppOnce.Do(func() {
        sharedAppInst, errSharedApp = NewAlloraApp(
            log.NewNopLogger(), dbm.NewMemDB(), nil, true, simtestutil.EmptyAppOptions{},
        )
    })
    require.NoError(t, errSharedApp)
    return sharedAppInst
}
```

(Named `errSharedApp` rather than `sharedAppErr` to satisfy the `errname` linter.)

app/querytx_test.go
xmariachi: Right — the wrapped case existed to cover nested decoding and was not asserting it. f31f367 adds:

```go
outer, ok := tx.GetMsgs()[0].(*authz.MsgExec)
require.True(t, ok)
require.Len(t, outer.Msgs, 1)
require.Equal(t, sdk.MsgTypeURL(tc.msg), outer.Msgs[0].TypeUrl)
require.NotNil(t, outer.Msgs[0].GetCachedValue())
```

Current behaviour already satisfied this (the inner `Any` caches as `*emissionsv9.InsertWorkerPayloadRequest`), so this is purely about the test being able to fail.

app/querytx_test.go
xmariachi: Fixed in f31f367. Rejoined the sentence and dropped the `Scenario N:` prefixes entirely rather than renumbering — the test names already carry the meaning, and the numbering was what drifted.

app/querytx.go
xmariachi: Valid, and partially addressed in f31f367 — leaving this open for the part I did not do.

Confirmed the exposure is latent: no `CustomGetSigner` exists anywhere in the repo today, so `SigningContext().Validate()` passes and nothing is broken now. But you are right that the failure mode is a hard boot failure landing far from its cause.

What f31f367 adds is a drift guard rather than the structural fix you asked about:

```go
func TestQueryRegistryIsSupersetOfConsensus(t *testing.T) {
	a := sharedApp(t)
	ifaces := a.interfaceRegistry.ListAllInterfaces()
	require.NotEmpty(t, ifaces, "consensus registry must expose interfaces to compare against")
	for _, iface := range ifaces {
		for _, impl := range a.interfaceRegistry.ListImplementations(iface) {
			_, err := a.queryInterfaceRegistry.Resolve(impl)
			require.NoErrorf(t, err, "query registry must resolve %s", impl)
		}
	}
}
```

That currently checks 512 implementations across 28 interfaces, so a message registered for consensus but missed on the read path fails in CI with a precise message instead of at app construction.

On your actual question — reusing the apps signing context instead of rebuilding from two fields — I did not find a clean way: the SDK does not expose the app registrys custom-signer set for re-injection into `NewInterfaceRegistryWithOptions`. Open to suggestions if you know one; otherwise the parity test is the mitigation and this stays a known assumption.

app/querytx.go
xmariachi: Agreed on the wording, and not done in f31f367 — that commit is code only, so leaving this open.

Confirmed your read of the impact: the gateway marshals a page as a unit, so one `mint.v2` tx fails the whole query rather than rendering partially. I will update the description to say that explicitly, and to note the historical scan found no `mint.v2` messages in gov history (only `mint.v5`, proposal #12), so live impact is nil.

app/querytx_test.go
xmariachi: Two of the three are in f31f367; the gateway test is not, so leaving this open.

Done:
- `RecalculateTargetEmissionRequest` is now asserted alongside `UpdateParamsRequest` — the mint.v2 test loops over both type URLs against the query registry, the consensus registry, and the global gogo registry.
- `consensus ⊆ query` is frozen by `TestQueryRegistryIsSupersetOfConsensus` (512 impls / 28 interfaces).

Not done — tracked as follow-up: the gateway-marshaling test. This is the substantive one and I did not want to rush it: REST resolves response `Any`s via the *app* registry, not the query registry, which is exactly why `mint.v2` cannot render there while v9 can. Since the reported bug is a REST endpoint, the plan is a test that marshals a `GetTxResponse` through the app-registry `ProtoCodec.MarshalJSON` for a v9 tx (must render) and a `mint.v2` tx (asserts and documents the limitation), alongside an end-to-end test driving `GetTx`, `GetTxsEvent` and `GetBlockWithTxs` so the structural `intoAny`/`protoTxProvider` assertions actually execute in CI.

test/api/historical_tx_decode_test.go
```diff
@@ -0,0 +1,71 @@
...
+	require.Zero(t, resp.TxResults[0].GasUsed, "a rejected-at-decode tx must not consume gas")
+
+	// The state the PR #985 hazard mutated must be unchanged.
+	ctxAfter := chain.GetContext()
```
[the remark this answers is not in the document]
xmariachi: Checked this one carefully because it is the only P2 raised, and I do not think it holds — no change made.

Two of the three premises are correct: `SignAndDeliver` never calls `Commit()`, and `chain.GetContext()` is `NewUncachedContext(false, header)` = `sdk.NewContext(app.cms, ...)`, i.e. the root multistore. The step that does not follow is that `app.cms` is untouched before `Commit()`. `BaseApp.FinalizeBlock` ends with:

```go
res, err = app.internalFinalizeBlock(context.Background(), req)
if res != nil {
    res.AppHash = app.workingHash()
}
```

and `workingHash()` (`baseapp/abci.go`) begins:

```go
// The write to the FinalizeBlock state writes all state transitions to the root
// MultiStore (app.cms) so when Commit() is called it persists those values.
app.finalizeBlockState.ms.Write()
```

So the blocks writes are already in `app.cms` when `FinalizeBlock` returns; `Commit()` persists to disk and bumps the version, it is not what makes them visible.

Verified by positive control — routable bank send, same harness, no `Commit()`:

```
BEFORE bal=10000000000000000000stake seq=0
RESULT code=0 gasUsed=142303
AFTER  bal=9999999998999987655stake seq=1
```

Balance moved by exactly 1_000_012_345 (1e9 fee + 12_345 sent) and the sequence incremented, both read through `chain.GetContext()`. So the assertions are live and would catch a decoder widening.

Adding `chain.App.Commit()` would also desync the harness: ibctestings `commitBlock` additionally rotates `Vals`/`NextVals`, sets `LastHeader` and advances `CurrentHeader`, so a bare `Commit()` leaves `chain.CurrentHeader` pointing at the executed height.

The concern behind this was still worth acting on, though: nothing in-repo demonstrated the assertions were sensitive. f31f367 adds `TestHarnessObservesExecutedTx` as a permanent positive control, so if a future refactor of `SignAndDeliver` ever does make those reads blind, it fails loudly instead of passing silently.

test/api/harness_test.go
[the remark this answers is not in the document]
xmariachi: Fixed in f31f367 — `panic(fmt.Errorf("initialize AlloraApp test fixture: %w", err))`, so a construction failure surfaces with its real cause instead of a nil dereference in the coordinator.

app/querytx_test.go
```diff
@@ -0,0 +1,369 @@
...
+
+// One AlloraApp is enough for every read-only assertion here; build it once.
+var (
+	sharedAppOnce sync.Once
+	sharedAppInst *AlloraApp
+)
+
+func sharedApp(t *testing.T) *AlloraApp {
+	t.Helper()
+	sharedAppOnce.Do(func() {
```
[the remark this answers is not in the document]
xmariachi: Fixed in f31f367 — your `Goexit` analysis was exactly right. The app and its error are now stored together and `require.NoError(t, errSharedApp)` runs on every call, so a failed construction cannot leave later tests with a nil app.

test/api/historical_tx_decode_test.go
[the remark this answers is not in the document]
xmariachi: Fixed in f31f367 — both comments reworded to describe the behaviour, with no PR references.

app/querytx.go
```diff
@@ -0,0 +1,264 @@
...
+	registerHistoricalWhitelistMsgs(reg)
+
+	if err := reg.SigningContext().Validate(); err != nil {
+		return nil, err
+	}
+	return reg, nil
+}
+
+// registerHistoricalWhitelistMsgs registers the v7/v8 whitelist txs their own
```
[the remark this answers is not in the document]
xmariachi: comment aligned on https://github.com/allora-network/allora-chain/pull/986/commits/8975879da942359341f7c1061a49c564a14d921f

test/api/historical_tx_decode_test.go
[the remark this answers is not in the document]
xmariachi: Detailed follow-up on this one, since it's the only P2 and the fix it proposes would be harmful — so it's worth pinning down precisely why the assertions are **not** vacuous.

### The claim, and which parts are true

> balAfter/seqAfter are identical to balBefore/seqBefore even if the tx moved state, because `SignAndDeliver` never `Commit()`s and `chain.GetContext()` reads the committed multistore.

Two of the three premises are correct:

1. ✅ `SignAndDeliver` calls `chain.App.GetBaseApp().FinalizeBlock(...)` and never `Commit()`.
2. ✅ `chain.GetContext()` is `NewUncachedContext(false, header)` = `sdk.NewContext(app.cms, ...)` — the **root** multistore, not a cache branch.

The third — that `app.cms` is unchanged until `Commit()` — is where it breaks.

### The step that's missing: `FinalizeBlock` flushes into `app.cms` before it returns

`BaseApp.FinalizeBlock` ends with:

```go
res, err = app.internalFinalizeBlock(context.Background(), req)
if res != nil {
    res.AppHash = app.workingHash()   // <- here
}
```

and `workingHash()` (`baseapp/abci.go`) is, verbatim:

```go
// The write to the FinalizeBlock state writes all state transitions to the root
// MultiStore (app.cms) so when Commit() is called it persists those values.
func (app *BaseApp) workingHash() []byte {
    app.finalizeBlockState.ms.Write()   // <- all block writes land in app.cms
    commitHash := app.cms.WorkingHash()
    ...
}
```

`finalizeBlockState.ms` is a cache-multistore over `app.cms`; `.Write()` pushes every dirty key into the underlying stores' working trees, which subsequent reads of `app.cms` observe immediately. `Commit()` (`app.cms.Commit()`) persists those to disk and bumps the store version — it is **not** what makes them visible. The SDK's own comment on `workingHash` says exactly this.

So after `FinalizeBlock` returns, a fresh `chain.GetContext()` over `app.cms` **does** see the block's writes, with no `Commit()`.

### Positive control (measured, not argued)

Same harness, a routable bank send instead of the rejected historical tx, no `Commit()`:

```
BEFORE  bal=10000000000000000000stake  seq=0
RESULT  code=0  gasUsed=142303
AFTER   bal=9999999998999987655stake   seq=1
```

Balance moved by exactly `1_000_012_345` (`1e9` fee + `12_345` sent) and the sequence incremented — both read through `chain.GetContext()`, with no `Commit()`. So the balance/sequence assertions are live and would fail if the historical tx ever started executing (i.e. if the decoder widened and the ante handler ran).

For contrast, the historical tx actually under test measures `code=2 codespace="sdk" gasUsed=0` — rejected at decode, nothing executed — which is why `balBefore == balAfter` there is a real result, not an artifact.

### Why the suggested fix would make it worse

Inserting `chain.App.Commit()` before `ctxAfter` would (a) change nothing observable, and (b) desync the harness: ibctesting's `commitBlock` also rotates `Vals`/`NextVals`, sets `LastHeader`, and advances `CurrentHeader`. A bare `Commit()` leaves `chain.CurrentHeader` pointing at the just-executed height, so any later chain interaction in the test is subtly wrong.

### What did change

The concern was still worth acting on, because nothing in-repo demonstrated the assertions were sensitive. `f31f367` adds `TestHarnessObservesExecutedTx` as a permanent positive control: if a future refactor of `SignAndDeliver` ever makes those reads blind, it fails loudly there instead of silently passing here.

Resolving as not-a-defect on that basis.

app/querytx.go
xmariachi: Done — updated the description's **Known limitation** to state that the gateway marshals a result page as a unit, so a single `mint.v2` tx fails the whole query rather than rendering partially, and that a scan of gov history found no `mint.v2` messages at all (only `mint.v5`, proposal #12), so there is no page it can affect. Resolving.

app/querytx.go
xmariachi: Closing this out — summary of where it landed.

**Latent, not active.** No `CustomGetSigner` exists anywhere in the repo, and every Msg-service input (379 across the linked proto set) carries the `(cosmos.msg.v1.signer)` annotation, so none is needed. The query registry already `Validate()`s with zero custom signers, which is the exact equivalent of "every message is annotated."

**No clean structural fix.** I looked at reusing the app's signing context instead of rebuilding from the two codecs, as you suggested. `signing.Context` exposes the address codecs (`AddressCodec()`/`ValidatorAddressCodec()`) but not the custom-getSigners set — it's write-only via `DefineCustomGetSigners`, nothing reads it back — so there's no way to lift them off the app registry and re-inject them into `NewInterfaceRegistryWithOptions`. If you know a cleaner path I'm happy to revisit.

**What changed.** Documented the invariant in-code at `buildQueryInterfaceRegistry`, so the failure mode you flagged is explicit for the next editor:

```go
// Only the address codecs are copied, not any CustomGetSigners: none exist today
// (every Msg carries the (cosmos.msg.v1.signer) annotation). If one is ever added
// it must be mirrored here too, or Validate() below fails and the node won't start.
```

**Enforcement.** That `SigningContext().Validate()` at construction is the guard: a custom signer added to the app registry but not mirrored here makes `NewAlloraApp` fail, so every app-building test (`sharedApp`, the `test/api` harness) goes red in CI. It can't reach production silently — it's a loud boot-time failure, the safe direction — so no dedicated test is warranted.

Resolving on that basis.

zale144 (approved):

PR #993, opened 2026-09-09
ENGN-9363 fix: open the reputer submission window at the end of the reveal epoch

helder-moreira:
## Bug

Reputer submission windows overlap when a topic's `ground_truth_lag` is not a whole multiple of `epoch_length`. `extraLag = epoch_length - (ground_truth_lag % epoch_length)` (zero when it divides evenly).

`UpdateReputerNonce` closes a nonce on the topic's own epoch grid, at the first grid block `>= nonce+ground_truth_lag+epoch_length`, which for an uninterrupted grid is exactly `nonce+ground_truth_lag+extraLag+epoch_length`. But the reputer submission window's *lower* bound (`BlockWithinReputerSubmissionWindowOfNonce`) was `nonce+ground_truth_lag` — `extraLag` blocks *before* that. So the next nonce's window opened while the previous nonce was still open.

A reputer submitting during that overlap stores its bundle for the *newer* nonce. `lossBundles` is keyed `(topic_id, reputer)` with no nonce, so when the *older* nonce closes, `CloseReputerNonce` reads that bundle, `CalcNetworkLosses` skips every report whose `ReputerRequestNonce.BlockHeight` doesn't match the nonce being closed, `SumWeight` stays zero, and it fails permanently with `"Sum weight for loss is 0"`. The topic then never produces `EventNetworkLossSet` again, and since `SetTopicRewardNonce` is only reached on success, it pays no further rewards. Every reputer transaction still succeeds, because it's accepted for the *next* nonce — the failure is silent except for this error.

Topics with an evenly-dividing lag never see this (`extraLag == 0`), which is why it doesn't affect all topics.

## Live confirmation

Grepped a running testnet validator's own logs just now (not simulated):

```json
{"level":"warn","expected":10839023,"actual":10839077,"reputer":"allo1…905s9ps","message":"Reputer bundle has incorrect block height"}
{"level":"error","topicId":69,"nonce":{"block_height":10839023},"error":"Error normalizing combined loss: Sum weight for loss is 0: fraction: divide by zero","message":"Error occurred before finalization in CloseReputerNonce, attempting cleanup anyway"}
{"level":"error","error":"Error normalizing combined loss: Sum weight for loss is 0: fraction: divide by zero","message":"Error closing reputer nonce"}
```

`actual - expected` equals the topic's own `epoch_length` in every occurrence (54 for topic 69 above). Firing repeatedly, live, on testnet topics 38, 41, 42, 64 and 69 (93 occurrences in one log pull). Also confirmed via chain state: for these topics, the bundle stored under any recently-closed nonce carries `reputer_nonce.block_height == closed_nonce + epoch_length`; topics with an evenly-dividing lag show `0`. Mainnet currently has no topic with a fractional lag, so it isn't exposed today, but nothing prevents creating one.

## Fix

`x/emissions/keeper/window_boundaries.go`: `lowerBound := revealedGroundTruthBlock + extraLag`. The window now opens at the end of the epoch in which ground truth is revealed, matching where the previous nonce actually closes, rather than at the revelation itself.

Extracted the bounds into one function, `ReputerSubmissionWindowBounds`, since the formula was previously duplicated across four call sites and the one that drifted (the acceptance lower bound) is what caused this:
- The topic query server (`GetReputerSubmissionWindowStatus`) was still computing the old bound and would have advertised a window as open that the message server rejects — fixed to use the shared helper.
- The "window not open" error message previously reported wrong bounds unconditionally (start omitted `extraLag`, end used `epoch_length*2` instead of `epoch_length`) — now reports the real bounds.
- The window-opened event previously used its own separate range (`[reveal, reveal+extraLag]`) that no longer aligns with the fixed acceptance start after a topic is reactivated with a shifted epoch grid — an executable test reproduces the exact gap (event fires at a block where a submission would still be rejected) and the fix makes the event fire only at the true window start.

## Testing

- 3 pre-existing test expectations in `window_boundaries_test.go` re-derived for the new bounds (moved, not weakened — verified each old assertion still has a home); added regression cases for the overlap itself and the event/accept-gap.
- Query server test re-derived for the same reason.
- New `ReputerSubmissionWindowBounds` unit tests, including the invariant that consecutive windows tile with no gap and no overlap (`prevWindowEnd == thisWindowStart`).
- Mutation-tested: reverting the lower-bound change fails the keeper and queryserver suites.
- Full suite: `go test ./x/emissions/...` — 21/21 packages pass.
- End-to-end: drove the real `EndBlocker` and message servers over many epochs for a fractional-lag topic on `origin/dev` (traps, matching production) vs this branch (does not trap); an aligned-lag control is unchanged on both.

## Known residual, disclosed rather than hidden

A one-block overlap remains at the exact close/open boundary block, because `DeliverTx` executes before `EndBlock` in the same block — a transaction for the next nonce can still be included in the same block that closes the previous one. This is not new: it already exists identically on every topic whose lag divides evenly, and this PR does not change that shared behavior. Closing it fully would need either a nonce-aware reset in `CloseReputerNonce` or shifting every topic's boundary by one block, both bigger changes deserving their own review.

## Consensus impact

This changes transaction acceptance (`MsgInsertReputerPayload` succeeds/fails differently at some block heights than before), so it requires a coordinated upgrade height. No store migration; `ConsensusVersion` is unchanged. `CHANGELOG.md` flags it under `### Security`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # [Unreleased]
 
+### Security
+
+* The reputer submission window now opens at the end of the epoch in which ground truth is revealed (`nonce + ground_truth_lag + extraLag`) rather than at the revelation itself. Previously, on a topic whose `ground_truth_lag` was not a whole multiple of `epoch_length`, a nonce's window opened `extraLag` blocks before the *previous* nonce closed on the epoch grid. A reputer submitting in that gap stored a bundle for the newer nonce, and because reputer loss bundles are keyed by `(topic, reputer)` with no nonce, closing the older nonce read that bundle, discarded every report on the block-height check and failed with `Sum weight for loss is 0`. The topic then produced no further network losses and, since `SetTopicRewardNonce` is only reached on success, paid no rewards — permanently, with every transaction still reporting success. Topics whose lag divides evenly were unaffected (`extraLag` is zero for them). **This changes transaction acceptance and therefore requires a coordinated upgrade height.** No store migration; `ConsensusVersion` is unchanged.
+
 ### Added
 
 * [#968](https://github.com/allora-network/allora-chain/pull/968) Per-topic `max_top_inferers_to_reward` (additive field on `emissions.v10.Topic`): the cap on inferers admitted per topic is now topic-level, set at create/`UpdateTopic` (`0` uses the global default; guarded while a worker submission window is open). Stored caps are kept verbatim, and admission resolves them against the global bounds, so the globals stay live. The emissions v16 migration backfills existing topics, so behavior is unchanged across the `v0.18.0` upgrade.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,10 +65,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # [Unreleased]
 
-### Security
-
-* The reputer submission window now opens at the end of the epoch in which ground truth is revealed (`nonce + ground_truth_lag + extraLag`) rather than at the revelation itself. Previously, on a topic whose `ground_truth_lag` was not a whole multiple of `epoch_length`, a nonce's window opened `extraLag` blocks before the *previous* nonce closed on the epoch grid. A reputer submitting in that gap stored a bundle for the newer nonce, and because reputer loss bundles are keyed by `(topic, reputer)` with no nonce, closing the older nonce read that bundle, discarded every report on the block-height check and failed with `Sum weight for loss is 0`. The topic then produced no further network losses and, since `SetTopicRewardNonce` is only reached on success, paid no rewards — permanently, with every transaction still reporting success. Topics whose lag divides evenly were unaffected (`extraLag` is zero for them). **This changes transaction acceptance and therefore requires a coordinated upgrade height.** No store migration; `ConsensusVersion` is unchanged.
-
 ### Added
 
 * [#968](https://github.com/allora-network/allora-chain/pull/968) Per-topic `max_top_inferers_to_reward` (additive field on `emissions.v10.Topic`): the cap on inferers admitted per topic is now topic-level, set at create/`UpdateTopic` (`0` uses the global default; guarded while a worker submission window is open). Stored caps are kept verbatim, and admission resolves them against the global bounds, so the globals stay live. The emissions v16 migration backfills existing topics, so behavior is unchanged across the `v0.18.0` upgrade.
@@ -86,6 +82,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security
 
+* The reputer submission window now opens at the end of the epoch in which ground truth is revealed (`nonce + ground_truth_lag + extraLag`) rather than at the revelation itself. Previously, on a topic whose `ground_truth_lag` was not a whole multiple of `epoch_length`, a nonce's window opened `extraLag` blocks before the *previous* nonce closed on the epoch grid. A reputer submitting in that gap stored a bundle for the newer nonce, and because reputer loss bundles are keyed by `(topic, reputer)` with no nonce, closing the older nonce read that bundle, discarded every report on the block-height check and failed with `Sum weight for loss is 0`. The topic then produced no further network losses and, since `SetTopicRewardNonce` is only reached on success, paid no rewards — permanently, with every transaction still reporting success. Topics whose lag divides evenly were unaffected (`extraLag` is zero for them). **This changes transaction acceptance and therefore requires a coordinated upgrade height.** No store migration; `ConsensusVersion` is unchanged.
 * [#980](https://github.com/allora-network/allora-chain/pull/980) Patch a fast-node cache/commit race in `iavl` that caused `AppHash` divergence: the ecosystem treasury account is deleted every block at zero balance, and a concurrent balance query during commit could read its value one block stale, minting the wrong amount and forking the node. Points `iavl` at `github.com/allora-network/iavl` (`v1.2.6` + backport of [cosmos/iavl#1142](https://github.com/cosmos/iavl/pull/1142)). Not consensus-breaking; no migration required.
 
 ### API Breaking Changes
--- a/x/emissions/module/rewards/nonce_management.go
+++ b/x/emissions/module/rewards/nonce_management.go
@@ -15,14 +15,24 @@ func UpdateReputerNonce(ctx sdk.Context, k keeper.Keeper, topic types.Topic, blo
 		ctx.Logger().Warn("Error getting unfulfilled worker nonces", "error", err)
 		return err
 	}
-	extraLag := topic.GroundTruthLag % topic.EpochLength
-	if extraLag != 0 {
-		extraLag = topic.EpochLength - extraLag
-	}
 	for _, nonce := range nonces.Nonces {
-		if block >= nonce.ReputerNonce.BlockHeight+topic.GroundTruthLag && block <= nonce.ReputerNonce.BlockHeight+topic.GroundTruthLag+extraLag {
-			windowEndBlock := nonce.ReputerNonce.BlockHeight + topic.GroundTruthLag + extraLag + topic.EpochLength
-			types.EmitNewReputerSubmissionWindowOpenedEvent(ctx, topic.Id, nonce.ReputerNonce.BlockHeight, windowEndBlock)
+		// Must match keeper.ReputerSubmissionWindowBounds, the single source of
+		// truth for when a submission is actually accepted. This used to be its
+		// own formula ending at +extraLag instead of the window's real start
+		// (+extraLag was folded into the acceptance lower bound to close a
+		// permanent-trap bug -- see CHANGELOG). On an uninterrupted epoch grid the
+		// two formulas coincide, but a topic that was inactivated and reactivated
+		// gets a new churn schedule with no relation to an existing nonce's grid
+		// (topic_activation.go), so a reactivated topic could hit a block inside
+		// the old range and announce the window open while it was still rejecting
+		// submissions until the true start.
+		windowStart, windowEnd, err := keeper.ReputerSubmissionWindowBounds(topic, *nonce)
+		if err != nil {
+			ctx.Logger().Warn("Error computing reputer submission window bounds", "error", err)
+			continue
+		}
+		if block == windowStart {
+			types.EmitNewReputerSubmissionWindowOpenedEvent(ctx, topic.Id, nonce.ReputerNonce.BlockHeight, windowEnd)
 		}
 		// Check if current blockheight has reached the blockheight of the nonce + groundTruthLag + epochLength
 		// This means one epochLength is allowed for reputation responses to be sent since ground truth is revealed.
--- a/x/emissions/module/rewards/nonce_management.go
+++ b/x/emissions/module/rewards/nonce_management.go
@@ -15,37 +15,29 @@ func UpdateReputerNonce(ctx sdk.Context, k keeper.Keeper, topic types.Topic, blo
 		ctx.Logger().Warn("Error getting unfulfilled worker nonces", "error", err)
 		return err
 	}
+	var updateErr error
 	for _, nonce := range nonces.Nonces {
-		// Must match keeper.ReputerSubmissionWindowBounds, the single source of
-		// truth for when a submission is actually accepted. This used to be its
-		// own formula ending at +extraLag instead of the window's real start
-		// (+extraLag was folded into the acceptance lower bound to close a
-		// permanent-trap bug -- see CHANGELOG). On an uninterrupted epoch grid the
-		// two formulas coincide, but a topic that was inactivated and reactivated
-		// gets a new churn schedule with no relation to an existing nonce's grid
-		// (topic_activation.go), so a reactivated topic could hit a block inside
-		// the old range and announce the window open while it was still rejecting
-		// submissions until the true start.
-		windowStart, windowEnd, err := keeper.ReputerSubmissionWindowBounds(topic, *nonce)
-		if err != nil {
-			ctx.Logger().Warn("Error computing reputer submission window bounds", "error", err)
+		windowStart, windowEnd, boundsErr := keeper.ReputerSubmissionWindowBounds(topic, *nonce)
+		if boundsErr != nil {
+			ctx.Logger().Warn("Error computing reputer submission window bounds", "error", boundsErr)
+			updateErr = boundsErr
 			continue
 		}
-		if block == windowStart {
+		// Active topics are processed once per epoch, so this half-open interval
+		// emits at most once and attributes a shared boundary to the newer nonce.
+		if block >= windowStart && block < windowEnd {
 			types.EmitNewReputerSubmissionWindowOpenedEvent(ctx, topic.Id, nonce.ReputerNonce.BlockHeight, windowEnd)
 		}
-		// Check if current blockheight has reached the blockheight of the nonce + groundTruthLag + epochLength
-		// This means one epochLength is allowed for reputation responses to be sent since ground truth is revealed.
-		closingReputerNonceMinBlockHeight := nonce.ReputerNonce.BlockHeight + topic.GroundTruthLag + topic.EpochLength
-		if block >= closingReputerNonceMinBlockHeight {
-			ctx.Logger().Debug("ABCI EndBlocker: Closing reputer nonce", "topic", topic.Id, "nonce", nonce, "min", closingReputerNonceMinBlockHeight)
-			err = allorautils.CloseReputerNonce(&k, ctx, topic, *nonce.ReputerNonce)
-			if err != nil {
-				ctx.Logger().Error("Error closing reputer nonce", "error", err)
+		if block >= windowEnd {
+			ctx.Logger().Debug("ABCI EndBlocker: Closing reputer nonce", "topic", topic.Id, "nonce", nonce, "min", windowEnd)
+			closeErr := allorautils.CloseReputerNonce(&k, ctx, topic, *nonce.ReputerNonce)
+			if closeErr != nil {
+				ctx.Logger().Error("Error closing reputer nonce", "error", closeErr)
+				updateErr = closeErr
 			}
 		}
 	}
-	return err
+	return updateErr
 }
 
 // Prune reputer and worker nonces
--- a/x/emissions/module/rewards/nonce_management_test.go
+++ b/x/emissions/module/rewards/nonce_management_test.go
@@ -2,11 +2,77 @@ package rewards_test
 
 import (
 	"cosmossdk.io/collections"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 
 	actorutils "github.com/allora-network/allora-chain/x/emissions/keeper/actor_utils"
+	"github.com/allora-network/allora-chain/x/emissions/module/rewards"
 	"github.com/allora-network/allora-chain/x/emissions/testutil"
 	"github.com/allora-network/allora-chain/x/emissions/types"
 )
+
+func countReputerSubmissionWindowOpenedEvents(s *RewardsTestSuite, eventStart int) int {
+	count := 0
+	for _, event := range s.Ctx().EventManager().Events()[eventStart:] {
+		if event.Type == "emissions.v10.EventReputerSubmissionWindowOpened" {
+			count++
+		}
+	}
+	return count
+}
+
+func (s *RewardsTestSuite) TestUpdateReputerNonceWaitsUntilWindowEnd() {
+	topic := s.MockTopic()
+	topic.Id = 9_000_001
+	topic.EpochLength = 100
+	topic.GroundTruthLag = 130
+	nonce := types.Nonce{BlockHeight: 1000}
+
+	err := s.NonceKeeper().AddReputerNonce(s.Ctx(), topic.Id, &nonce)
+	s.Require().NoError(err)
+
+	oldCloseBlock := nonce.BlockHeight + topic.GroundTruthLag + topic.EpochLength
+	s.WithBlockHeight(oldCloseBlock)
+	err = rewards.UpdateReputerNonce(s.Ctx(), *s.EmissionsKeeper(), topic, oldCloseBlock)
+	s.Require().NoError(err)
+
+	unfulfilled, err := s.NonceKeeper().IsReputerNonceUnfulfilled(s.Ctx(), topic.Id, &nonce)
+	s.Require().NoError(err)
+	s.Require().True(unfulfilled)
+}
+
+func (s *RewardsTestSuite) TestUpdateReputerNonceReturnsCloseError() {
+	topic := s.MockTopic()
+	topic.Id = 9_000_002
+	topic.EpochLength = 100
+	topic.GroundTruthLag = 100
+	nonce := types.Nonce{BlockHeight: 1000}
+
+	err := s.NonceKeeper().AddReputerNonce(s.Ctx(), topic.Id, &nonce)
+	s.Require().NoError(err)
+
+	windowEnd := nonce.BlockHeight + topic.GroundTruthLag + topic.EpochLength
+	s.WithBlockHeight(windowEnd)
+	err = rewards.UpdateReputerNonce(s.Ctx(), *s.EmissionsKeeper(), topic, windowEnd)
+	s.Require().ErrorIs(err, sdkerrors.ErrNotFound)
+}
+
+func (s *RewardsTestSuite) TestUpdateReputerNonceEmitsMissedOpenEventWithinWindow() {
+	topic := s.MockTopic()
+	topic.Id = 9_000_003
+	topic.EpochLength = 100
+	topic.GroundTruthLag = 130
+	nonce := types.Nonce{BlockHeight: 1000}
+
+	err := s.NonceKeeper().AddReputerNonce(s.Ctx(), topic.Id, &nonce)
+	s.Require().NoError(err)
+
+	eventStart := len(s.Ctx().EventManager().Events())
+	block := int64(1250)
+	s.WithBlockHeight(block)
+	err = rewards.UpdateReputerNonce(s.Ctx(), *s.EmissionsKeeper(), topic, block)
+	s.Require().NoError(err)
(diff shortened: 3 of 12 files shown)


x/emissions/module/rewards/nonce_management.go
```diff
@@ -15,14 +15,24 @@ func UpdateReputerNonce(ctx sdk.Context, k keeper.Keeper, topic types.Topic, blo
...
+		// truth for when a submission is actually accepted. This used to be its
+		// own formula ending at +extraLag instead of the window's real start
+		// (+extraLag was folded into the acceptance lower bound to close a
+		// permanent-trap bug -- see CHANGELOG). On an uninterrupted epoch grid the
+		// two formulas coincide, but a topic that was inactivated and reactivated
+		// gets a new churn schedule with no relation to an existing nonce's grid
+		// (topic_activation.go), so a reactivated topic could hit a block inside
+		// the old range and announce the window open while it was still rejecting
+		// submissions until the true start.
+		windowStart, windowEnd, err := keeper.ReputerSubmissionWindowBounds(topic, *nonce)
```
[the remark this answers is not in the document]
helder-moreira: Fixed in 86f61287: nonce closure now uses the shared windowEnd boundary. A regression test verifies fractional-lag nonces remain open until that boundary.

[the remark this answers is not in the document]
helder-moreira: Fixed in 86f61287: bounds and close errors use distinct variables, and UpdateReputerNonce retains and returns closure failures. Covered by a regression test.

x/emissions/module/rewards/nonce_management.go
```diff
@@ -15,14 +15,24 @@ func UpdateReputerNonce(ctx sdk.Context, k keeper.Keeper, topic types.Topic, blo
...
+		if err != nil {
+			ctx.Logger().Warn("Error computing reputer submission window bounds", "error", err)
+			continue
+		}
+		if block == windowStart {
```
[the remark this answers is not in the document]
helder-moreira: Fixed in 86f61287: scheduled nonce updates catch up within the half-open window, and ActivateTopic emits the missed event immediately when reactivation happens during an open window. Both paths have regression coverage.

CHANGELOG.md
```diff
@@ -65,6 +65,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # [Unreleased]
 
+### Security
```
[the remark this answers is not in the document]
helder-moreira: Fixed in 86f61287: moved the reputer-window entry into the existing Unreleased Security section and removed the duplicate heading.

x/emissions/module/rewards/nonce_management.go
```diff
@@ -15,27 +15,29 @@ func UpdateReputerNonce(ctx sdk.Context, k keeper.Keeper, topic types.Topic, blo
...
-			ctx.Logger().Debug("ABCI EndBlocker: Closing reputer nonce", "topic", topic.Id, "nonce", nonce, "min", closingReputerNonceMinBlockHeight)
-			err = allorautils.CloseReputerNonce(&k, ctx, topic, *nonce.ReputerNonce)
-			if err != nil {
-				ctx.Logger().Error("Error closing reputer nonce", "error", err)
+		// Active topics are processed once per epoch, so this half-open interval
+		// emits at most once and attributes a shared boundary to the newer nonce.
+		if block >= windowStart && block < windowEnd {
+			types.EmitNewReputerSubmissionWindowOpenedEvent(ctx, topic.Id, nonce.ReputerNonce.BlockHeight, windowEnd)
+		}
+		if block >= windowEnd {
```
xmariachi, line 31: **Issue 1 — P1, severity High, confidence High (reproduced on this branch).**

This close and the acceptance check in `InsertReputerPayload` share block `B = windowEnd`, and `DeliverTx` runs before `EndBlock`. A payload for nonce *k+1* included in `B` is accepted (`B >= windowStart` for *k+1*), `AppendReputerLoss` marks the reputer active and writes the bundle under `(topic, reputer)`, and then this line closes nonce *k* in the same block: the deferred cleanup calls `ResetActiveReputersForTopic` and `ResetReputersIndividualSubmissionsForTopic`, which clear the entire topic prefix and destroy the payload that was just accepted. If the submitting reputer was not active for *k*, its mismatched bundle is also counted at *k*'s close, `CalcNetworkLosses` discards every report on the block-height check, and the close fails with `Sum weight for loss is 0: fraction: divide by zero` — the exact production error this PR is fixing elsewhere. Cost is two nonces per occurrence, and the reputer's transaction reported success, so a well-behaved client never retries.

Reproduced on this branch with a harness driving the real `EndBlocker`, message server and status query, for a client that observes `NextWindowStartBlock == h+1` and lands its transaction at the window start:

```
--- FAIL: TestReputerWindowEdgeBlockEagerClient/aligned_lag
    nonce 300: payload accepted at block 400 (window start) but no network loss was ever set; no bundles were stored at its close
--- FAIL: TestReputerWindowEdgeBlockEagerClient/fractional_lag
    nonce 200: payload accepted at block 400 (window start) but no network loss was ever set; no bundles were stored at its close
```

Note this fails for the **aligned** lag as well, so it is not confined to the topic shape this PR addresses and it is reachable on mainnet. The description does disclose a residual one-block overlap, but describes it as pre-existing and shared, which understates it: the newer payload is accepted and then silently destroyed, and in cosmos-sdk v0.50.14 `runMsgs` skips message handlers outside `execModeFinalize`/`execModeSimulate`, so the window check never gates the mempool — mempool timing alone decides whether a transaction lands in `B`.

Not a blocker for this PR, since it is pre-existing and not worsened here. Recommended fix is the nonce-aware close described in the review body: filter bundles whose `ReputerRequestNonce` does not match the closing nonce before they are appended, and in the deferred cleanup retire only bundles and active flags whose nonce is at or below the closing nonce. No state layout change, and it closes this together with issue 2. The narrower alternative, a half-open window (`upper = start + epoch_length - 1`, closing at `start + epoch_length`), removes only this block, costs one block of submission time, and would require inverting the `prevEnd == start` assertion in `TestReputerSubmissionWindowBounds`.

x/emissions/module/rewards/nonce_management.go
```diff
@@ -15,27 +15,29 @@ func UpdateReputerNonce(ctx sdk.Context, k keeper.Keeper, topic types.Topic, blo
...
-		// Check if current blockheight has reached the blockheight of the nonce + groundTruthLag + epochLength
-		// This means one epochLength is allowed for reputation responses to be sent since ground truth is revealed.
-		closingReputerNonceMinBlockHeight := nonce.ReputerNonce.BlockHeight + topic.GroundTruthLag + topic.EpochLength
-		if block >= closingReputerNonceMinBlockHeight {
-			ctx.Logger().Debug("ABCI EndBlocker: Closing reputer nonce", "topic", topic.Id, "nonce", nonce, "min", closingReputerNonceMinBlockHeight)
-			err = allorautils.CloseReputerNonce(&k, ctx, topic, *nonce.ReputerNonce)
-			if err != nil {
-				ctx.Logger().Error("Error closing reputer nonce", "error", err)
+		// Active topics are processed once per epoch, so this half-open interval
+		// emits at most once and attributes a shared boundary to the newer nonce.
```
xmariachi, line 27: **Issue 2 — P1, severity High, confidence High (code inspection).**

This comment states the assumption the whole tiling argument rests on, and it does not always hold. `CloseReputerNonce` is only reachable through `UpdateNoncesOfActiveTopics`, so a topic that is not processed at its grid block does not close its nonce there, and nonce *k* then stays unfulfilled for a full epoch while nonce *k+1*'s window is entirely open and `GetReputerSubmissionWindowStatus` advertises *k+1* as the current nonce. That is the original overlap, restored at full width, with the same consequence: the payload for *k+1* is consumed and destroyed by *k*'s late close, and both nonces are lost.

Paths that skip or move the scheduled close:

- the topic is inactivated in `GetAndUpdateActiveTopicWeights` because its weight fell below `MinTopicWeight`;
- the topic is evicted from its block by a higher-weight topic in `addTopicToActiveSetRespectingLimitsWithoutMinWeightReset`, which calls `inactivateTopicWithoutMinWeightReset` on the lowest-weight topic;
- `weight.Equal(zero)`, which `continue`s before `UpdateReputerNonce`;
- any error in `GetTopic`, `UpdateTopicEpochLastEnded`, `AddWorkerNonce`, `AddWorkerWindowTopicId` or `PruneReputerAndWorkerNonces`, each of which `continue`s before reaching this function;
- `ActivateTopic`, reachable from any funding or payload transaction via `sendEffectiveRevenueActivateTopicIfWeightSufficient`, which re-grids the topic to `currentBlock + EpochLength` and shifts every subsequent close off the nonce grid by up to `epoch_length - 1` blocks.

Note that `InsertReputerPayload` does not check whether the topic is active, so reputers keep submitting throughout. When the topic resumes and both nonces close in the same `EndBlock`, whichever closes first wipes the topic prefix and the second closes empty.

For completeness, one trigger I initially suspected does **not** apply: `SkimTopTopicsByWeightDesc` uses the same `MaxActiveTopicsPerBlock` bound that `addTopicToActiveSetRespectingLimitsWithoutMinWeightReset` already enforces on the per-block active set, so the skim is effectively a no-op and does not drop topics. Eviction and inactivation are the realistic triggers, not the skim.

Unlike issue 1, the reference offchain node hits this without doing anything unusual, because it follows the status query. Same fix, same PR: the nonce-aware close makes the timing of the close irrelevant to the integrity of stored bundles.

x/emissions/module/rewards/nonce_management.go
```diff
@@ -15,27 +15,29 @@ func UpdateReputerNonce(ctx sdk.Context, k keeper.Keeper, topic types.Topic, blo
...
+		if block >= windowStart && block < windowEnd {
+			types.EmitNewReputerSubmissionWindowOpenedEvent(ctx, topic.Id, nonce.ReputerNonce.BlockHeight, windowEnd)
+		}
+		if block >= windowEnd {
+			ctx.Logger().Debug("ABCI EndBlocker: Closing reputer nonce", "topic", topic.Id, "nonce", nonce, "min", windowEnd)
+			closeErr := allorautils.CloseReputerNonce(&k, ctx, topic, *nonce.ReputerNonce)
```
xmariachi, line 33: **Issue 3 — P2, severity Medium-High, confidence Medium (read path traced, scoring impact not executed).**

When this close proceeds with a mixed set, bundles belonging to a different nonce are not merely ignored. In `CloseReputerNonce` a bundle whose `ReputerRequestNonce.ReputerNonce.BlockHeight` differs from the closing nonce is still appended to `lossBundlesByReputer` and persisted by `InsertActiveReputerLosses(ctx, topic.Id, nonce.BlockHeight, ...)`; it is only skipped later, inside `CalcNetworkLosses`, with the `Reputer bundle has incorrect block height` warning. If at least one other reputer did submit for the closing nonce, the close succeeds, `SetTopicRewardNonce` is set, and `GenerateRewardsDistributionByTopicParticipant` reads the persisted set back via `GetReputerLossBundlesAtBlock` and passes it to `GenerateReputerScores`. The mismatched reputer is then scored and rewarded for nonce *k* using losses computed against nonce *k+1*. The `GetReputerLossBundlesAtBlock` query also serves that incorrect association to clients.

This is reward misattribution rather than a liveness failure, which is why it is separated from issues 1 and 2, but it shares their root cause and it is the cheapest of the three to address:

```go
if bundle.ReputerRequestNonce.ReputerNonce.BlockHeight != nonce.BlockHeight {
    ctx.Logger().Warn("Skipping bundle submitted for a different nonce", ...)
    continue
}
```

placed before `FilterUnacceptedWorkersFromReputerValueBundle`, with `EmitNewActiveReputersSetEvent` narrowed to the retained set. Worth noting that this also changes the observed failure text in the mixed case from `Sum weight for loss is 0 ... divide by zero` to `no valid losses found for reputers`, which is more accurate but will need updating anywhere that error string is being alerted on.

x/emissions/module/rewards/nonce_management_test.go
```diff
@@ -2,12 +2,78 @@ package rewards_test
...
+	err := s.NonceKeeper().AddReputerNonce(s.Ctx(), topic.Id, &nonce)
+	s.Require().NoError(err)
+
+	windowEnd := nonce.BlockHeight + topic.GroundTruthLag + topic.EpochLength
+	s.WithBlockHeight(windowEnd)
+	err = rewards.UpdateReputerNonce(s.Ctx(), *s.EmissionsKeeper(), topic, windowEnd)
+	s.Require().ErrorIs(err, sdkerrors.ErrNotFound)
+}
+
+func (s *RewardsTestSuite) TestUpdateReputerNonceEmitsMissedOpenEventWithinWindow() {
```
xmariachi, line 59: **Issue 4 — P2, severity Medium, confidence High.**

These new unit tests are well targeted, but the fixed path still has no end-to-end coverage, and that gap is why this defect survived from #771 (March 2025) into a live testnet fleet.

Every existing harness uses a lag that divides evenly, so none of them could ever have reproduced it:

- `test/fuzz/topics_test.go` sets `GroundTruthLag: data.epochLength`;
- `test/integration/*` topics use 5/10 or `epochLength`/`epochLength`;
- `test/fuzz/inferences_test.go` waits `GroundTruthLag + 1`.

More immediately, `x/emissions/testutil/test_utils.go` `FullTopicPass` submits at `reputerNonce + GroundTruthLag + 1` and carries the comment "with the current design, `ground_truth_lag` needs to be the same as `epoch_length`". After this change that helper would have its payloads rejected for any fractional-lag topic, because it submits `extraLag - 1` blocks before the window opens. So the harness still encodes the bound this PR is removing and cannot be pointed at a fractional-lag topic without being fixed first.

Two concrete suggestions:

1. Derive the submission height in `FullTopicPass` from `keeper.ReputerSubmissionWindowBounds` instead of `reveal + 1`, and drop the limitation from the comment.
2. Add a fractional-lag topic to the fuzzer's topic generator so the shape is exercised continuously rather than only in the unit table.

I have a reference-client end-to-end test that drives the real `EndBlocker`, message server and status query over several epochs of a 130/100 topic and asserts that every accepted submission yields a network loss for the nonce it was accepted for. It passes on this branch and fails on `dev` with the production error. Happy to contribute it here, it would make this PR self-verifying at the behavioural level rather than only at the arithmetic level.

xmariachi (commented):
## Direction first: we may prefer to make reputer submissions nonce-aware instead

Before reviewing the change itself, a note on direction, because it may supersede this PR.

The failure addressed here is a symptom of a storage decision rather than of the window arithmetic. Reputer loss bundles and the active-reputer set are keyed `(topic_id, reputer)` with no nonce component, and `CloseReputerNonce` consumes every bundle stored for the topic and then clears the whole topic prefix (`ResetActiveReputersForTopic` plus `ResetReputersIndividualSubmissionsForTopic`). The submission window is therefore load-bearing in a way it was never designed to be: it is the only mechanism preventing two nonces from contending for the same single-slot-per-reputer storage. **Any** interval in which a newer nonce can accept a payload before an older nonce has been closed destroys that payload and can fail the older close, however the bounds are computed.

This PR removes the largest such interval, `extraLag` blocks wide, recurring every epoch on every topic whose ground truth lag is not a whole multiple of its epoch length. It does not, and structurally cannot, remove all of them, because the remaining ones are not produced by the bounds:

- the block on which one nonce closes and the next opens is shared, and `DeliverTx` executes before `EndBlock`;
- whenever a close does not run at its scheduled block (topic inactivated, evicted from its block by a higher-weight topic, zero weight, any error path in `UpdateNoncesOfActiveTopics`, or re-gridded by `ActivateTopic`), the original overlap returns in full for an entire epoch.

We therefore expect to prefer making the reputer submission path nonce-aware: `CloseReputerNonce` should consume only bundles whose `ReputerRequestNonce` matches the nonce being closed, and should retire only those bundles and their active flags, leaving payloads belonging to later nonces intact for their own close. That is a strictly better long-term position than continuing to depend on "whichever window is currently open", it decouples transaction admission timing from loss accounting, and it requires no change to the state layout and no protobuf change. **If we take that route it will be raised as a separate PR**, and this one may be superseded, or reduced to the shared-helper extraction, which is valuable on its own.

Both changes alter state-machine behaviour and need a coordinated upgrade height, so the decision worth taking now is whether they ride the same upgrade rather than two.

---

## Review of the change as proposed

**Verdict: sound, minimal and correct for what it targets. Recommend landing it as a first step, with the residual gaps below tracked and scheduled, ideally into the same upgrade.** The lower bound now coincides exactly with the block on which the previous nonce is closed on the epoch grid, which is the right invariant given the current storage model, and consolidating four hand-rolled copies of the formula into `ReputerSubmissionWindowBounds` removes the class of drift that produced this bug (the acceptance bound and the advertised bound had diverged).

### Verification performed

Reviewed against a worktree at the PR head.

| Check | Result |
|---|---|
| `go test` on keeper, queryserver, msgserver and rewards packages | pass |
| `golangci-lint` on `x/emissions/keeper/...` and `x/emissions/module/...` | clean |
| Reference-client end-to-end test driving the real `EndBlocker`, message server and status query, fractional lag 130/100 | **passes on this branch, fails on `dev`** with the production error `Sum weight for loss is 0: fraction: divide by zero`, and the diagnostic shows the bundle consumed at nonce 200's close carried nonce 300 |
| Same harness with a client that lands its transaction on the first block of the window | **fails on this branch** for both aligned (100/100) and fractional (130/100) lag, see the inline comment on the close site |
| Live topic enumeration | mainnet: 23 topics, none with a fractional lag. Testnet: 20 of 85 with fractional lags (28, 29, 31-36, 38, 40-44, 46, 47, 49, 64, 65, 69), consistent with the log evidence in the description |
| `allora-offchain-node` `lib/window_waker.go` | submits only after observing the opened event or `IsOpen` at a height, so it lands at window start + 1 or later and does not hit the shared block |

### Window edges

Writing `B = nonce_k + ground_truth_lag + extraLag + epoch_length`, the block on which nonce *k* closes and nonce *k+1* opens:

- **Block B** remains a real trap, and the description understates it. The payload for *k+1* is not merely "accepted during an overlap", it is accepted and then destroyed by *k*'s cleanup in the same block, while the transaction reports success. Worth noting that in cosmos-sdk v0.50.14 `runMsgs` skips message handlers outside `execModeFinalize`/`execModeSimulate`, so the window check never gates the mempool: a payload signed against a not-yet-open window waits in the mempool and is executed by whichever block includes it. Detail inline.
- **Block B+1 is clean.** Nonce *k* is fulfilled by the deferred cleanup at B, which always runs once the defer is registered, so a late payload for *k* is rejected with `ErrUnfulfilledNonceNotFound` and a payload for *k+1* is accepted and survives to its own close. No hole here.
- **Nonce creation edge is clean.** The reputer nonce is added at `nonce + worker_submission_window`, bounded above by the epoch length and therefore by the ground truth lag, long before its window opens, and `CloseWorkerNonce` fulfils the worker nonce via a defer, so a reputer nonce whose worker nonce is still unfulfilled cannot be closed.
- **Off-grid nonces improve.** The previous close condition `block >= nonce + lag + epoch_length` could close a nonce up to `extraLag` blocks before its own acceptance window ended. `block >= windowEnd` cannot. Good catch, and worth calling out in the description as a second defect fixed.
- **Event timing is unchanged for on-grid nonces**, since the only grid block inside the old emission range was `reveal + extraLag`, which is the new window start. No client-visible change for the reference client.
- The **one edge the bounds cannot cover** is a close that does not happen on schedule; detail inline.

### On priorities

Reassessed with the standard that this is a chain where silent value loss and reward misattribution are not acceptable outcomes, and several were raised:

| # | Issue | Priority | Severity | Confidence |
|---|---|---|---|---|
| 1 | Shared block B accepts a payload for the next nonce and then destroys it, and can fail the current close | P1 | High | High, reproduced |
| 2 | Any close that does not run on schedule reopens the full overlap for an epoch | P1 | High | High |
| 3 | Bundles belonging to another nonce are stored under the closing nonce and reach reputer scoring and rewards | P2 | Medium-High | Medium |
| 4 | No end-to-end coverage of fractional lags, and `FullTopicPass` hardcodes `reveal + 1` | P2 | Medium | High |
| 5 | Upgrade transition is lossy for one cycle on fractional-lag topics and is undocumented | P2 | Low, but deadline-bound to the release notes | Medium-High |
| 7 | Event type strings hardcoded to `emissions.v10.` in new tests | P3 | Low | High |

Issues 1 and 2 are both pre-existing and neither is a blocker for merging this PR; they are P1 as work items, and both are closed by the nonce-aware change described at the top, which is the main argument for doing it. Issue 5 is low severity but has a hard deadline, since it must reach the release notes before the upgrade height is announced.

Two further minor points, not worth inline comments:

- **CHANGELOG entry format.** The new entry is a paragraph and is the only one in `Unreleased` without a `[#993]` link, whereas neighbouring entries are a single linked sentence. Suggest reducing it to one sentence with the PR link and keeping the full explanation here in the PR body; the file's own guidance also asks for breaking changes to be carried into the release notes rather than only bolded in the entry.
- **Defensive guards in `ReputerSubmissionWindowBounds`.** `nonce.ReputerNonce.BlockHeight + topic.GroundTruthLag` is computed before being checked against `math.MaxInt64`, and `topic.GroundTruthLag % topic.EpochLength` would panic on a zero epoch length. Both are unreachable today: `CreateNewTopic` validates `EpochLength` against `MinEpochLength`, and `GenesisState.Validate` does not import topics at all. Raised only because this helper is now reachable from query paths on any stored topic, and a panic on this path in `EndBlocker` would halt the chain. Defence in depth, not a defect.

### Smaller notes

- The description omits one mechanism that defines who is actually at risk: `AppendReputerLoss` rejects a second submission from a reputer already marked active, so a reputer that submitted for nonce *k* cannot overwrite its own bundle with *k+1*. Only reputers not active for *k*, meaning late joiners and reputers whose previous payload was itself destroyed, trigger the failure. That is also what sustains the cascade on single-reputer topics and is worth stating.
- `UpdateReputerNonce` now returns the last observed error rather than whatever the final iteration left in `err`. The caller only logs it, so there is no halt risk, and `TestUpdateReputerNonceReturnsCloseError` pins the behaviour. Good change.
- `emitOpenReputerSubmissionWindowEvents` cannot double-emit with the grid path for the same nonce: if activation occurs at a block inside `[start, end)`, the next grid block is at or beyond `end` and fails the half-open check. Events are not part of `LastResultsHash`, so this emission is not consensus-relevant.
- `TestReputerSubmissionWindowBounds` asserts `prevEnd == start` as the desired invariant. That equality is precisely the shared block of issue 1. If the half-open alternative is ever preferred over the nonce-aware change, this is the assertion to invert.
- Testnet topic 65 has `epoch_length=1286` and `ground_truth_lag=5142`, so `extraLag` is only 2. The failure needs a client landing in the gap, not a wide gap, which is worth remembering when judging exposure.

CHANGELOG.md
```diff
@@ -82,6 +82,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security
 
+* The reputer submission window now opens at the end of the epoch in which ground truth is revealed (`nonce + ground_truth_lag + extraLag`) rather than at the revelation itself. Previously, on a topic whose `ground_truth_lag` was not a whole multiple of `epoch_length`, a nonce's window opened `extraLag` blocks before the *previous* nonce closed on the epoch grid. A reputer submitting in that gap stored a bundle for the newer nonce, and because reputer loss bundles are keyed by `(topic, reputer)` with no nonce, closing the older nonce read that bundle, discarded every report on the block-height check and failed with `Sum weight for loss is 0`. The topic then produced no further network losses and, since `SetTopicRewardNonce` is only reached on success, paid no rewards — permanently, with every transaction still reporting success. Topics whose lag divides evenly were unaffected (`extraLag` is zero for them). **This changes transaction acceptance and therefore requires a coordinated upgrade height.** No store migration; `ConsensusVersion` is unchanged.
```
xmariachi, line 85: **Issue 5 — P2, severity Low, but deadline-bound to the release notes.**

The upgrade transition is lossy for one cycle on topics that are currently exposed, and that is not documented anywhere in the PR or the entry.

At the upgrade height, fractional-lag topics on testnet will already hold payloads accepted for nonce *k+1* under the old lower bound while nonce *k* is still open. After the upgrade, nonce *k* closes at `reveal_k + extraLag + epoch_length`, reads those bundles, fails with `Sum weight for loss is 0` if no reputer submitted for *k*, and its cleanup destroys them. Nonce *k+1* then closes empty unless someone resubmits inside its window. From *k+2* onwards the cycle is clean, because clients can no longer be steered early. So: self-healing, no state corruption, but expect up to two failed closes per affected topic immediately after the upgrade, on the 20 testnet topics listed in the review.

Mainnet has no fractional-lag topic today and is unaffected.

Worth a sentence here and in the release notes, so that the post-upgrade error burst is recognised as the expected tail of the fix rather than as a regression introduced by it. Misreading those errors during an upgrade window is exactly the kind of thing that triggers an unnecessary rollback decision.

x/emissions/module/rewards/nonce_management_test.go
```diff
@@ -2,12 +2,78 @@ package rewards_test
...
 	actorutils "github.com/allora-network/allora-chain/x/emissions/keeper/actor_utils"
+	"github.com/allora-network/allora-chain/x/emissions/module/rewards"
 	"github.com/allora-network/allora-chain/x/emissions/testutil"
 	"github.com/allora-network/allora-chain/x/emissions/types"
 )
 
+func countReputerSubmissionWindowOpenedEvents(s *RewardsTestSuite, eventStart int) int {
+	count := 0
+	for _, event := range s.Ctx().EventManager().Events()[eventStart:] {
+		if event.Type == "emissions.v10.EventReputerSubmissionWindowOpened" {
```
xmariachi, line 16: **Issue 7 — P3, severity Low, confidence High.**

Matching on the literal `"emissions.v10.EventReputerSubmissionWindowOpened"` here, and in `TestActivateTopicEmitsOpenReputerWindowEvent`, couples these assertions to the proto package version. At the next emissions version bump every such string has to be found and edited by hand, and until then the test fails with an unhelpful "expected 1, got 0" rather than naming the cause.

Either of these is version-proof:

```go
proto.MessageName(&types.EventReputerSubmissionWindowOpened{})
```

or parsing and type-switching, which additionally lets the assertion check the payload rather than only the count:

```go
if evt, err := sdk.ParseTypedEvent(abci.Event(event)); err == nil {
    if opened, ok := evt.(*types.EventReputerSubmissionWindowOpened); ok { ... }
}
```

Minor, and entirely reasonable to defer.

